PR #9536, opened 2015-01-08
runtime: s/runtime.rtype/runtime._type/g in runtime-gdb.py

custa:
wrong struct name


bradfitz:
We don't use pull requests. See https://golang.org/doc/contribute.html

PR #9703, opened 2015-01-27
cmd/dist: update build msg notice

sagivo:
simple change - should be `$PATH` instead of just `PATH`.


minux:
Sorry we don't take pull requests.

Please see golang.org/doc/contribute.html

minux:
And the change is not entirely right either. On windows, it should be %PATH%.

I vote to keep it as is.

PR #9874, opened 2015-02-14
net/http: Adding description for undocumented behavior

laszlothewiz:
Added documentation for the redirect that happens, when a path ending with `index.html` is provided to `ServeFile`


ianlancetaylor:
Thanks, but we don't use pull requests for Go.  Please see http://golang.org/doc/contribute.html .

PR #9220, opened 2014-12-08
The Go project does not use pull requests.

adg:
## Please do not send pull requests to the golang/\* repositories.

We do, however, take contributions gladly.

See https://golang.org/doc/contribute.html

Thanks!

PR #9925, opened 2015-02-18
math: Create sumto.go

PlastecProfiles:
This is an efficient way to add number from 1 to say 100.


minux:
Sorry, we don't use pull requests.

See https://github.com/golang/go/pull/9220.

minux:
And this API addition hasn't been discussed, and I'm sure
we won't add that.

PR #10082, opened 2015-03-04
cmd/go: Update vcs.go

orchid-hybrid:
Never use HTTP for git.

This is a dangerous security flaw, but easily fixed.
- https://twitter.com/wiretapped/status/573173567728308224


cespare:
The Go project does not use pull requests.

Please see [the contribution guidelines](https://golang.org/doc/contribute.html).

PR #10114, opened 2015-03-08
sort: Spelling mistake

zg:
I noticed a spelling mistake in the comments.


minux:

zg:
Oops, my mistake.

Issue #10322, opened 2015-04-02
x/tools/go/vcs: FromDir fails to recognize git submodules


jpcummins:
I'm using `godep save` in a project that is imported as a git submodule. The command fails with the error:

    godep: directory "/home/vagrant/workspace/go/src" is not using a known version control system

Digging deeper, I found that godep is using https://godoc.org/golang.org/x/tools/go/vcs#FromDir to determine if the current directory contains a git repo. This command does not correctly identify git submodules because the [implementation searches for .git directory](https://github.com/golang/go/blob/5bb89eb/src/cmd/go/vcs.go#L425). Git submodules don't have a .git directory, just a .git file that contains a link to the parent repo.

dgsb:
I have a patch for this issue

bradfitz:
@dgsb, if you haven't seen it yet: https://golang.org/doc/contribute.html

dgsb:
The changes have been posted on gerrit. I'm not sure if I shoud have received an email notification.

dgsb:
Patch waiting for review https://go-review.googlesource.com/#/c/21430/

shurcooL:
Is this really an issue? How should `vcs.FromDir` handle git submodules inside git repositories?

From https://godoc.org/golang.org/x/tools/go/vcs#FromDir:

> FromDir inspects dir and its parents to determine the version control system and code repository to use. On return, root is the import path corresponding to the root of the repository.

The documentation doesn't make it explicit or clear, but I interpret that to mean that it will get the vcs that contains dir at the top-most level. So if you're inside a submodule directory of a git repo, it should still return the top level git repo, no?

Can you link to the issue in `godep`, perhaps the issue is in how it's using `vcs.FromDir` and should be fixed there instead. It's hard to know without more details about the situation where it fails, so can you please elaborate on what you mean by "a project that is imported as a git submodule"?

dgsb:
> So if you're inside a submodule directory of a git repo, it should still return the top level git repo, no?
No, I indeed I don't think it should, a subrepository is repository by itself.
In my specific use case I have a git repository which is used only to gathers repository (not only go related) together.
Some of the go registered as submodule are used to follow upstream which I may contribute to, and at some point in time I wanted to vendor these into a consuming project with godep.
With the current implementation it's not possible.
The godep related issue is tools/godep#445

shurcooL:
Can you provide more details about such a setup with a git repository that gathers other repositories as submodules?

[...]

PR #10331, opened 2015-04-03
go/vcs: Removed directory check for vcs detection

jpcummins:
https://github.com/golang/go/issues/10322

Git uses a file (instead of a directory) for submodule repositories.


minux:

PR #10349, opened 2015-04-04
net: Update comment on the `To4` function.

yawboakye:
The current implementation of `To4` returns `nil` only if the IP address
is invalid, i.e. has the wrong length.

### DIFF
--- a/src/net/ip.go
+++ b/src/net/ip.go
@@ -176,7 +176,7 @@ func isZeros(p IP) bool {
 }
 
 // To4 converts the IPv4 address ip to a 4-byte representation.
-// If ip is not an IPv4 address, To4 returns nil.
+// If ip is not an IP address (it is the wrong length), To4 returns nil.
 func (ip IP) To4() IP {
 	if len(ip) == IPv4len {
 		return ip


minux:

Please see https://golang.org/doc/contribute.html for the contribution guideline.

And btw, I don't think the change is an improvement.

yawboakye:
@minux I didn't know you didn't accept pull requests (via GitHub?). Also the change is not an improvement, yes. It doesn't reduce CPU cycles or the memory requirements of `To4`. But it corrects a misleading comment, a very wrong description of how the function behaves. If the comments are meant to explain the function then it's important to keep it correct. imo.

minux:
The old comment says "if ip is not an IPv4 address, To4 returns nil",
which is exactly what the code does.

Your proposed change changed that sentence to
"If ip is not an IP address (it is the wrong length), To4 returns nil."

This is more confusing that it was. If the user passes in an IPv6
address (not an IPv4-mapped IPv6 address), and expect To4
to return a non-nil address. According to your docs, it should work
because the user has passed an IP address.

If you say an IPv6 address is not the "correct" length for To4, then
what about IPv4-mapped IPv6 addresses? They have the length
of IPv6 address, but To4 can still convert them to IPv4.

yawboakye:
Got it. Thanks :)

PR #10352, opened 2015-04-06
cmd/yacc: Use golint-preferred initialism in generated code.

jaqx0r:
Rename yyEofCode to yyEOFCode to appease golint's naming rules.


mikioh:

jaqx0r:
Nah, probably won't. This scratched an itch and the barrier to contribute
was so low via github.

PR #10434, opened 2015-04-13
net/http: Update Request documentation to include detail about Form population.

bobbytables:
The documentation lacks in the semantics of how the Form field is populated on http.Request. This updates the Form field on Request to talk about the business logic behind population.


ianlancetaylor:
I'm sorry, we don't take pull requests.  To contribute please see http://golang.org/doc/contribute.html .  Thanks.

PR #10498, opened 2015-04-18
database/sql: Updated some documentation for "prepare".

dlochrie:
The general explanation is the same, but it was edited for grammar and readability.


mdempsky:
The Go project does not use pull requests.

Thanks!

dlochrie:
Ok, thanks. Seems like a lot of work for a text change!

PR #10665, opened 2015-05-02
net/url: Catch URL Parse error with port and no scheme

grbritz:
I found that when parsing URLs with a port and no scheme, the parser would return the hostname in the Scheme field, and an empty string in the host field. This test should catch that.

### DIFF
--- a/src/net/url/url_test.go
+++ b/src/net/url/url_test.go
@@ -369,6 +369,14 @@ var urltests = []URLTest{
 		},
 		"http://[fe80::1%25en01-._~]:8080/",
 	},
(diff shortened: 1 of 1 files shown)


minux:
Please see https://github.com/golang/go/pull/9220.

grbritz:
Ah I hadn't seen that. I don't want to go through the process of becoming a contributor for such a simple change. Can anyone pick this up? It would be nice of the URL parser was more robust.

bradfitz:
@grbritz, a host:port isn't a URL. If that's what you've got, there's already `net.SplitHostPort`.

PR #10744, opened 2015-05-07
doc: Update signature for File.Read to match code example

brettcannon:
The signature then also matches the documentation for the os package.


Thomasdezeeuw:
Go doesn't use pull requests; https://github.com/golang/go/pull/9220.

brettcannon:
Sorry about that! Forgot while I was reading the docs and found the small error. I'll make a proper patch submission.

minux:
I think the original signature is fine.

PR #10962, opened 2015-05-26
time: Update docstring for Format in format.go

AlexFinkel:
The reference time described in the docstring for Format doesn't provide a value for nanoseconds. This hides the functionality provided by the Format function to render milli, micro and nanoseconds.

This has led to some confusion that Format can't handle sub-second precision: https://groups.google.com/forum/#!topic/golang-nuts/TIinEnPu5wE

### DIFF
--- a/src/time/format.go
+++ b/src/time/format.go
@@ -401,7 +401,7 @@ func (t Time) String() string {
 // Format returns a textual representation of the time value formatted
 // according to layout, which defines the format by showing how the reference
 // time, defined to be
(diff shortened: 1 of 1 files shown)


adg:
Hi, please note that we don't take pull requests, but rather use a different code review system. See https://golang.org/doc/contribute.html

As for this particular issue, I'm not sure this fix is correct. Please file an issue so we can discuss the right fix. Thanks.

AlexFinkel:
Thanks for letting me know adg, filed issue https://github.com/golang/go/issues/10963.

PR #10974, opened 2015-05-28
cmd/cgo: provides directly calling a c function without runtime thread pool involved.

xiezhenye:
currently using c.func_call() instead of C.func_call() to perform a direct cgo call.


minux:
Sorry, but we do not take pull requests.

Please see the contribution guideline at
https://golang.org/contribute.html.

Note: this kind of change needs to be discussed on the mailing list first
when the tree is not frozen for releases (it's now).

PR #11649, opened 2015-07-10
net/http: Expose http server for ACME letsencrypt

coolaj86:
https://LetsEncrypt.org's release is just around the corner, but it's currently not possible to dynamically retrieve and renew certificates in go with an active http server.

To do so requires fixing a tiny regression introduced sometime between the move from httputils and now. See https://golang.org/pkg/net/http/httputil/#ServerConn.

We need to re-expose `http.NewConn` and `http.conn.Serve`.

I've got a working demo of dynamically loading certificates with this change here:
https://gist.github.com/coolaj86/16ed8fd810e19dec71be

I've already signed the CLA and with a little coaching I'm sure I could turn my example into an appropriate test case.

### DIFF
--- a/src/pkg/net/http/server.go
+++ b/src/pkg/net/http/server.go
@@ -172,7 +172,7 @@ func (w *response) ReadFrom(src io.Reader) (n int64, err error) {
 const noLimit int64 = (1 << 63) - 1
 
 // Create new connection from rwc.
-func (srv *Server) newConn(rwc net.Conn) (c *conn, err error) {
+func (srv *Server) NewConn(rwc net.Conn) (c *conn, err error) {
 	c = new(conn)
 	c.remoteAddr = rwc.RemoteAddr().String()
 	c.server = srv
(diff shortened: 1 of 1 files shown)


mikioh:
Please take a look at #9220.

coolaj86:
Do I really have to go through a 2 hour process to make a 4-line fix? I mean, I need the regression fixed, but I already spent so many hours figuring it out and now more hours figuring out some process that I'm probably only going to use once or twice.... sadness

</3

The ache of my bleeding heart...

Oh the bemoanments.... agaoeurcgkhnetsus.O͇̹̺ͅƝ̴ȳ̳ TH̘Ë͖́̉ ͠P̯͍̭O̚​N̐Y̡ H̸̡̪̯ͨ͊̽̅̾̎Ȩ̬̩̾͛ͪ̈́̀́͘ ̶̧̨̱̹̭̯ͧ̾ͬC̷̙̲̝͖ͭ̏ͥͮ͟Oͮ͏̮̪̝͍M̲̖͊̒ͪͩͬ̚̚͜Ȇ̴̟̟͙̞ͩ͌͝

PR #12072, opened 2015-08-07
all: typofix - https://github.com/vlajos/misspell_fixer

vlajos:


griesemer:
Thanks, but we don't accept pull requests (README.md on
https://github.com/golang/go). See also:

https://golang.org/doc/contribute.html

For how to contribute.

- gri

vlajos:
Hmm. Thank you.
Unfortunately I don't have too much time to pass all steps/phases.

Anyway if you want, feel free to run my script to fix some small typos.
(There were a couple of false positives, so be careful)

griesemer:
Ok, thanks. This may have to wait for 1.6 - we're in the code freeze before
the 1.5 release.
- gri

PR #12264, opened 2015-08-21
encoding/asn1: Indefinite-length structured types and "ANY DEFINED BY"

dtbartle:
@zenazn could you take a look at this and see if it seems sane before I send it upstream? In particular, I'd like feedback on the "definedby" mechanism.


mikioh:
Please take a look at #9220.

PR #12858, opened 2015-10-07
Add new day duration.

alienchow:
Unlike a Month or Year which might vary, a day duration is always 24 hours, so unsure why this isn't in the time package.


adg:
@alienchow A day [isn't](https://en.wikipedia.org/wiki/Leap_second) [always](https://en.wikipedia.org/wiki/Daylight_saving_time) 24 hours.

alienchow:
@adg Interesting. I knew about DST but didn't know about leap second. Thanks for the heads up. :+1:

PR #13597, opened 2015-12-13
Add a Delta() method to determine how many are waiting

purpleidea:
This is my first crack at improving go core. I'm not a pro C programmer, so I suspect this might be able to be simplified to remote the atomic operation. This patch is useful so that when you pass a WaitGroup struct you can also examine it easily and know the net delta.

### DIFF
--- a/src/sync/waitgroup.go
+++ b/src/sync/waitgroup.go
@@ -136,3 +136,11 @@ func (wg *WaitGroup) Wait() {
 		}
 	}
 }
+
+// Return the net delta added to the wait counter.
+func (wg *WaitGroup) Delta() int {
+	statep := wg.state()
+	state := atomic.AddUint64(statep, uint64(0)<<32)
+	v := int32(state >> 32)
+	return v
+}


purpleidea:
Gahhh... Dear @gopherbot, consider instead submitting my patch to gerrit on my behalf, since it's not worth it for me to contribute with such a barrier for a simple patch like this.

minux:
We can't add Delta method, because it's impossible
to use it correctly.

Because there could be concurrent Add/Done calls,
you can't use the value returned from Delta in any
useful way.

purpleidea:
On Sun, Dec 13, 2015 at 3:09 AM, Minux Ma <notifications@github.com> wrote:
>
> We can't add Delta method, because it's impossible
> to use it correctly.
>
> Because there could be concurrent Add/Done calls,
> you can't use the value returned from Delta in any
> useful way.

Actually, while I'm not an expert on concurrency, and this may be
true, it doesn't mean that you can't use Delta in any meaningful way.

One use case could be deciding if a passed wg struct is meant to be
further used or not based on if it has a Delta() of 0. If it has a
positive value then know it's in use or prepared by the parent object.
This is a way to include a bool flag inside the same parameter.

Hope this clears it up.

minux:
Why we can't use the passed in WaitGroup if Delta() is zero?

If you want to propose this, please follow the proposal process
and send a proposal. Please see https://golang.org/s/proposal-process.

(I think adding Delta() is unlikely to be accepted, and I suggest
you search for old discussions on this first.)

purpleidea:
On Sun, Dec 13, 2015 at 6:09 AM, Minux Ma <notifications@github.com> wrote:

> Why we can't use the passed in WaitGroup if Delta() is zero?
>
> If you want to propose this, please follow the proposal process
> and send a proposal. Please see https://golang.org/s/proposal-process.
>
> (I think adding Delta() is unlikely to be accepted, and I suggest
> you search for old discussions on this first.)
>


Thanks for replying anyways. Cheers

PR #13689, opened 2015-12-19
Fix typo in README.md

sgerhardt:


bradfitz:
I've filed a bug about this if you don't want to use Gerrit. We can take care of it. Thanks!

sgerhardt:
Cool, sounds good. I'll move onto to looking at bigger things than minor typos in a readme then. :-)

PR #13899, opened 2016-01-10
fix undefined: base64.RawURLEncoding

vyachin:


vyachin:
simple test 

name := base64.RawURLEncoding.EncodeToString("string")

without this smaller fix i have error 

undefined: base64.RawURLEncoding

minux:
This patch is wrong. We intentionally want
(*Encoding).WithPadding to not modify
the receiver (otherwise it's possible to
modify existing base64.StdEncoding.

Please see the relevant CL for discussion.

Issue #17326, opened 2016-10-03
go/build: permit vendoring golang.org from std without breaking users


bradfitz:
In #16333 a user reported that they were unable to vendor code which the Go standard library itself vendored.

To "fix" the problem for the 1.7 release, we renamed the "$GOROOT/src/vendor/golang.org" directory to "golang_org" with an underscore, so it wouldn't conflict with users.

This is slightly annoying for us to maintain in the stdlib.

We should probably fix it properly in `go/build` and/or `cmd/go`.

/cc @ianlancetaylor @rsc

ianlancetaylor:
The standard library no longer has a top level vendor directory.  The contents have moved to `internal/x` to fix #26988 and #27285.  I think that fixes this problem, so closing.

Issue #16333, opened 2016-07-12
go/build: regression in go 1.7rc1, cannot vendor golang.org/x/net/http2


vanackere:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go are you using (`go version`)?

go 1.7rc1

2. What operating system and processor architecture are you using (`go env`)?
```
OARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/vince/dev"
GORACE=""
GOROOT="/home/vince/go"
GOTOOLDIR="/home/vince/go/pkg/tool/linux_amd64"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build541667125=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
```

3. What did you do?

Put a copy of golang.org/x/net/http2 at $GOPATH/src/vendor/golang.org/x/net/http2

Try to build the following program using go build :

```go
package vendoring17

import _ "golang.org/x/net/http2"
```

4. What did you expect to see?

Everything was working fine using go 1.6

5. What did you see instead?

With go 1.7 I get :

package github.com/vanackere/bugs/vendoring17
	imports golang.org/x/net/http2: use of vendored package not allowed

This bug appears at https://github.com/golang/go/commit/0104a31b8fbcbe52728a08867b26415d282c35d2 where golang.org/x/net/http2/hpack moves from internal to vendor

From a first look it appears that (*Context) Import is confused by this change and assumes that vendor/golang.org/x/net/http2 is a package located in GOROOT. I don't have a proper fix for the moment (still looking into it) but I believe that this bug should be fixed for go 1.7.

bradfitz:
I can reproduce. Thanks for the report.

ianlancetaylor:
As far as I can tell this only fails when the directory is precisely `$GOPATH/src/vendor/golang.org/x/net/http2`.  It does not fail if the vendor directory is not at top level.  And it does not fail if the vendor directory is something other than `golang.org/x/net/http2`.

ianlancetaylor:
[...]

PR #20468, opened 2017-05-23
golang_org -> golang.org

bpmason1:
Please do not send pull requests to the golang/* repositories.

We do, however, take contributions gladly.

Thanks!


bradfitz:
Also, the underscore is intentional as a workaround. See #16333 and #17326

PR #23754, opened 2018-02-09
doc: minor wording improvement to the diagnostics guide

rakyll:


bradfitz:
Btw, the Go 1.10 tree is still frozen, but this was kinda special, as we are testing GerritBot and this is a safe change.

PR #23755, opened 2018-02-09
cmd/go/internal/get: fix regexes for "any" character matching

ofpiyush:
Minor bug: `hubajazz.net` and `gitbapache.org` would match, probably shouldn't


vuthithanhbinh1220:
https://go-review.googlesource.com/c/go/+/92997#message-da0832b01586fcfae4eb7f43cb71b1af5bf4427c

andybons:
Apologies for the duplicate posts on this.

PR #23761, opened 2018-02-09
image/gif: support non-looping animated gifs (LoopCount=-1)

### LINKED ISSUE #15768: image/gif: EncodeAll fails to allow non looping animated gif
1. What version of Go are you using (`go version`)?
   go version go1.6.2 
2. What operating system and processor architecture are you using (`go env`)?
   darwin/amd64
3. What did you do?
   In package image/gif, I'm trying to use EncodeAll() to create a non-looping animated gif with several frames, but it will always loop. In the GIF specifications `0` means forever, so `-1` should the way to _go_ to disable looping, unfortunately setting this value does not work in go.
4. What did you expect to see?
   I should be able to set `LoopCount = -1` and my resulting gif should not loop.
   I want an animated gif, but not a looping animated gif.
5. What did you see instead?
   My output gif is looping forever (same as LoopCount = `0`).

The first problem lies in [src/image/gif/write.go](https://golang.org/src/image/gif/writer.go)

```
   299      if g.LoopCount < 0 {
   300          g.LoopCount = 0
   301      }
```

This should be removed to allow a negative value.

The second problem is that Netscape Looping extension header is written in the file whenever there are several frames:

```
   147      if len(e.g.Image) > 1 {
```

I believe this should be changed to:

```
if len(e.g.Image) > 1 && e.g.LoopCount >= 0 {
```

Another way to fix the problem would be to change the meaning of LoopCount '0' value to actual "no loop", but this could break the expected behaviour for some people. As in other languages, setting `-1` sounds ok as long as it's commented somewhere.

pteichman:
The Netscape looping application extension encodes how many
times the animation should restart, and if it's present
there is no way to signal that a GIF should play only once.

Use LoopCount=-1 to signal when a decoded GIF had no looping
extension, and update the encoder to omit that extension
block when LoopCount=-1.

Fixes #15768

### DIFF
--- a/src/image/gif/reader.go
+++ b/src/image/gif/reader.go
@@ -570,10 +570,11 @@ func Decode(r io.Reader) (image.Image, error) {
 type GIF struct {
 	Image []*image.Paletted // The successive images.
 	Delay []int             // The successive delay times, one per frame, in 100ths of a second.
-	// LoopCount is the number of times an animation will be restarted
-	// during display. This means for LoopCount=N, each frame will be
-	// displayed N+1 times. Use 0 for infinite looping, -1 to show each
-	// frame once.
+	// LoopCount controls the number of times an animation will be
+	// restarted during display.
+	// A LoopCount of 0 means to loop forever.
+	// A LoopCount of -1 means to show each frame only once.
+	// Otherwise, the animation is looped LoopCount+1 times.
 	LoopCount int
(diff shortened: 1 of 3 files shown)


andybons:
Apologies for the barrage of duplicate posts on this. Working out some kinks.

pteichman:
@andybons No worries, happy to help iron them out.

PR #23805, opened 2018-02-13
add function for bool in math/rand

### LINKED ISSUE #23804: Add function for random bool in math/rand
I want to be able to call `rand.Bool()` and receive a pseudo-random bool when I import `math/random`

kirinrastogi:
Closes #23804

I wasn't sure how to test this.

### DIFF
--- a/src/math/rand/rand.go
+++ b/src/math/rand/rand.go
@@ -210,6 +210,11 @@ again:
 	return f
 }
 
+// Bool returns a pseudo-random bool in [true, false].
+func (r *Rand) Bool() bool {
+	return r.Int63()%2 == 0
+}
+
 // Perm returns, as a slice of n ints, a pseudo-random permutation of the integers [0,n).
 func (r *Rand) Perm(n int) []int {
 	m := make([]int, n)


kirinrastogi:
I signed it!

src/math/rand/rand.go
```diff
@@ -210,6 +210,11 @@ again:
 	return f
 }
 
+// Bool returns a pseudo-random bool in [true, false].
```
mewmew: While `true` and `false` is not specified as as `1` and `0` (respectively) in Go, given the widespread use of that definition in other languages, the range should be swapped, as follows.

```diff
-[true, false].
+[false, true].
```

WissCoast:
@kirinrastogi any updates?

src/math/rand/rand.go
```diff
@@ -210,6 +210,11 @@ again:
...
+func (r *Rand) Bool() bool {
+	return r.Int63()%2 == 0
```
bbuck, line 215: Would it not be more efficient to do something like this?

```go
	return r.Int63() & 1 == 0
```

PR #23893, opened 2018-02-17
cmd/cgo: fix \'error: function declaration isn’t a prototype [-Werror…

ghost:
While working on a very simple example of calling a Go function from C I ran into a bug that is fairly edge case. It is only present when one tries to export a function with ZERO parameters, if you supply at least one parameter it will work without issue.

```
# WORKS
//export HelloWorld
func HelloWorld(i int) {
	fmt.Println("GO Code!!", i)
}
```

However if you export a function in a Go source without parameters like:

```
# DOES NOT WORKS
//export HelloWorld
func HelloWorld() {
	fmt.Println("GO Code!!")
}
```

It results in the following error:

`"cgo-gcc-export-header-prolog:41:1: error: function declaration isn’t a prototype [-Werror=strict-prototypes]"`

This error occurs because the Go based C Library header is not rendered correctly by `cgo`. Currently `cgo` renders the exported function in the header file incorrectly:
`
extern void HelloWorld();`

But C requires requires a function without parameters being defined in a header to be definitively declared meaning 0 input parameters is void not blank:

`extern void HelloWorld(void);`

I have corrected this by checking for 0 parameters and in this condition writing void in between the parenthesis, otherwise it continues as it did to write the 1 or more parameters inside the parenthesis. This solves the issue and works for all cases now. 

Please let me know if there are any issues with this pull request.

### DIFF
--- a/src/cmd/cgo/out.go
+++ b/src/cmd/cgo/out.go
@@ -1019,23 +1019,25 @@ func (p *Package) writeGccgoExports(fgo2, fm, fgcc, fgcch io.Writer) {
 			fmt.Fprintf(fgcch, "};\n")
 			fmt.Fprintf(cdeclBuf, "struct %s_return", exp.ExpName)
 		}
-
 		cRet := cdeclBuf.String()
-
 		cdeclBuf = new(bytes.Buffer)
 		fmt.Fprintf(cdeclBuf, "(")
-		if fn.Recv != nil {
-			fmt.Fprintf(cdeclBuf, "%s recv", p.cgoType(fn.Recv.List[0].Type).C.String())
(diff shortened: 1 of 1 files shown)


andybons:
Unless the CLA is signed, we cannot accept your change.

Additionally, the first comment contains Markdown, which isn’t permitted per https://github.com/golang/go/blob/master/.github/PULL_REQUEST_TEMPLATE.md

andybons:
@ghost friendly ping on the CLA.

davecheney:
I'm closing this pull request as the original user has deleted there account.

https://help.github.com/articles/deleting-your-user-account/

PR #23994, opened 2018-02-21
net/http: lock the read-only mutex in shouldRedirect

dmathieu:
Since that method uses 'mux.m', we need to lock the mutex to avoid data races.

### DIFF
--- a/src/net/http/serve_test.go
+++ b/src/net/http/serve_test.go
@@ -581,6 +581,16 @@ func TestServeWithSlashRedirectForHostPatterns(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


andybons:
You have to change the base branch to `master` instead of `release-branch.go1.10`

m90:
Is there any info an which future release of go will contain this fix? I can't find it either in 1.10.2 or 1.11 milestones.

ianlancetaylor:
@m90 Note that few people read comments on GitHub pull requests.  For better places to ask questions see https://golang.org/wiki/Questions .

This fix will be in the future 1.11 release.

PR #24002, opened 2018-02-21
net/http: remove a duplicated article

ktr0731:
I removed a duplicated article in `net/http/h2_bundle.go`


andybons:
This was done in https://github.com/golang/net/commit/cbe0f9307d0156177f9dd5dc85da1a31abc5f2fb

PR #24004, opened 2018-02-21
net/http: optionally add headers on redirect

### LINKED ISSUE #19973: net/http: add knobs for which headers Client forwards on redirect?

1.8.1

amd64/darwin

### What did you do?

Making a request to upload a file that is setting the `content-type` of `application/x-compressed` which results in a successful `302` redirect to a task monitoring endpoint.
 
The request succeeds and the task monitoring endpoint returns a 200.

### What did you see instead?

The endpoint fails due to the fact that it has received the additional header.

I know I can implement a `CheckRedirect` function that would allow me to strip out the headers but it seems like the go 1.7.X behavior was more desirable (providing the function to forward the headers).

It actually might be nice to make this a flag on the client (so headers can be optionally be forwarded or not). Having to implement `CheckRedirect` just to not send along headers feels a tad bit heavy handed but I am open to it.

I also know that the API server, which I have no control over, could be modified to ignore additional headers but the code I have talking to this server worked in 1.7.X and now doesn't work in 1.8.1 (slightly surprising).

zachgersh:
Fixes #19973

### DIFF
--- a/src/net/http/client.go
+++ b/src/net/http/client.go
@@ -82,6 +82,10 @@ type Client struct {
 	//
 	// If RedirectHeader is nil, all headers will be forwarded on all
 	// subsequent redirects.
+	//
+	// Headers "Authorization", "Www-Authenticate", "Cookie",
+	// and "Cookie2" bypass this function and are sent along
+	// on all redirects.
 	RedirectHeader func(headerKey string) bool
 
 	// Jar specifies the cookie jar.
--- a/src/net/http/client.go
+++ b/src/net/http/client.go
@@ -75,6 +75,15 @@ type Client struct {
 	// which is to stop after 10 consecutive requests.
 	CheckRedirect func(req *Request, via []*Request) error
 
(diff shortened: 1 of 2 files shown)


src/net/http/client.go
```diff
@@ -698,6 +708,35 @@ func (c *Client) makeHeadersCopier(ireq *Request) func(*Request) {
 	}
 }
 
+func shouldCopyHeaderOnRedirect(canonicalHeaderKey string, initial, dest *url.URL, headersPolicy func(string) bool) bool {
+	switch canonicalHeaderKey {
+	case "Authorization", "Www-Authenticate", "Cookie", "Cookie2":
```
pdaures: Is it Www-Authenticate or WWW-Authenticate?

zachgersh:
I've actually decided to close this. I feel like adding this additional knob doesn't seem like it will impact that many people and the current default behavior is probably fine as is.

PR #24123, opened 2018-02-26
net/http,doc: use HTTP status code constants

KimMachineGun:
There are a few places where the integer value is used.
Use the equivalent constants to improve readability.


DEATH-TROOPER:
I signed it!

PR #24148, opened 2018-02-27
encoding/asn1: update function "isPrintable" to accept '@' character as printable

lotus-wu:
As we know , email address will apear in x.509 certificate, when parse email address ,we need to accept '@' character as printable string.


andybons:
Hi @lotus-wu,
You’ll need to sign the CLA before we can proceed with this review.

Thanks

lotus-wu:
Hi @andybons , I signed it today!

PR #24234, opened 2018-03-04
encoding/json: Fix for #21092 - Zero out target before decoding

### LINKED ISSUE #21092: encoding/json: unmarshal into slice reuses element data between len and cap
go1.8

GOOS=nacl
GOARCH=amd64p32

### What did you do?
https://play.golang.org/p/lbYUhgOe--

According to the [json Unmarshal docs](https://golang.org/pkg/encoding/json/#Unmarshal), Unmarshal resets the slice length to zero and then appends each element to the slice.

### What did you see instead?
Instead Unmarshal resets the slice length to zero and then modifies the elements of the underlying slice.

JeremyLoy:
Fixes #21092

[playground example demonstrating zero technique](https://play.golang.org/p/9j7KTFg8OVs)
[playground example demonstrating bug](https://play.golang.org/p/NHo3ywlPZli)

### DIFF
--- a/src/encoding/json/decode.go
+++ b/src/encoding/json/decode.go
@@ -175,6 +175,13 @@ func (d *decodeState) unmarshal(v interface{}) error {
 
 	d.scan.reset()
 	d.scanWhile(scanSkipSpace)
+
+	// Set rv to a new zero value of its underlying type.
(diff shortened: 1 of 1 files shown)


JeremyLoy:
I signed it!

PR #24264, opened 2018-03-06
regexp: implement look-behind assertion

makenowjust:
Look-behind assertion is one of the regular expression extension.
https://www.regular-expressions.info/lookaround.html

This implements positive look-behind assertion (?<=expr) and negative
look-behind assertion (?<!expr).
It supports variable-length look-behind.
(Variable-length look-behind is also supported by V8 and .NET.)

To emulate look-behind assertion, it runs the main regexp
automaton and some look-behind automata in parallel, and
an automaton refers look-behind automaton state when
the automaton encounters look-behind assertion.
Noteworthy point is it reads input string only once even if
regexp contains look-behind. It is unique feature.

It does not support captures in look-behind
because the meaning of captures in look-behind is unknown.
(and implementing is so hard, hehe ;)

I believe additional cost of matching the regexp without look-behind
is little.


MakeNowJust:
I signed it!

agnivade:
Following Russ' comment, it doesn't look like this is the right way. I will go ahead and close the PR. Please do feel free to discuss this in a new issue on how to better implement this. Thank you.

PR #24307, opened 2018-03-08
os: enable symlink creation on Windows 10

### LINKED ISSUE #22874: os: Symlink creation should work on Windows without elevation
On Windows 10 Creators Update, with Developer Mode enabled, symlinks can be created by regular users without elevation. This can be tested by opening a command prompt and running `mklink a b` for example. However, creating links using `os.Symlink` in Go 1.9.2 windows/amd64 fails with `A required privilege is not held by the client.`

The cause seems to be that the application needs to pass the `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` flag to `CreateSymbolicLink` in order to be able to create symbolic links without elevation. I can't exactly follow the reasoning behind introducing and requiring this flag, but it seems like we should probably set it in the `os.Symlink` implementation for Windows.

https://msdn.microsoft.com/en-us/library/windows/desktop/aa363866(v=vs.85).aspx

Given that this flag exists at all we should probably expect further bizarre stuff like the call failing if we *are* privileged and passing that flag, or if we are privileged and on a Windows version where the flag doesn't exist yet, or whatever. So it probably requires more careful testing than just slapping on the flag and seeing it succeed...

fkollmann:
Fixes #22874


fkollmann:
More information on the topic is provided by Git on Windows:

https://github.com/git-for-windows/git/wiki/Symbolic-Links#allowing-non-administrators-to-create-symbolic-links

fkollmann:
Empty comment for triggering CLA check.

PR #24322, opened 2018-03-09
runtime/cgo: Add initial NetBSD Thread Sanitizer support

krytarowski:
Recognize NetBSD in:
 - go/internal/work/init.go
 - race.bash
 - runtime/race/race.go

Add __ps_strings symbol in runtime/cgo/netbsd.go as this is
used internally in the TSan library for NetBSD and used for
ReExec().

Tested on NetBSD/amd64 v. 8.99.12.

Around 98% tests are passing for the ./race.bash target.


krytarowski:
NetBSD 8.0 or later needed. Older versions might work, but I won't invest into them. 8.0 is already matched as the minimum version recommended for Go.

krytarowski:
CC: @dvyukov

`./race.bash` gives something like:

```
Passed 340 of 347 tests (97.98%, 0+, 7-)
0 expected failures (0 has not fail)
```

http://netbsd.org/~kamil/llvm/go-tsan-2017-12-06.txt

I'm using a locally generated syso file for NetBSD/amd64:  http://netbsd.org/~kamil/llvm/race_netbsd_amd64.syso

I assume that it has to be added by upstream maintainers.

dvyukov:
There is one failure related to exit status, which is I guess a plain bug somewhere and needs to be fixed.
For the remaining failures I don't have a good explanation, because core race detection logic and compiler instrumentation are platform-independent. Are these failures flaky? If you run it 10 times do you see the same set of failures each time?

Re syso file, yes, somebody needs to regenerate all of them and add netbsd.
The official syso files are created with:
https://github.com/golang/build/blob/master/cmd/racebuild/racebuild.go
we need to add netbsd support there (probably mostly copy-paste freebsd part).
Filed #24354 for this.

krytarowski:
Few tests are indeed racy.

In this case I follow "Pareto principle" and move on upstreaming my local changes, as more than 95% of tests report success. In future I plan to be back to sanitizers in LLVM and correct remaining bugs and enable the execution of regression tests on LLVM/NetBSD buildbot. It will be followed by squashing the Go/TSan bugs.

It will be easier to fix the remaining bugs having TSan/NetBSD in the Go sources.

dvyukov:
As you probably saw in llvm tsan tests we use "tsan-invisible-barrier" in tests, which effectively turns them into non-parallel/non-concurrent (though, tsan runtime still thinks that they are parallel and detects races). That eliminates flakiness due to non-determinism and timings. It would be good to have the same for go tests as well.

krytarowski:
I don't have strong opinions, but if we will support it - I will make sure to handle it correctly [in future].

petkatedinev:
Да

krytarowski:
I will have a look!

krytarowski:
Results for:

```
$ ./all.bash
$ ./race.bash
```

and a local diff to enable `-race` for NetBSD/amd64 originally proposed in https://github.com/golang/go/pull/24322

http://netbsd.org/~kamil/golang/golang-amd64-2018-06-03-race-HEAD.txt

```
$ uname -a
NetBSD chieftec 8.99.19 NetBSD 8.99.19 (GENERIC) #6: Thu May 31 02:36:15 CEST 2018  root@chieftec:/public/netbsd-root/sys/arch/amd64/compile/GENERIC amd64
```

krytarowski:
Tested with:

```
$ md5 ./runtime/race/race_netbsd_amd64.syso                                                                               
MD5 (./runtime/race/race_netbsd_amd64.syso) = 3bff62c1ae46943d4d6aedbbde4b4c51
```

krytarowski:
For the record, C/C++ version of TSan (from HEAD of LLVM) for NetBSD/amd64:

```
$ check-tsan

  Expected Passes    : 288
  Expected Failures  : 1
  Unsupported Tests  : 84
  Unexpected Failures: 8
```

krytarowski:
https://github.com/golang/go/commit/8994444491d984491d8070ed30aeffc1abe344a2

ping?

krytarowski:
ping

krytarowski:
(I've used github's online conflict resolver.. hopefully everything is fine).

krytarowski:
@bsiegert please help to upstream this, thank you in advance!

PR #24424, opened 2018-03-16
fixes #24422 leading double quote in non-quoted \t separated field

### LINKED ISSUE #24422: encoding/csv leading " in non-quoted field leads to error

1.9.2

yes

darwin
amd64

### What did you do?

If possible, provide a recipe for reproducing the error.
A complete runnable program is good.
A link on play.golang.org is best.

https://play.golang.org/p/nrG_caEeYL7

A quote `"` that appears at the beginning of the field but not at the end of the field `"Medaka Box" Stray Bushiroad Storage Box Collection Vol.50` leads to an erroneous error: `record on line 2: wrong number of fields`, even though this file has a uniform number of tab `'\t'` delimited fields for every row.  This file parses correctly in Excel, OpenOffice, Numbers etc.  The data comes from a legitimate source, generated tab separated file from an Amazon report.

If I turn magic quotes off, then I get a different error: `parse error on line 2, column 11: extraneous or missing " in quoted-field`

I expect not to see an error.  A partially quoted field in a non-quoted field is a legit use case that other systems/software generate and can consume.

### What did you see instead?

record on line 2: wrong number of fields

homanchou:
fixes issue https://github.com/golang/go/issues/24422 leading double quote in non-quoted field with \t separator

### DIFF
--- a/src/encoding/csv/reader.go
+++ b/src/encoding/csv/reader.go
@@ -289,7 +289,7 @@ parseField:
 		if r.TrimLeadingSpace {
 			line = bytes.TrimLeftFunc(line, unicode.IsSpace)
 		}
-		if len(line) == 0 || line[0] != '"' || r.Comma == '\t' {
+		if len(line) == 0 || line[0] != '"' || r.FieldsPerRecord == bytes.Count(line, []byte(string(r.Comma)))+1 {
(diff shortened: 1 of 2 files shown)


homanchou:
I signed it!

PR #24528, opened 2018-03-25
net/url: create helper function "SetQueryParameters"

HaraldNordgren:


HaraldNordgren:
From my first experience with Go, I was confused about the round-about way that we set query parameters:

```go
v := u.Query()
v.Set("a", 1)
v.Set("b", 2)
u.RawQuery = v.Encode()
```

so I though I would create a helper function to hopefully make life a tiny bit easier for other developers.

HaraldNordgren:
Closing this after https://github.com/golang/go/issues/24552 was closed.

89z:
I didnt care for this implementation, as I would like to be able to set the
parameters one by one. So I did this instead:

~~~go
type Uri struct {
   url.URL
   url.Values
}

func NewUri() Uri {
   return Uri{
      Values: url.Values{},
   }
}

func (u *Uri) QuerySet(key, value string) {
   u.Values.Set(key, value)
   u.RawQuery = u.Values.Encode()
}
~~~

https://pkg.go.dev/github.com/89z/x@v1.6.0#Uri

PR #24805, opened 2018-04-11
42comment

jy451:
the Gerrit change.

  + No trailing period
+ No Markdown


mvdan:
Opened by mistaske? There's nothing interesting here that I can see.

PR #24868, opened 2018-04-15
cmd/go/internal/load: assign .ImportPath and .Root for a testdata package in GOPATH properly

linzhp:
go/build.Import failed to assign values .ImportPath and .Root if a package is a local import and is a sub package of testdata. Then in cmd/go/internal/load.VendoredImportPath, the vendored import path is not expanded if the parent.Root is empty. Without the "vendor" part in the parent package, the check `filepath.Join(root, parent.ImportPath) != dir` in cmd/go/internal/load.VendoredImportPath succeeds in child imports, leading to #24854.

This PR makes go/build assign .ImportPath and .Root for a testdata package in GOPATH properly, and fixes #24854.

### DIFF
--- a/src/cmd/go/go_test.go
+++ b/src/cmd/go/go_test.go
@@ -6042,3 +6042,28 @@ func TestNoRelativeTmpdir(t *testing.T) {
 		tg.grepStderr("relative tmpdir", "wrong error")
 	}
 }
+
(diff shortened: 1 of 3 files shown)


linzhp:
My company added me to the authorized contributor Google group.  Could you verify?

PR #24896, opened 2018-04-17
make container/ring.Move method better

MaruHyl:
i think this change will make the method-(container/ring.Move) better , and ring_test.go passed , please review that


MaruHyl:
I signed it!

MaruHyl:
I signed it!

PR #24960, opened 2018-04-20
cmd/go/internal/version: fixes version arg

adamgj:
Fixes https://github.com/adamgj/go/issues/1

Expected `go -v` to return version

### DIFF
--- a/src/cmd/go/internal/version/version.go
+++ b/src/cmd/go/internal/version/version.go
@@ -14,7 +14,7 @@ import (
 
 var CmdVersion = &base.Command{
 	Run:       runVersion,
-	UsageLine: "version",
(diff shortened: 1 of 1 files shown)


adamgj:
I signed it!

agnivade:
This is a breaking change which breaks [Go1 compatiblity](https://golang.org/doc/go1compat). Programs around the world printing `go version` and parsing the version string will break.

For any breaking change, a proposal issue needs to be opened and accepted first.

agnivade:
@adamgj - As pointed out by Ian, this is not correct and will break existing programs. If you want `go -v` to return the version, please open a new proposal issue explaining your rationale.

PR #25025, opened 2018-04-23
net/http: explain more how Pusher.Push works

akyoto:
This will clarify that the resources are not completely pushed yet when `Push` returns and that it starts a separate goroutine. This might be implementation dependant but as I believe there is currently only one implementation it should be added to the documentation of the interface which most people will look up first.


blitzprog:
I signed it!

ahmadmiz (approved):

PR #25058, opened 2018-04-25
doc: fix a potential mistake about struct embedding

lifuhuang:
The original sentence does not make sense with the other parts in the same paragraph. It's true that it would also "be erroneous to embed log.Logger if the Job struct contained another field or method called Logger", but it would be a compilation error no matter whether the duplicate name is mentioned in the program outside the type definition or not. I think most probably the author was originally trying to express what I wrote, e.g. when a struct embeds two structs that have conflicting names (such as Command), it is OK if duplicate name is never mentioned in the program outside the type definition.


lifuhuang:
I signed it!

PR #25130, opened 2018-04-27
Remove "Setting GOROOT" custom install instruction

### LINKED ISSUE #25002: doc: setting GOROOT is not necessary when installing to custom location

go version go1.10.1 darwin/amd64

Yes.

### What did you do?

1. Visit https://golang.org/dl/ and download a go1.10.1 tarball
2. Install to a custom location:
    ```
    $ cd /tmp
    $ tar -xzf ~/Downloads/go1.10.1.darwin-amd64.tar.gz
    $ go/bin/go env GOROOT
    /tmp/go
    ```

3. Visit https://golang.org/doc/install#tarball_non_standard and see the installation instructions

No need to set GOROOT, as it is indeed [not necessary in Go 1.10](https://golang.org/doc/go1.10#goroot).

### What did you see instead?

> For example, if you installed Go to your home directory you should add commands like the following to $HOME/.profile:
> 
> ```
> export GOROOT=$HOME/go1.X
> export PATH=$PATH:$GOROOT/bin
> ```
> 
> Note: GOROOT must be set only when installing to a custom location.

---

In https://github.com/golang/go/issues/18678#issue-201154131, @minux proposed to derive GOROOT from `os.Executable`...

> And then we can eliminate all mentions of GOROOT in user facing documents (in fact, we should actively discourage setting GOROOT in user facing documents.)

The change was implemented in [CL 42533](https://go-review.googlesource.com/42533), but the documentation was not updated to remove suggestions to set GOROOT.

JoeCortopassi:
doc: Setting GOROOT is no longer necessary for custom installation as of 1.10 (https://go-review.googlesource.com/c/go/+/42533)

Fixes https://github.com/golang/go/issues/25002

the Gerrit change.

  + No trailing period
+ No Markdown

### DIFF
--- a/doc/install.html
+++ b/doc/install.html
@@ -114,31 +114,6 @@ or execute them from the profile using a command such as
 <code>source $HOME/.profile</code>.
 </p>
 
-<h4 id="tarball_non_standard">Installing to a custom location</h4>
-
-<p>
-The Go binary distributions assume they will be installed in
-<code>/usr/local/go</code> (or <code>c:\Go</code> under Windows),
-but it is possible to install the Go tools to a different location.
-In this case you must set the <code>GOROOT</code> environment variable
-to point to the directory in which it was installed.
-</p>
-
-<p>
-For example, if you installed Go to your home directory you should add
-commands like the following to <code>$HOME/.profile</code>:
-</p>
-
-<pre>
-export GOROOT=$HOME/go1.X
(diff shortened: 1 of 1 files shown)


agnivade:
You should abandon [golang.org/cl/109955](https://go-review.googlesource.com/c/go/+/109955).

PR #25154, opened 2018-04-29
encoding/json: include property name for marshal errors

### LINKED ISSUE #25153: encoding/json: include property name for marshal errors

```
$ go version
go version go1.10 linux/amd64
```

Yes

```
$ go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/dominic/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/dominic/dev"
GORACE=""
GOROOT="/usr/lib/go-1.10"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go-1.10/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build203824871=/tmp/go-build -gno-record-gcc-switches"
```

### What did you do?

When encountering an error using `json.Marshal()` (and related functions) the error returned gives no indication where the error was encountered. The only included metadata is about the type of the value that was being encoded, but that is usually insufficient to narrow down problems without the original input value. (which may not always be at hand)

### What is being proposed?

The proposal would be to add the property name for marshal errors (using dot notation to signify hierarchy). For example:

> json: error calling MarshalJSON for type time.Time: Time.MarshalJSON: year outside of range [0,9999]

would be updated to be more specific:

> json: error calling MarshalJSON for timestamp (type time.Time): Time.MarshalJSON: year outside of range [0,9999]

and in the case of a more complex/nested data structure:

> json: error calling MarshalJSON for events.[0].timestamp (type time.Time): Time.MarshalJSON: year outside of range [0,9999]

### Why is this useful?

I use Go to do a lot of data processing, which is generally from 3rd-parties (aka: customers) so the structure of data is not usually known ahead of time, not to mention weird values being sent due to programming errors and a host of other problems.

Last week, I encountered the first error message above and needed to do some extra work just to determine which property was causing the error, when the encoder should be able to communicate that to the caller.

dominicbarnes:
When a marshal error is encountered, the error message (and underlying `MarshalerError`) does not include any context about where the error was encountered, even if the value is complex like a map, slice or struct.

This change will include that metadata in both places, allowing a caller to have more context in the event of a JSON marshal error.

Fixes #25153

### DIFF
--- a/src/encoding/json/encode.go
+++ b/src/encoding/json/encode.go
@@ -695,7 +695,7 @@ func (me *mapEncoder) encode(e *encodeState, v reflect.Value, opts encOpts) {
 	for i, v := range keys {
 		sv[i].v = v
 		if err := sv[i].resolve(); err != nil {
-			e.error(&MarshalerError{v.Type(), v, opts.nest(v.String()).path, err})
+			e.error(&MarshalerError{v.Type(), v, append(opts.path, v.String()), err})
 		}
 	}
 	sort.Slice(sv, func(i, j int) bool { return sv[i].s < sv[j].s })
--- a/src/encoding/json/encode.go
+++ b/src/encoding/json/encode.go
@@ -783,7 +783,7 @@ func (ae *arrayEncoder) encode(e *encodeState, v reflect.Value, opts encOpts) {
 		if i > 0 {
 			e.WriteByte(',')
 		}
-		ae.elemEnc(e, v.Index(i), opts.nest("["+strconv.Itoa(i)+"]"))
+		ae.elemEnc(e, v.Index(i), opts.nest(strconv.Itoa(i)))
 	}
 	e.WriteByte(']')
 }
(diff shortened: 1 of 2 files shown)


dominicbarnes:
I'm contemplating also including the invalid value on the `MarshalerError` struct (but not in the `Error()` message). The main thing I want to avoid is adding an unnecessary amount of overhead, but to make errors easier to act upon, maybe that would be reasonable?

dominicbarnes:
I added `MarshalerError.Value reflect.Value` to the PR as a separate commit for consideration. (if approved, I will squash it into the other commit) With this, it would be possible to check `err.(*json.MarshalerError)`, extract the path to the error and the value that was being encoded.

salif (approved):

PR #25182, opened 2018-04-30
http: Capitalization of HTTP 418 I'm a Teapot

mithunsasidharan:
In [rfc2324](https://tools.ietf.org/html/rfc2324) (1998), the HTTP 418 message reads I'm a teapot, whereas in the latter [rfc7168](https://tools.ietf.org/html/rfc7168) (2014) it's spelled I'm a Teapot — note that capital T.

This is consistent with other canonical status messages, all of which have Every Word Capitalized.


mithunsasidharan:
I signed it!

PR #25193, opened 2018-05-01
debug/pe: parse the import directory correctly

### LINKED ISSUE #16103: debug/pe: ImportedSymbols returns no symbols
Running this program:

```
package main

import (
    "debug/pe"
    "fmt"
    "os"
)

func main() {
    for _, v := range os.Args[1:] {
        f, err := pe.Open(v)
        if err != nil {
            fmt.Fprintf(os.Stderr, "can not create PE structure for %s: %v\n", v, err)
            continue
        }
        defer f.Close()

        libs, err := f.ImportedSymbols()
        if err != nil {
            fmt.Fprintf(os.Stderr, "can not get imported symbols: %v\n", err)
            continue
        }
        for _, v := range libs {
            fmt.Printf("%s\n", v)
        }
    }
}
```

with [PsService.zip](https://github.com/golang/go/files/318935/PsService.zip) file as argument, lists no imported dlls / functions. But it should:

```
$ objdump -x PsService.exe | sed '/The Import Tables/,$!d' | head
The Import Tables (interpreted .rdata section contents)
 vma:            Hint    Time      Forward  DLL       First
                 Table   Stamp     Chain    Name      Thunk
 00026aa4       00026e18 00000000 00000000 00026e7e 000202ac

        DLL Name: VERSION.dll
        vma:  Hint/Ord Member-Name Bound-To
        26e4e       0  GetFileVersionInfoA
        26e64       3  GetFileVersionInfoSizeA
        26e3c      13  VerQueryValueA
$
```

The reason for this is because debug/pe assumes that imported symbols live in section named ".idata", but it is not always true (as you can see from objdump output). We should use PE "The Data Directory" to determine location of imported info.

Alex

arizvisa:
This parses the import table properly which allows for debug/pe
to extract import symbols from pecoffs linked with an import
table in a section named something other than ".idata"

The section names in a pecoff object aren't guaranteed to actually
mean anything, so hardcoding a search for the ".idata" section
is not guaranteed to find the import table in all shared libraries.
This resulted in debug/pe being unable to read import symbols
from some libraries.

The proper way to locate the import table is to validate the
number of data directory entries, locate the import entry, and
then use the va to identify the section containing the import
table. This patch does exactly this.

Fixes #16103.


arizvisa:
signed contributor's license agreement.

PR #25195, opened 2018-05-01
encoding/asn1: allow Marshaling and Unmarshaling private tag class

hamano:
ASN.1 has an private class, but current implementation does not support it.


hamano:
I signed CLA.

PR #25291, opened 2018-05-08
cmd/dist: explicit option for crosscompilation

### LINKED ISSUE #25177: cmd/dist: cross-libc same os/arch build failing since go 1.10

go version go1.10.1 linux/amd64

I also have reproduced this with the master branch from a few days ago.

Yes

GOARCH="amd64"
GOBIN=""
GOCACHE="/home/anisse/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/anisse/bin/prefix/gopath"
GORACE=""
GOROOT="/home/anisse/dev/buildroot/musl_min/host/lib/go"
GOTMPDIR=""
GOTOOLDIR="/home/anisse/dev/buildroot/musl_min/host/lib/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="/usr/bin/gcc"
CXX="/usr/bin/g++"
CGO_ENABLED="0"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build290222412=/tmp/go-build -gno-record-gcc-switches"

### What did you do?

Since go 1.10, cross-building go for a different libc on the same arch and os, stopped working. This is due to the new build caching infrastructure calling "go list" during the build. 

Technically, go made the decision long ago (why?) to always use $CC even if $CC_FOR_TARGET is available when gohostos=goos and gohostarch=goarch. This worked until the new go caching infrastructure changed way go is built, and "go list" started to be called during build.

Note: obviously, this only happens when building with CGO_ENABLED=1.

I was able to workaround this by removing the legacy build behaviour:
https://github.com/golang/go/blob/master/src/cmd/dist/build.go#L258-L260

Building go works.

### What did you see instead?

Building go failed.

This can be seen in the buildroot autobuild logs for musl and uclibc x86_64 targets:
http://autobuild.buildroot.net/results/2ba/2ba1eee95bdad910549ef64fc9a075d2136b4936/build-end.log

It tries calling "go list" built using the cross-libc compiler and fails with "no such file or directory" since it's not the same ld linker.

You can find other uclibc/musl failed logs here:
http://autobuild.buildroot.net/?start=100&reason=host-go%

As well as the buildroot host-go package:
https://git.buildroot.net/buildroot/tree/package/go/go.mk (requires knowledge of buildroot).

angeloc:
If GOHOSTOS == GOOS || GOHOSTARCH == GOARCH the go build system assume
it's not cross compiling and uses the same compiler for both
CC_FOR_TARGET and CC even if they are declared different. During go
compilation, this produces the error:

fork/exec
/accts/mlweber1/rclinux/rc-buildroot-test/scripts/instance-2/output/build/host-go-1.10/bin/go:
no such file or directory

cause the go binary produced is linked against a different libc and
cannot be run on the host machine.

This is a problem in case the cross compilation mandates a different
toolchain for host and target like happens in cross compilation
environments like buildroot. This patch adds GO_ASSUME_CROSSCOMPILING
varible to assure that in case of cross compilation CC_FOR_TARGET can be
different from CC.

Fixes #25177

### DIFF
--- a/src/cmd/dist/build.go
+++ b/src/cmd/dist/build.go
@@ -292,12 +292,13 @@ func xinit() {
 // $CC_FOR_goos_goarch, if set, applies only to goos/goarch.
 func compilerEnv(envName, def string) map[string]string {
 	m := map[string]string{"": def}
+	crosscompiling := os.Getenv("GO_ASSUME_CROSSCOMPILING")
 
 	if env := os.Getenv(envName); env != "" {
 		m[""] = env
 	}
 	if env := os.Getenv(envName + "_FOR_TARGET"); env != "" {
-		if gohostos != goos || gohostarch != goarch {
+		if gohostos != goos || gohostarch != goarch || crosscompiling == "1" {
 			m[gohostos+"/"+gohostarch] = m[""]
 		}
 		m[""] = env
--- a/src/cmd/go/internal/help/helpdoc.go
+++ b/src/cmd/go/internal/help/helpdoc.go
@@ -682,6 +682,9 @@ Additional information available from 'go env' but not read from the environment
 		The directory where the go tools (compile, cover, doc, etc...) are installed.
 	GOVERSION
 		The version of the installed Go tree, as reported by runtime.Version.
+	GO_ASSUME_CROSSCOMPILING
+		Force the go build system to honor the different CC_FOR_TARGET even
+		if the architecture is the same, useful when cross compiling.
 	`,
 }
 
--- a/src/make.bash
+++ b/src/make.bash
@@ -72,6 +72,10 @@
 # that name even though they put newer Go toolchains there.
 
 bootgo=1.20.6
+#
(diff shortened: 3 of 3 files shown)


ahmadmiz (approved):

angeloc:
May I know why this was closed?

heschi:
I closed PRs older than a year to reduce load on the Gerrit importer. If you plan to continue work on it I'm happy to reopen, though I now realize that will create a new CL. Sorry for the trouble.

angeloc:
We have this patch in buildroot to guarantee a correct cross compilation, so I would like to keep it open waiting for a final decision.

Issue #25506, opened 2018-05-22
runtime: implement timers, JavaScript callbacks back into WebAssembly


bradfitz:
After some discussion last week with @cherrymui, @rsc, @aclements, et al, we decided that users would be really sad if Go 1.11 was released with WebAssembly support and then realized they couldn't really use it in a browser well, without working timers & callbacks back into Go.

@cherrymui did a little prototype of callbacks and it was enough to show a demo of a Go webassembly program updating images in the DOM and responding to button clicks.

So we decided it was worth spending a bit of time and making sure the wasm support is what users will expect for Go 1.11.

@neelance, what's your availability look like to finish that in the next month or two?

I see you updated your branch (https://github.com/neelance/go/commits/wasm-wip) 3 days ago with https://github.com/neelance/go/commit/773eb4660f4aa163becf70b55d967c879029c2ab

Do you need any help? @cherrymui and perhaps @aclements could help if needed.

Please use this issue number in any related CLs. Thanks!

neelance:
Thanks for offering the help! Callbacks are already working quite well. I just cleaned up the commit some more and pushed the CL for initial feedback/review: https://go-review.googlesource.com/c/go/+/114197

Do we also want to use this to get support for `http.Transport` into this release? There is already a POC at https://github.com/johanbrandhorst/fetch.

bradfitz:
An `http.DefaultTransport` value implementing `http.RoundTripper` I assume? SGTM for Go 1.11. Let's do it.

If the author wants to send it via a PR or CL I can review. (PRs get converted into Gerrit these days)

shurcooL:
I'm happy to offer help with implementing and/or reviewing here as well.

I'm the author of the original Fetch API-based `http.RoundTripper` implementation [in GopherJS](https://github.com/gopherjs/gopherjs/pull/454) that https://github.com/johanbrandhorst/fetch is based on, so I'm very familiar with how it works.

johanbrandhorst:
@shurcooL I'll try and get something in as soon as I can. I'm thinking we should implement the XHR transport as well and maybe make that the default as even among browsers that support WASM, they don't all support Fetch out of the box.

neelance:
`fetch` can be polyfilled and services like https://polyfill.io/ make it trivially easy to do so efficiently. I nowadays always prefer the newer (stable) APIs + polyfills.

bradfitz:
[...]

PR #25550, opened 2018-05-24
net/http: add js/wasm compatible DefaultTransport

johanbrandhorst:
Adds a new Transport type for the js/wasm target that uses the
JavaScript Fetch API for sending HTTP requests. Support for
streaming response bodies is used when available, falling back
to reading the entire response into memory at once.

Updates #25506


johanbrandhorst:
I've tried to push this up sooner rather than later to start discussing the implementation. I am aware the tests probably do not pass yet, but I will work with @neelance and @shurcooL to get my local environment up to scratch.

Issue #26022, opened 2018-06-23
go/ast: provide test for Filter func


arthurkushman:
### What version of Go are you using (`go version`)?
go version go1.9.2 linux/amd64

### What operating system and processor architecture are you using (`go env`)?
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/srv/go"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build760291292=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"

### What did you do?
Add TestFilter function to test comments after filtering

### What did you expect to see?
TestFilter function runs with all checks passed

PR #26023, opened 2018-06-23
go/ast: add Filter func test on AST

arthurkushman:
The existing implementation of Filter should behave as if it removes (filter out) comments from parsed AST (by providing filter Node argument) leaving the rest untouched. 

The TestFilter - checks for length differences of both filtered and unfiltered as general and then for excluded from ASTree.

Add FilterTest #26022


arthurkushman:
I signed it!

PR #26054, opened 2018-06-25
text/html: MIME type "application/ld+json: should be escaped as JS

### LINKED ISSUE #26053: html/template: "application/ld+json" MIME type is not understood as JS
The text/html package does not understand the "application/ld+json" MIME type to be JS. This is a standard format endorsed by the W3C and various search engines. Search engines like Google require this data to be inline in a page's HTML using a tag like `<script type="application/ld+json">JSON DATA HERE</script>`, so it is important for the text/html template engine to be aware of how to escape JSON-LD data, as the data cannot be placed in an external script file. For more information on JSON-LD see [1] and [2].

I plan to submit a trivial PR to add json+ld support to text/html, but I am creating a GitHub issue first to describe the issue as per the contribution policy.

[1] https://en.wikipedia.org/wiki/JSON-LD
[2] https://developers.google.com/search/docs/guides/intro-structured-data

eklitzke:
Teach the text/html package to understand the "application/ld+json" MIME
type (used for JSON-LD data) as JS, so it can be apply the correct contextual
escaping.

Fixes #26053

### DIFF
--- a/src/html/template/js.go
+++ b/src/html/template/js.go
@@ -391,6 +391,7 @@ func isJSType(mimeType string) bool {
 		"application/ecmascript",
 		"application/javascript",
 		"application/json",
+		"application/ld+json",
 		"application/x-ecmascript",
 		"application/x-javascript",
 		"text/ecmascript",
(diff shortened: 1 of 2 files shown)


agnivade:
@eklitzke - Wanted to ping you here since I did not see you copied in the gerrit CL. Would you be able to take a look at Filippo's comments ? Thanks !

PR #26271, opened 2018-07-08
runtime: fix abort() for ppc64x

zeldin:
The test case "Abort" expects a SIGTRAP.  Before this change, abort()
would generate a SIGSEGV instead.

This of course begs the question why abort() is expected to generate a
SIGTRAP and not a SIGABRT, but I'm sure you have some profound reason
for this.  :-)


zeldin:
I signed it.  Let's see if I need to change the email address in the commits too, or if the match on git user id works...

PR #26516, opened 2018-07-21
doc/contribute: add examples for finding issues on the issue tracker

leighmcculloch:
For contributors looking for new issues to contribute to it can be
difficult to find issues that need a fix and don't already have a fix
being considered. There are several labels that help guide the way
already, like `NeedsFix`, `HelpWanted`. But many issues with this label
will already have a CL. For new contributors this can be especially
difficult.

Fixes #26494


leighmcculloch:
Screenshot of the area of the page that's changing:

screen shot 2018-07-20 at 6 19 03 pm

PR #26527, opened 2018-07-22
go/token: fix test Example_retrievePositionInfo fail

wgliang:
Example_retrievePositionInfo test fail in go/token.


wgliang:
I'm sorry, it's my env wrong.

PR #26604, opened 2018-07-25
crypto/tls: delete session tickets on TLS handshake failure

### LINKED ISSUE #24919: crypto/tls: clients don't delete tickets that fail handshakes from ClientSessionCache

1.9.2

Yes

```
go version go1.9.2 linux/amd64
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/ubuntu/src/go"
GORACE=""
GOROOT="/usr/lib/go-1.9"
GOTOOLDIR="/usr/lib/go-1.9/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build198947136=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOROOT/bin/go version: go version go1.9.2 linux/amd64
GOROOT/bin/go tool compile -V: compile version go1.9.2
uname -sr: Linux 4.13.0-38-generic
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.3 LTS
Release:	16.04
Codename:	xenial
/lib/x86_64-linux-gnu/libc.so.6: GNU C Library (Ubuntu GLIBC 2.23-0ubuntu10) stable release version 2.23, by Roland McGrath et al.
gdb --version: GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
```

### What did you do?

https://play.golang.org/p/2qH4SFg2Gjn

Created a TLS client and a server, and tested TLS session resumption behavior.

If a server accepts a ticket sent by the client but the handshake fails, the client SHOULD delete the ticket from its cache, according to Section 3.2 of RFC 5077. https://tools.ietf.org/html/rfc5077. The next connection to the server should fallback to a full TLS handshake.

### What did you see instead?

If a client and server attempt to resume a TLS session past the expiry of the client certificate embedded in the ticket, the client gets wedged. The server accepts the ticket, but handshake fails because of the expired certs. The ticket is not deleted from the cache however, so subsequent attempts by the client to connect to the server fail in the same way.

santoshankr:
When a server accepts a session ticket presented by a client, but
the TLS handshake fails, RFC 5077 recommends that the client delete
the ticket. Because adding a Delete method to the interface for
ClientSessionCache would break existing implementations, we have the
handshake implementation overwrite the ticket with a nil instead.

Fixes #24919

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -853,15 +853,21 @@ func NewLRUClientSessionCache(capacity int) ClientSessionCache {
 	}
 }
 
-// Put adds the provided (sessionKey, cs) pair to the cache.
+// Put adds the provided (sessionKey, cs) pair to the cache. If cs is nil, the entry
+// corresponding to sessionKey is removed from the cache instead.
 func (c *lruSessionCache) Put(sessionKey string, cs *ClientSessionState) {
 	c.Lock()
 	defer c.Unlock()
 
 	if elem, ok := c.m[sessionKey]; ok {
-		entry := elem.Value.(*lruSessionCacheEntry)
-		entry.state = cs
-		c.q.MoveToFront(elem)
+		if cs == nil {
+			c.q.Remove(elem)
+			delete(c.m, sessionKey)
+		} else {
+			entry := elem.Value.(*lruSessionCacheEntry)
+			entry.state = cs
+			c.q.MoveToFront(elem)
+		}
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


santoshankr:
CLA signed!

Issue #26628, opened 2018-07-26
cmd/go: use quiet flag to allow tool integrations with Go modules


marwan-at-work:
### What version of Go are you using (`go version`)?
Tip - Darwin

### What did you do?
`go list -m -versions -json github.com/pkg/errors@latest` to list all the available upstream versions

### What did you expect to see?
I'd like to **only** get the JSON output and not mod-fetching output. 
The boolean QuietLookup in modfetch/cache.go is not being set anywhere which is causing the following to show

### What did you see instead?

```
go: finding github.com/pkg/errors v0.8.0
{
	"Path": "github.com/pkg/errors",
	"Version": "v0.8.0",
	"Versions": [
		"v0.1.0",
		"v0.2.0",
		"v0.3.0",
		"v0.4.0",
		"v0.5.0",
		"v0.5.1",
		"v0.6.0",
		"v0.7.0",
		"v0.7.1",
		"v0.8.0"
	],
	"Time": "2016-09-29T01:48:01Z"
}
```

myitcv:
Please can you confirm the full env details here (you've abbreviated the template) - what commit were you using etc?

I can't reproduce the issue you're seeing based on the steps above either.

In any case, stdout should always contain valid JSON, per https://github.com/golang/go/issues/26218#issuecomment-402916995

marwan-at-work:
@myitcv I believe I was parsing both stderr and stdout, so I'll try again with just stdout and it should probably work. That being said, maybe the CL is still ok to have control over the QuietLookup variable in cache. Otherwise, using stdout is fine for my usecase. Thanks

myitcv:
@marwan-at-work - I'm going to close per https://github.com/golang/go/issues/26628#issuecomment-409279217, but please shout if you're seeing behaviour contrary to https://github.com/golang/go/issues/26218#issuecomment-402916995

PR #26631, opened 2018-07-26
cmd/go/internal/list: use quiet flag to enable json only output

marwan-at-work:
Fixes https://github.com/golang/go/issues/26628


zegl:
@bradfitz How come this PR has not been imported to Gerrit?

marwan-at-work:
@zegl seems that it's in Gerrit https://go-review.googlesource.com/c/go/+/126235

PR #26646, opened 2018-07-27
database/sql: allow Scan to work with structs

moltzaum:
The default behavior of structs is now to sequentially scan the struct fields.
If some of the fields are unexported, the new implementation will defer to the
old behavior. This change does not affect any types that already implement the
Scanner interface.

Testing for embedded struct added.

Updates #22697 (database/sql wishlish)
Fixes jmoiron/sqlx#162


moltzaum:
I signed it!

PR #26674, opened 2018-07-29
path: present the correct error message in TestMatch

johngibb:
The message was hardcoded to indicate that the test wanted a nil error, even
though in some cases a specific error was wanted. This patch fixes the
message to print the wanted error.


johngibb:
I signed it!

PR #26737, opened 2018-08-01
all: use buf.String() instead of string(buf.Bytes()) in tests

as:


bradfitz:
Seems like this timed out.

Issue #26769, opened 2018-08-02
No way to set options when making outgoing http calls WASM 


eginez:
Please answer these questions before submitting your issue. Thanks!

### What version of Go are you using (`go version`)?
go version devel +c814ac44c0 Thu Jul 19 21:30:27 2018 +0000 darwin/amd64

### Does this issue reproduce with the latest release?
Yes

### What operating system and processor architecture are you using (`go env`)?
GOARCH="wasm"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="js"
GOPATH="/Users/eginez/repos/goland"
GOPROXY=""
GORACE=""
GOROOT="/Users/eginez/repos/go"
GOTMPDIR=""
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/pm/tlg5p5nx4vj3s05h4rfsv3t40000gn/T/go-build136489492=/tmp/go-build -gno-record-gcc-switches -fno-common"

### What did you do?
 The native http client, maps to the `fetch()` api. With the current client it is impossible to set additional options as supported by the fetch call, see: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch.

### What did you expect to see?
A way to specify options for the fetch call

agnivade:
Yes, this is known. Currently, there is no good way to expose fetch parameters from user level to the `http.RoundTripper` which is what we are using to make the fetch call. It is not very straightforward, as we have to make some decisions on how best to integrate browser options into a Go API.

See a very small discussion here - https://go-review.googlesource.com/c/go/+/114515/8/src/net/http/roundtrip_js.go#34

The main tracking issue for a better wasm http client is https://github.com/golang/go/issues/25695. I would suggest that you comment on that issue with your suggestions on how best we can take this forward.

agnivade:
Ah, using `WithContext` sounds like a good idea. I was overcomplicating things.

[...]

PR #26784, opened 2018-08-03
net/http: support configuring fetch options

johanbrandhorst:
The default WASM RoundTripper is implemented using
the browser Fetch API. Some options don't readily map to
existing http.Request options, so we use the precedent
set by the TrailerPrefix constant to allow a user to configure
the "mode" and "credentials" options by supplying them
as headers in the http.Request.

Updates #26769

### DIFF
--- a/src/net/http/roundtrip_js.go
+++ b/src/net/http/roundtrip_js.go
@@ -17,16 +17,26 @@ import (
 	"syscall/js"
 )
 
+// ModeHeader is a Request.Header map key that, if present,
+// signals that the map entry is actually an option to the Fetch API mode setting.
+// Valid values are: "cors", "no-cors", "same-origin", "navigate"
+// The default is "same-origin".
+//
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
I tested this change with a couple of different settings for `mode` and `credentials` and it seems to work. Hopefully by the time this can be merged (1.12 cycle) we'll have a way to write tests as well.

eginez:
@johanbrandhorst thanks for the quick turnaround on this issue. One quick comment I have on your PR. As a stop gap and until we can rehaul parts of the client, I see no problems with this approach. On the other hand I am really not inclined with the solution as presented if it this is going to be the long term answer. The idea that @bradfitz had about using context seems less magical to me. Thoughts?

johanbrandhorst:
@eginez, again please add feedback to https://go-review.googlesource.com/c/go/+/127718, not here.

PR #26948, opened 2018-08-12
net: halve the allocs in ParseCIDR by sharing slice backing

cstockton:
Share a slice backing between the host address, network ip and mask. Add tests to verify that each slice header has len==cap to prevent introducing new behavior into Go programs. This has a small tradeoff of allocating a larger slice backing when the address is invalid. Earlier error detection of invalid prefix length helps balance this cost and a new benchmark for ParseCIDR helps measure it.

This yields a 22%~ speedup for all nil err cidr tests:
```
  name               old time/op    new time/op    delta
  ParseCIDR/IPv4-24    9.17µs ± 6%    7.20µs ± 7%  -21.47%  (p=0.000 n=20+20)
  ParseCIDR/IPv6-24    9.02µs ± 6%    6.95µs ± 9%  -23.02%  (p=0.000 n=20+20)
  ParseCIDR/IPv4-24    1.51kB ± 0%    1.55kB ± 0%   +2.65%  (p=0.000 n=20+20)
  ParseCIDR/IPv6-24    1.51kB ± 0%    1.55kB ± 0%   +2.65%  (p=0.000 n=20+20)
  ParseCIDR/IPv4-24      68.0 ± 0%      34.0 ± 0%  -50.00%  (p=0.000 n=20+20)
  ParseCIDR/IPv6-24      68.0 ± 0%      34.0 ± 0%  -50.00%  (p=0.000 n=20+20)
```

Including non-nil err cidr tests gains around 25%~:
```
  name               old time/op    new time/op    delta
  ParseCIDR/IPv4-24    11.8µs ±11%     8.9µs ± 8%  -24.88%  (p=0.000 n=20+20)
  ParseCIDR/IPv6-24    11.7µs ± 7%     8.7µs ± 5%  -25.93%  (p=0.000 n=20+20)
  ParseCIDR/IPv4-24    1.98kB ± 0%    2.00kB ± 0%   +1.21%  (p=0.000 n=20+20)
  ParseCIDR/IPv6-24    1.98kB ± 0%    2.00kB ± 0%   +1.21%  (p=0.000 n=20+20)
  ParseCIDR/IPv4-24      87.0 ± 0%      48.0 ± 0%  -44.83%  (p=0.000 n=20+20)
  ParseCIDR/IPv6-24      87.0 ± 0%      48.0 ± 0%  -44.83%  (p=0.000 n=20+20)
```

### DIFF
--- a/src/net/ip.go
+++ b/src/net/ip.go
@@ -75,8 +75,18 @@ func CIDRMask(ones, bits int) IPMask {
 	if ones < 0 || ones > bits {
 		return nil
 	}
-	l := bits / 8
-	m := make(IPMask, l)
+	return putCIDRMask(nil, ones, bits/8)
+}
+
+// putCIDRMask will put an IPMask into `m` consisting of `ones` 1 bits
+// followed by 0s up to a total length of `bits' bits if the length
+// of m is < `l` bytes and returns the same slice m. If m did not have
+// sufficient length for the mask l a new m is returned instead.
+func putCIDRMask(m IPMask, ones, l int) IPMask {
+	if len(m) < l {
+		m = make(IPMask, l)
+	}
+
 	n := uint(ones)
 	for i := 0; i < l; i++ {
 		if n >= 8 {
@@ -243,6 +253,12 @@ func allFF(b []byte) bool {
 
 // Mask returns the result of masking the IP address ip with mask.
 func (ip IP) Mask(mask IPMask) IP {
+	return putIPMask(nil, ip, mask)
+}
+
+// putIPMask will return out when it has sufficient len for the result
+// of masking the IP address ip with mask, otherwise returning a new slice.
+func putIPMask(out, ip IP, mask IPMask) []byte {
 	if len(mask) == IPv6len && len(ip) == IPv4len && allFF(mask[:12]) {
 		mask = mask[12:]
 	}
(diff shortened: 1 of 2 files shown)


cstockton:
Sorry if this pull request is a bit of a mess, I probably should have just used the gerrit workflow with the codereview tool.

cstockton:
I've thought about this a good bit since I've opened it and have decided to withdraw this pull request. I created a library at the time of this pull request for internal use that could ParseCIDR into an existing buffer and fetched [CIDR masks](https://gist.github.com/cstockton/02638ab68bb06f4171da402b3bc8f56f) from a lookup table. The fact I wouldn't stop using it in favor of the stdlib for a pull req I made due to the memory cost just isn't sitting right with me, and I don't think any variation (1, 2, or 3 allocs) will under the constraints of the existing API.

If the Go team would consider an alternative pull request that added some public methods on those types I would be happy to implement an adaption of my patches / internal lib, i.e.:

```Go
// Init is like CIDRMask, but puts the mask into m(.| and returns the possibly extended m.)
func (m *IPMask) Init(ones, bits int)
func (m IPMask) Init(ones, bits int) IPMask

// will reuse the existing IP and IPMask, but removes the ability to get designated host
func (n *IPNet) (ParseCIDR|Parse|InitCIDR|???)(cidr string) error

// you could supplement that with something like the below, but feels a little off.
func (ip *IP) ParseCIDR() error
```

None of my existing thoughts about how to bolt onto the existing API are very elegant. Maybe something else will come to me and I'll open a separate proposal for it, or someone else has a better idea and is free to tag me with it.

PR #27214, opened 2018-08-25
src/make.bat: add missing go.exe and dist.exe extensions

l-we:
'go' is not an internal or external command, nor is it a runnable program
'disk' is not an internal or external command, nor is it a runnable program


l-we:
I signed it!

PR #27407, opened 2018-08-31
log: export a default logger

hugelgupf:
I keep seeing this redefined all over. Thought I'd just export the default one.

### DIFF
--- a/src/log/log.go
+++ b/src/log/log.go
@@ -70,6 +70,12 @@ func (l *Logger) SetOutput(w io.Writer) {
 	l.out = w
 }
 
+// Default is the default logger.
+//
+// Calling SetOutput does not affect this logger.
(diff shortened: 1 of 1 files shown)


hugelgupf:
I don't see how this is a significant change at all.

This changes nothing about the API guarantees the `log` package makes. Yes, the newly exported symbol must remain there forever -- but it already had to, because `std` can never be changed according to the API guarantee.

agnivade:
It is a new API. Any new API (regardless of how small or big it is) needs to go through the proposal process.

If what you say makes sense to the team, then it will be accepted anyways and then the change can be made. There is no harm done in following the process.

ainar-g:
I feel like this change is harmful, since it encourages people to use mutable global state.

PR #27444, opened 2018-09-02
go/types: correct misspelling in function doc

leighmcculloch:
The indirectType function comment uses the phrase 'layed out'. In the
context of that phrase, where something is being placed or sprawled,
the word should be 'laid'. 'Layed' is a misspelling of 'laid'.


leighmcculloch:
I did not create an issue before opening this, because it is only a misspell in a comment.

PR #27570, opened 2018-09-08
time: support colon at start of TZ value

BusyJay:
According to POSIX, there are three formats for TZ variable. When
it refers to timezone file, it should starts with a colon. This commit
removes the colon if it exists, so that it keeps compatible with both
the spec and the old behavior.


opoplawski:
We use TZ=:/etc/localtime here for performance optimization, but it is breaking TZ setting in filebeat due to this problem.  Would be great to see this fixed.

PR #27579, opened 2018-09-09
encoding/asn1: add ability to marshal to general string

### LINKED ISSUE #18832: ASN1: Cannot marshal into a GeneralString
go version go1.7.5 linux/amd64

GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GORACE=""
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build372878858=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1

### The issue
It is not possible to marshal into a an asn1 general string. It is possible to unmarshal from one but it is only possible to marshal into one of ia5, printable or utf8.
[https://play.golang.org/p/cwv9ZGV5K5](https://play.golang.org/p/cwv9ZGV5K5)

### Suggestion
There should be an struct tag such as "generalstring" that works in the same way as the "ia5" tag but sets the asn1 tag value of 27.

jcmturner:
The current implementation cannot marshal a string into asn1 general
string encoded bytes. This capability has been added.

Fixes #18832

### DIFF
--- a/src/encoding/asn1/asn1.go
+++ b/src/encoding/asn1/asn1.go
@@ -677,6 +677,8 @@ func parseField(v reflect.Value, bytes []byte, initOffset int, params fieldParam
 				result, err = parseIA5String(innerBytes)
 			case TagT61String:
 				result, err = parseT61String(innerBytes)
+			case TagGeneralString:
+				result, err = parseT61String(innerBytes)
 			case TagUTF8String:
 				result, err = parseUTF8String(innerBytes)
 			case TagInteger:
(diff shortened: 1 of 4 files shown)


t0rr3sp3dr0:
@bradfitz @odeke-em @agl @FiloSottile, can we move forward with this PR? It has been more than 4 years since the issue was first reported and it is still a problem on the latest version of Go.

I mentioned you because you are the reviewers assigned for this PR on Gerrit.

PR #27624, opened 2018-09-11
container/list: change Len(), Front(), and Back() methods to allow nil receiver

roustem:
At the moment List.Len() panics when called with an empty receiver:
https://play.golang.org/p/60wWutjR-mt

This is not consistent with the built-in `len` implementation that allows `nil` parameters: https://golang.org/pkg/builtin/#len


roustem:
I signed it!

roustem:
Closing this PR after discussion on golang.org/cl/134636

PR #27697, opened 2018-09-15
html/template: add support for JavaScript modules.

tomuta:
html/template does not properly treat JavaScript code as JavaScript when using a <script>
tag with "module" set as the name attribute.

See also:
https://www.w3.org/TR/html5/semantics-scripting.html#element-attrdef-script-type and
https://html.spec.whatwg.org/multipage/scripting.html#the-script-element:module-script-2

Fixes #31327


tomuta:
I signed it!

PR #27758, opened 2018-09-19
reflect: fix deep equal bug on comparing func types

everdance:
deepEqual function only compare nil values for func types, need also compare func pointer address.
bug example: 
https://play.golang.org/p/TKKxpYIzs9W

### DIFF
--- a/src/reflect/all_test.go
+++ b/src/reflect/all_test.go
@@ -779,6 +779,7 @@ var (
 	fn1 func()             // nil.
 	fn2 func()             // nil.
 	fn3 = func() { fn1() } // Not nil.
+	fn4 = fn3
 )
 
 type self struct{}
@@ -811,6 +812,8 @@ var deepEqualTests = []DeepEqualTest{
 	{error(nil), error(nil), true},
 	{map[int]string{1: "one", 2: "two"}, map[int]string{2: "two", 1: "one"}, true},
 	{fn1, fn2, true},
+	{fn3, fn3, true},
+	{fn3, fn4, true},
 
 	// Inequalities
 	{1, 2, false},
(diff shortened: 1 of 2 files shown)


everdance:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here (e.g. `I signed it!`) and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google.com/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).

I signed it!

everdance:
the we need change the language specification, how it can be logically sound that you can do `f1 := func() {}; fn2 := f1` but not `f2 == f1`

randall77:
The language spec already covers this:

> Slice, map, and function values are not comparable. However, as a special case, a slice, map, or function value may be compared to the predeclared identifier nil.

So this code won't even compile:
```
var a func()
var b func()
var c = a == b
```

`deepEqual` is in kind of a tough spot here, because it's being asked to do something - compare two values of function type - when the language spec declares such comparisons to be illegal. Not just having an undefined result, but it's a question you're not allowed to ask. `deepEqual` chose `false` as the answer. Another possible answer is to panic, but that has its own issues.

What would you have `deepEqual` do instead? Comparing the function address isn't enough - some functions might be closures.  And due to the way shared library linking is done, identical references to the same function in different packages might have different function addresses (one would be the address of a trampoline that jumps to the other).

everdance:
ok, it's way more complicated. close this now.

PR #27871, opened 2018-09-26
os: add support for long path names on unix RemoveAll

### LINKED ISSUE #27029: os: RemoveAll fails for long file paths
`go version go1.10.3 linux/amd64`

Yes

Linux, amd64

### What did you do?

Make a long file path (> 4096 chars). `os.RemoveAll(path)`.

```go
package main

import (
        "io/ioutil"
        "os"
        "path/filepath"
        "syscall"
        "fmt"
)

func main() {
        tmpDir, err := ioutil.TempDir("", "")
        if err != nil {
                panic(err)
        }
        err = createVeryLongFilePath(tmpDir)
        if err != nil {
                panic(err)
        }

        fmt.Println(tmpDir)
        err = os.RemoveAll(tmpDir)
        if err != nil {
                panic(err)
        }
}

func createVeryLongFilePath(startPath string) error {
        currentPath := startPath
        for i := 0; i < 41; i++ {
                name := ""
                for j := 0; j < 100; j++ {
                        name = name + "a"
                }
                if err := mkdirAt(currentPath, name); err != nil {
                        return err
                }
                currentPath = filepath.Join(currentPath, name)
        }

        return nil
}

func mkdirAt(atPath, path string) error {
        fd, err := syscall.Open(atPath, syscall.O_RDONLY, 0)
        if err != nil {
                return err
        }
        defer syscall.Close(fd)

        return syscall.Mkdirat(fd, path, 755)
}
```

Success.

### What did you see instead?
it fails with `file name too long`.

### Suggested solutions

We have worked around this using `unix.Openat` and `unix.Unlinkat` [in our code base](https://github.com/cloudfoundry/grootfs/blob/master/removeall/removeall_unix.go), but this solution would only work on unix.

The problem really is with os.Remove, whose unix syscalls don't accept long file paths. An alternate solution would be to alter os.Remove to split long path names on unix systems.

We'd be happy to submit pull requests for either approach.
cc @callisto13

ostenbom:
On unix systems, long enough path names will fail when performing syscalls
like `Lstat`. The current RemoveAll uses several of these syscalls, and so
will fail for long paths. This can be risky, as it can let users "hide"
files from the system or otherwise make long enough paths for programs
to fail. By using `Unlinkat` and `Openat` syscalls instead, RemoveAll is
safer on unix systems. Initially implemented for linux, darwin, and several bsds.

Fixes #27029

Co-authored-by: Giuseppe Capizzi <gcapizzi@pivotal.io>
Co-authored-by: Julia Nedialkova <yulia.nedyalkova@sap.com>

### DIFF
--- a/src/internal/syscall/unix/at.go
+++ b/src/internal/syscall/unix/at.go
@@ -1,0 +1,58 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build linux darwin freebsd openbsd netbsd dragonfly
+
+package unix
+
+import (
+	"syscall"
+	"unsafe"
+)
+
+func Unlinkat(dirfd int, path string, flags int) error {
+	var p *byte
+	p, err := syscall.BytePtrFromString(path)
+	if err != nil {
+		return err
+	}
+
+	_, _, errno := syscall.Syscall(unlinkatTrap, uintptr(dirfd), uintptr(unsafe.Pointer(p)), uintptr(flags))
+	if errno != 0 {
+		return errno
+	}
+
+	return nil
+}
+
+func Openat(dirfd int, path string, flags int, perm uint32) (int, error) {
+	var p *byte
+	p, err := syscall.BytePtrFromString(path)
+	if err != nil {
+		return 0, err
+	}
+
+	fd, _, errno := syscall.Syscall6(openatTrap, uintptr(dirfd), uintptr(unsafe.Pointer(p)), uintptr(flags), uintptr(perm), 0, 0)
+	if errno != 0 {
+		return 0, errno
+	}
+
+	return int(fd), nil
+}
+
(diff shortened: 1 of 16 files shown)


gcapizzi:
I am OK with my commits being contributed this project 🙂

ianlancetaylor:
@gcapizzi @yulianedyalkova GitHub is showing that you've made commits to this CL but I'm not seeing you as having signed the Google CLA.  Can you confirm that you've signed it, or that there is a corporate CLA covering your work?  Thanks.  (For more information, see https://golang.org/doc/contribute.html#cla).

gcapizzi:
> Can you confirm that you've signed it, or that there is a corporate CLA covering your work?

I should be covered by Pivotal's CLA.

ianlancetaylor:
OK, and it looks like @yulianedyalkova is covered by SAP's assignment.  Thanks.

yulianedyalkova:
@ianlancetaylor yes, that should be the case. And if it is still needed - I am also ok with my commits being contributed to the project.

PR #27965, opened 2018-10-02
Use 404 error Status Text instead of hardcoded value

faultysegment:
net/http: Use Status Text for 404 Error in the NotFound function instead of hardcoded string

At the moment NotFound function that is used as handler function for NotFoundHandler uses a hardcoded string as response. This commit uses appropriate Status Text from net/http/status.go for this purpose to avoid hardcoded values


areldar:
I signed it!

PR #27971, opened 2018-10-02
net/http: add Post example

esell:
Currently there is no example for http.Post in the
documentation. This adds an example.

### DIFF
--- a/src/net/http/example_test.go
+++ b/src/net/http/example_test.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 )
 
 func ExampleHijacker() {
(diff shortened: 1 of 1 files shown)


src/net/http/example_test.go
```diff
@@ -191,3 +192,19 @@ func ExampleNotFoundHandler() {
 
 	log.Fatal(http.ListenAndServe(":8080", mux))
 }
+
+func ExamplePost() {
+	r, e := http.Post("http://127.0.0.1/count", "text/plain", strings.NewReader("POST data"))
```
AnikHasibul: I prefer `err` than `e`, I know, it's an example test but `err` would be better for an example.

award28: I agree with @AnikHasibul . It's go convention to use `err` instead of `e`.

award28: You should also give all of the variables more meaningful names.

```suggestion
func ExamplePost() {
	req, err := http.Post("http://127.0.0.1/count", "text/plain", strings.NewReader("POST data"))
        defer req.Body.Close()
	if err != nil {
		fmt.Println(err.Error())
	}
 	body, err := ioutil.ReadAll(req.Body)
	if err != nil {
		fmt.Println(err.Error())
	}
 	fmt.Printf("%s\n", body)
 }	
```

award28 (changes requested):

PR #28050, opened 2018-10-06
os: fix backslash behavior in Expand

lifubang:
os.Setenv("foo", "bar")
If we use os.ExpandEnv("\\$foo"), we will get \bar, not \$foo.
So, I think we need to fix this problem.

Please check, thanks.


lifubang:
I signed it!

PR #28172, opened 2018-10-12
doc: fix error prone ebnf grammar about unary_expr production

y1yang0:
consider the following expr:
```go
const (
	EOF = -(iota + 1)
        ....
)
```
if we write recursive-descent parser according to official language specification, we would get recursive calling.
In fact, official implementation is also not conformed to those grammar:
```go
// UnaryExpr = PrimaryExpr | unary_op UnaryExpr .
func (p *parser) unaryExpr() Expr {
	if trace {
		defer p.trace("unaryExpr")()
	}
        //unary_op first
	switch p.tok {
	case _Operator, _Star:...
	case _Arrow:...
	}
       // primaryExpr second
	return p.pexpr(true)
}
```

### DIFF
--- a/doc/go_spec.html
+++ b/doc/go_spec.html
@@ -3413,7 +3413,7 @@ Operators combine operands into expressions.
 
 <pre class="ebnf">
 Expression = UnaryExpr | Expression binary_op Expression .
-UnaryExpr  = PrimaryExpr | unary_op UnaryExpr .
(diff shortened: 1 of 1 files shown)


racaljk:
there are many unordered/ill-form production rules in lang spec which causing potential recursive calling  if we totally follow their orders, I hope we can reorder them carefully;)
(based on [my practise project ](https://github.com/racaljk/g5))

PR #28287, opened 2018-10-19
path/filepath: implement IsPatternValid(pattern)

fstab:
The current glob pattern implementation in `match.go` does not provide any way to learn if pattern is valid.

It does not help to call `Match(pattern, s)` to test if you get an error, because depending on `s` this might not return an error even though the pattern is invalid (like `Match("a[", "a")`).

As patterns are often provided by the user, it would be great to have a way to validate the pattern when reading the user input.

This PR implements `IsPatternValid(pattern)` to validate glob patterns. If `IsPatternValid(pattern)` returns `true`, you can be sure that `Match(pattern, s)` will never return an error.

### DIFF
--- a/src/path/filepath/match.go
+++ b/src/path/filepath/match.go
@@ -223,6 +223,65 @@ func getEsc(chunk string) (r rune, nchunk string, err error) {
 	return
 }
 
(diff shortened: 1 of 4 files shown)


fstab:
I think it's ok to continue on the other one.

PR #28354, opened 2018-10-24
crypto/tls: Deprecate tls 1.0 and tls 1.1

loganaden:
According to : https://tools.ietf.org/html/draft-ietf-tls-oldversions-deprecate-00


loganaden:
I signed it!

loganaden:
Try again.

PR #28358, opened 2018-10-24
syscall/js: add the Value.Truthy method

### LINKED ISSUE #28264: syscall/js: Add Truthy() and Falsy() to Value methods
go version go1.11 darwin/amd64

Yes.

js/wasm

---

JS has the concept of ["truthy" and "falsy"](https://developer.mozilla.org/en-US/docs/Glossary/Truthy).  It'd be nice if these were part of syscall/js, either as part and parcel of Value.Bool(), or failing that something along the lines of

```go
func (o Value) Falsy() bool {
	return !o.Truthy()
}

func (o Value) Truthy() bool {
	switch o.Type() {
	case TypeUndefined, TypeNull:
		return false
	case TypeBoolean:
		return o.Bool()
	case TypeNumber:
		if math.IsNaN(o.Float()) {
			return false
		}
		return o.Float() != 0
	case TypeString:
		return o.String() != ""
	case TypeSymbol, TypeObject, TypeFunction:
		return true
	}
	return true
}
```

theclapp:
Truthy returns the JavaScript "truthiness" of the given value.  In
JavaScript, false, 0, "", null, undefined, and NaN are "falsy", and
everything else is "truthy".

Fixes #28264

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1022,6 +1022,7 @@ Lai Jiangshan <eag0628@gmail.com>
 Lakshay Garg <lakshay.garg.1996@gmail.com>
 Lann Martin <lannm@google.com>
 Lanre Adelowo <yo@lanre.wtf>
+Larry Clapp <larry@theclapp.org> <lmclapp68@gmail.com>
 Larry Hosken <lahosken@golang.org>
 Lars Jeppesen <jeppesen.lars@gmail.com>
 Lars Lehtonen <lars.lehtonen@gmail.com>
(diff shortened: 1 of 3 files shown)


theclapp:
> bradfitz changed the title from syscall/js: add Truthy() as a method on js.Value to syscall/js: add the Value.Truthy method

Thanks, @bradfitz.  I updated the Git commit message; I didn't think to update the issue title to match.

bradfitz:
> Thanks, @bradfitz. I updated the Git commit message; I didn't think to update the issue title to match.

The commit message(s) you use are ignored for the PR->Gerrit mapping. Only the PR title & body are used.

PR #28385, opened 2018-10-25
encoding/xml: allow cdata/chardata to be named and nested with annotations

### LINKED ISSUE #16198: encoding/xml: no way to output cdata fierld into a nested node
> What version of Go are you using (`go version`)?

`go version go1.6.2 darwin/amd64`

> What operating system and processor architecture are you using (`go env`)?

```
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/akrylysov/go"
GORACE=""
GOROOT="/usr/local/Cellar/go/1.6.2/libexec"
GOTOOLDIR="/usr/local/Cellar/go/1.6.2/libexec/pkg/tool/darwin_amd64"
GO15VENDOREXPERIMENT="1"
CC="clang"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fno-common"
CXX="clang++"
CGO_ENABLED="1"
```

> What did you do?

```
type Test struct {
	Name string `xml:"Person>Name,cdata"`
}
```
https://play.golang.org/p/aVay4sw8xA

> What did you expect to see?

```
<Test><Person><Name><![CDATA[foo]]</Name></Person></Test>
```

> What did you see instead?

```
error: xml: invalid tag in field Name of type main.Test: "Person>Name,cdata"
```

I'm working on a web service which supports JSON and XML as output formats:
```
type Response struct {
    XMLName xml.Name `json:"-" xml:"response"`
    Foo string       `json:"foo" xml:"foo"`
    Bar string       `json:"bar" xml:"bar"`
}
```
It worked perfectly, but recently we got a new requirement from our external partner to wrap all XML values into CDATA. So instead of:
```
<response><foo>foo1</foo><bar>bar2</bar></response>
```
I need this:
```
<response><foo><![CDATA[foo1]]</foo><bar><![CDATA[bar2]]</bar></response>
```

Currently Go doesn't provide a way to do that without introducing a new nested structure like:
```
type CdataString struct {
	Value string `xml:",cdata"`
}

type Response struct {
    XMLName xml.Name `json:"-" xml:"response"`
    Foo CdataString  `json:"foo" xml:"foo"`
    Bar CdataString  `json:"bar" xml:"bar"`
}
```
which breaks JSON serialization.

crholm:
A fix that will allow for annotating cdata and chardata with
names and nested elements. e.g. `a-name,cdata` or `a>b>c,cdata`.
A typical usecase might be marshling to json and xml is requierd
and reducing the need for mapping the sturct to others.

Fixes #16198

Some reflection.
I'm not familiar with this code base and there might be some more
corner case that i'm missing in the addition to the tests i've added.

There is also a risk that the way of annotating might end up some
what confusing which might be worth considering. eg. 
`Name string`, turns into `<struct><Name>..</Name></struct>`
`Name string 'xml:"name"'`, turns into `<struct><name>..</name></struct>`
`Name string 'xml:",cdata"'` turns into `<struct><![CDATA[...]]></struct>`
`Name string 'xml:"name,cdata"'` turns into `<struct><name><![CDATA[...]]></name></struct>`

I would however argue that is still provides a needed feature 

There is also a case to be made that the same principle should apply
of annotation with innerxml as for cdata and chardata.

### DIFF
--- a/src/encoding/xml/marshal.go
+++ b/src/encoding/xml/marshal.go
@@ -802,10 +802,18 @@ func (p *printer) marshalStruct(tinfo *typeInfo, val reflect.Value) error {
 		vf := finfo.value(val)
 
 		switch finfo.flags & fMode {
-		case fCDATA, fCharData:
+		case fCharData, fCharData | fElement, fCDATA, fCDATA | fElement:
+
+			if finfo.flags&fOmitEmpty != 0 && isEmptyValue(vf) {
+				continue
+			}
+
 			emit := EscapeText
-			if finfo.flags&fMode == fCDATA {
+			if finfo.flags&fCDATA == fCDATA {
 				emit = emitCDATA
+			}
+			if finfo.flags&fElement == fElement {
+				emit = s.pushEmitTrim(finfo, emit)
 			}
 			if err := s.trim(finfo.parents); err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


crholm:
I signed it!

crholm:
I signed it! again

PR #28420, opened 2018-10-26
runtime: remove no-op pointer writes in treap rotations

zhiqiangxu:

### DIFF
--- a/src/runtime/sema.go
+++ b/src/runtime/sema.go
@@ -373,19 +373,11 @@ Found:
 func (root *semaRoot) rotateLeft(x *sudog) {
 	// p -> (x a (y b c))
 	p := x.parent
-	a, y := x.prev, x.next
-	b, c := y.prev, y.next
+	y := x.next
+	b := y.prev
 
 	y.prev = x
 	x.parent = y
-	y.next = c
-	if c != nil {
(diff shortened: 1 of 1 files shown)


src/runtime/sema.go
```diff
@@ -373,19 +373,11 @@ Found:
 func (root *semaRoot) rotateLeft(x *sudog) {
 	// p -> (x a (y b c))
 	p := x.parent
-	a, y := x.prev, x.next
```
jcs090218: I checked the whole file but the code not seems to be unnecessary. Even though, I think the code remain here is to exemplify what was the logic here before they code this?

jcs090218 (commented):
@zhiqiangxu I might need to explain why the code is unnecessary.

src/runtime/sema.go
```diff
@@ -373,19 +373,11 @@ Found:
...
-	b, c := y.prev, y.next
+	y := x.next
+	b := y.prev
 
 	y.prev = x
 	x.parent = y
-	y.next = c
```
AnikHasibul: Why did they even use this before? What was the logic?

AnikHasibul (approved):

PR #28533, opened 2018-11-01
go/printer: print parenthesized declarations if len(d.Specs) > 1

dtolpin:
Parenthesized declaration must be printed if len(d.Specs) > 1 even if d.Lparen==token.NoPos. This happens if the node tree is created programmatically. Otherwise, all but the first specifications just silently disappear from the output.


dtolpin:
Please advise me where and how I should add a unit test for this change in the golang source tree.

PR #28534, opened 2018-11-01
math/rand: add example for Intn func

cn0047:
Provide pretty simple example for Intn function.


cn007b:
Can't make changes here, so I created [new pull-request](https://github.com/golang/go/pull/30706), please take a look: [#30706](https://github.com/golang/go/pull/30706).

PR #28537, opened 2018-11-01
math: add architecture independent maxint and maxuint

Silentd00m:
Implemented architecture independent `MaxInt`, `MaxUint`, `MinInt` and `MinUint` constants for math package.


Silentd00m:
I signed it!

PR #28587, opened 2018-11-04
go/ast: break out after first variable in ExampleCommentMap

zigo101:
The current ExampleCommentMap might panic if there are more satisfied
comments in the parsed program.


AnikHasibul (approved):
LGTM

PR #28624, opened 2018-11-06
encoding/json: add simple example

cn0047:
Add simple example which uses Marshal and Unmarshal functions.

### DIFF
--- a/src/encoding/json/example_test.go
+++ b/src/encoding/json/example_test.go
@@ -24,12 +24,12 @@ func Example() {
 	result := map[string]string{}
 	json.Unmarshal(jsonData, &result)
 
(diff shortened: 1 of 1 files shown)


src/encoding/json/example_test.go
```diff
@@ -14,6 +14,26 @@ import (
...
+	data := map[string]string{"Name": "Platypus", "Order": "Monotremata"}
+
+	// To JSON:
+	jsonData, _ := json.Marshal(data)
+
+	// From JSON:
+	result := map[string]string{}
+	json.Unmarshal(jsonData, &result)
+
+	fmt.Printf("Origin data: %+v \n", data)
```
AnikHasibul: `Origin` or `Original`? Just for knowing.

PR #28735, opened 2018-11-12
encoding/json: document the unmarshalling of missing json fields

### LINKED ISSUE #27172: encoding/json: clarify merge semantics of Unmarshal

Go playground

Yes

Go playground

### What did you do?

https://play.golang.org/p/QM3FKwmR9h1

```
{john}
{smith}
```
### What did you see instead?

```
{john }
{john smith}
```

### In short...

The Unmarshal documentation says:
> Unmarshal parses the JSON-encoded data and stores the result in the value pointed to by v.

~ https://golang.org/pkg/encoding/json/#Unmarshal

I assumed that after Unmarshal returns, the struct would *exactly* match the provided JSON. I found this quite surprising and I think the documentation should mention that "Fields that are not present in the JSON document will retain their original values." 

The existing functionality does give one an elegant way to set default values by having Unmarshal only update those fields which are present in the JSON document. I do not dispute that it is useful. Still, I would like to see the documentation mention this behaviour explicitly.

hellozee:
The existing documentation misses covering what will happen if JSON
data with missing fields is unmarshalled, so a line is added to the
documentation of json.Unmarshall() mentioning the consequences of
unmarshalling JSON fields.

Fixes #27172

### DIFF
--- a/src/encoding/json/decode.go
+++ b/src/encoding/json/decode.go
@@ -21,7 +21,8 @@ import (
 
 // Unmarshal parses the JSON-encoded data and stores the result
 // in the value pointed to by v. If v is nil or not a pointer,
-// Unmarshal returns an InvalidUnmarshalError.
+// Unmarshal returns an InvalidUnmarshalError. Fields that are not present
+// in the JSON encoded data will retain their original values.
 //
 // Unmarshal uses the inverse of the encodings that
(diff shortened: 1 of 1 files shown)


AnikHasibul (approved):
LGTM

PR #28746, opened 2018-11-12
cmd/go: allow the user to specify ar via an environment variable

arizvisa:
This allows one to customize which ar to use by fetching its path
from the environment. This way one can swap it out for a
different implementation.


arizvisa:
In gccgo, "ar" and "rc" are both hardcoded for some particualr reason. This turns them into environment variables so that a user can use alternate versions without having to rebuild go. The environment variables that were chosen are the same as used in autoconf/automake.

PR #28776, opened 2018-11-13
cmd/internal/obj/arm64: fix spelling in arm64 doc

SCKelemen:
ln5: "instrutions" => "instructions";
ln159: "immedate" => "immediate";


SCKelemen:
This is a small change and does not change any Go source. It only updates the documentation, but it is in a go file.

PR #28807, opened 2018-11-15
strconv: fix a typo mistake in comments

awesee:
Fix a typo mistake in comments


bradfitz:
Not a typo. See https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals

PR #28811, opened 2018-11-15
cmd/compile/internal/ssa/gen: add the bx register, for consistency

xyproto:
* The `bx` register is added, just like for `ax`, `cx` and the others:

    bx         = buildReg("BX")

* When there is a list of registers, `bx` is sorted to come between `ax` and `cx`.


randall77:
This doesn't need fixing. The registers can be in any order.
We happened to pick the order in which the registers are numbered by the hardware. BX is #3, not #1.

xyproto:
ok, thanks for reviewing

PR #28851, opened 2018-11-18
crypto/tls: add SkipSNI option

taoso:
This change modifies Go to has the ability not to send the SNI extension information.

Fixes #28754

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -502,6 +502,9 @@ type Config struct {
 	// an IP address.
 	ServerName string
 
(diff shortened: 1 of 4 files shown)


lvht:
I signed it!

FiloSottile:
Copying my message from Gerrit here as gerritbot didn't:

> Closing due to the proposal at #28754 being declined. Sorry about the wasted effort, for the future we recommend waiting for proposals to be accepted before sending PRs, as even when they are not declined they might change significantly in the process.

PR #28936, opened 2018-11-24
net/http: close connection if `Connection: close` is set

jiajunhuang:
As https://tools.ietf.org/html/rfc2616#page-117 says: 

HTTP/1.1 defines the "close" connection option for the sender to signal that the connection will be closed after completion of the response.

But it does not specify the connection should be closed by client or server, for safety, if `Connection: close`
is set in response header, server should close the connection after write data. Recently we have been
affect by this issue: we write a lots of data to client, but the client does not close the connection(
keep-alive is set by default in HTTP/1.1), so memory is still hold by `response`, and memory is used up.

### DIFF
--- a/src/net/http/server.go
+++ b/src/net/http/server.go
@@ -1141,6 +1141,10 @@ func (w *response) WriteHeader(code int) {
 			w.handlerHeader.Del("Content-Length")
 		}
 	}
+
+	if w.handlerHeader.get("Connection") == "close" {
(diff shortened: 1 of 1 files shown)


jiajunhuang:
Hi, all, it seems that it's hard to write unit test for it, since `net.Conn` can't be accessed from `http.Request`. So I can test `response.Write` directly, but it's meaning less.

PR #28980, opened 2018-11-28
change func Handle panic->info,when http: multiple registrations

dusbin:
When I use http.HandleFunc()  to registrations function more times,it happen panic.
I think it return more better than happen a panic


Bin-DuS:
I signed it!

Bin-DuS:
I signed it!

agnivade:
Hi @Bin-DuS - The PR seems to have some additional code than just returning instead of panic'ing. In any case, changing the panic behavior to just return will be a breaking change at this point. And I believe the decision to panic was made deliberately.

If you believe this behavior should be changed, I would request you to open an issue and discuss there. Once it's approved, then we can make the change. Thank you.

PR #28981, opened 2018-11-28
math/big: better initial guess for nat.sqrt

sukop:
The proposed change introduces a better initial guess which is closer to the final value and therefore converges in fewer steps. Consider for example sqrt(8): previously the guess was 8, whereas now it is 4 (and the result is 2). All this change does is it computes the division by two more accurately while it keeps the guess ≥ √x.


sukop:
I signed it!

PR #29138, opened 2018-12-07
time: reuse the result of Time.sec in Time.Before

skaldesh:
The current implementation calls sec() for both the receiver t and
the parameter u twice, if the seconds of t are greater or equal than u's seconds.
The same optimization is already present in Time.After.


skaldesh:
I signed it!

I think the problem was, that my google mail was not added to my github account

PR #29179, opened 2018-12-11
runtime: PROCESS_ALL_ACCESS didn't check for 0XFFFF

### LINKED ISSUE #29176: Incorrect constant value
https://github.com/golang/go/blob/01e072db5d26c224dfbe7763a5b94ab23c163983/src/runtime/syscall_windows_test.go#L815

From winnt.h we can see that if we are running on a version of Windows greater than Vista, `0xFFFF` should be or-ed with the `STANDARD_RIGHTS_REQUIRED` and `SYNCHRONIZE` flags.

_winnt.h#L11279_
```c++
#if (NTDDI_VERSION >= NTDDI_VISTA)
#define PROCESS_ALL_ACCESS        (STANDARD_RIGHTS_REQUIRED | SYNCHRONIZE | \
                                   0xFFFF)
#else
#define PROCESS_ALL_ACCESS        (STANDARD_RIGHTS_REQUIRED | SYNCHRONIZE | \
                                   0xFFF)
#endif
```

laches1sm:
Fixes #29176. PROCESS_ALL_ACCESS now checks for 0xFFFF.

### DIFF
--- a/src/runtime/syscall_windows_test.go
+++ b/src/runtime/syscall_windows_test.go
@@ -812,7 +812,7 @@ func TestNumCPU(t *testing.T) {
 
 	const (
 		_CREATE_SUSPENDED   = 0x00000004
-		_PROCESS_ALL_ACCESS = syscall.STANDARD_RIGHTS_REQUIRED | syscall.SYNCHRONIZE | 0xfff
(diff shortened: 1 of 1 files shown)


laches1sm:
I've signed the CLA :)

Issue #29241, opened 2018-12-13
cmd/go: "go get" fails on import path patterns with wildcards ("...")


FiloSottile:
The fix for `cmd/go: remote command execution during "go get -u"` (#29230) as released in Go 1.10.6 and 1.11.3 introduced a regression causing commands like `go get example.com/...` to fail with `invalid import path: malformed import path "example.com/...": double dot` when fetching packages not already present in GOPATH.

See https://github.com/golang/go/issues/29236#issuecomment-447144051

jadr2ddude:
# Looking through the code, I saw the following mistakes that could have caused this error:

Error 1 - `strings.Contains(path, "..")` creates false positives with `something/...`
https://github.com/golang/go/commit/8954addb3294a5e664a9833354bafa58f163fe8f#diff-5b11f198c631a3d489e1be171853ab7bR40

Error 2 - the `...` path would cause this to error too, since it is all `.`
https://github.com/golang/go/commit/8954addb3294a5e664a9833354bafa58f163fe8f#diff-5b11f198c631a3d489e1be171853ab7bR70

# Suggested Fix

Remove the check listed in error 1. In addition to causing false positives, it is rendered unnecessary by the line mentioned in error 2. Leave the line in error 2 unchanged, but add a check that skips this validation for the last element if it exactly matches `...`.

jadr2ddude:
Is someone currently working on this? If not, I can try to create a fix.

jadr2ddude:
In absence of a response, I have started fixing this.

FiloSottile:
Thank you, but I believe @dmitshur and @bcmills are already moving on a fix.

jadr2ddude:
Ok. Thank you for letting me know.

willbuckner:
```
diff --git a/src/cmd/go/internal/get/path.go b/src/cmd/go/internal/get/path.go
index d443bd28a9..2a109d6638 100644
--- a/src/cmd/go/internal/get/path.go
+++ b/src/cmd/go/internal/get/path.go
@@ -41,9 +41,6 @@ func checkPath(path string, fileName bool) error {
        if path == "" {
                return fmt.Errorf("empty string")
        }
-       if strings.Contains(path, "..") {
-               return fmt.Errorf("double dot")
-       }
        if strings.Contains(path, "//") {
                return fmt.Errorf("double slash")
        }
@@ -59,8 +56,10 @@ func checkPath(path string, fileName bool) error {
                        elemStart = i + 1
                }
[...]

PR #29245, opened 2018-12-14
cmd/go: fix regression allowing '...' wildcard as last element in go get

willbuckner:
Updates #29241


willbuckner:
I signed it!

PR #29328, opened 2018-12-18
mime: encode CTL and non-US-ASCII characters in FormatMediaType

### LINKED ISSUE #7668: mime: FormatMediaType should quote control characters
<pre>What does 'go version' print?
go version devel +1afdecdd77c1 Sat Mar 29 17:10:25 2014 -0400 linux/amd64

What steps reproduce the problem?
<a href="http://play.golang.org/p/LcQhZpJkX2">http://play.golang.org/p/LcQhZpJkX2</a>

1. Call mime.FormatMediaType with parameters containing an ASCII control character
2. See the return value

What happened?
It returns a formatted string containing a control character as is.

What should have happened instead?
It should quote a control character with \ or return the empty string. Here is the
definition of quoted-string (<a href="http://www.ietf.org/rfc/rfc2616.txt)">http://www.ietf.org/rfc/rfc2616.txt)</a>. A control character
can appear as quoted-pair but not as qdtext.

       quoted-string  = ( &lt;&quot;&gt; *(qdtext | quoted-pair ) &lt;&quot;&gt; )
       quoted-pair    = &quot;\&quot; CHAR
       CHAR           = &lt;any US-ASCII character (octets 0 - 127)&gt;
       qdtext         = &lt;any TEXT except &lt;&quot;&gt;&gt;
       TEXT           = &lt;any OCTET except CTLs,
                        but including LWS&gt;
       CTL            = &lt;any US-ASCII control character
                        (octets 0 - 31) and DEL (127)&gt;

(It's actually surprising to me that it allows even a NUL after a backslash, but it's
what the standard says. It may be a spec bug?)</pre>

### LINKED ISSUE #9624: mime: FormatMediaType doesn't allow to create Content-Disposition
On Go 1.4 (darwin), `mime.FormatMediaType` checks whether the first argument (aptly called `mediatype`) contains a forward slash, and returns an empty string if it doesn't. This doesn't allow to use this function to format a `Content-Disposition` header, because the values for such header are not proper media types (they are usually `inline` or `attachment`). Notice that the documentation of the reverse function `mime.ParseMediaType` explicitly says that it can be used to parse `Content-Disposition` headers, so it looks like this limitation should be lifted.

andrius4669:
Encodes non-WSP CTL and non-US-ASCII UTF-8 characters using syntax specified in RFC 2231.

Fixes #7668
Fixes #9624

### DIFF
--- a/src/mime/mediatype.go
+++ b/src/mime/mediatype.go
@@ -19,7 +19,7 @@ import (
 // FormatMediaType returns the empty string.
 func FormatMediaType(t string, param map[string]string) string {
 	var b strings.Builder
-	if slash := strings.Index(t, "/"); slash == -1 {
+	if slash := strings.IndexByte(t, '/'); slash == -1 {
 		if !isToken(t) {
 			return ""
 		}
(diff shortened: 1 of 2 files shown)


patkujawa-wf:
Naive question: How can I tell what version of Go this will be / is release in?

andrius4669:
1.13.
you can see all the tags in https://github.com/golang/go/commit/0ac4ea79f725a0cc3797faa32ab698b8b3d77343

PR #29367, opened 2018-12-20
net: allow TCP only DNS requests in the pure Go resolver on unix

jfbus:
There is a DNS resolution bug in Kubernetes (UDP response packets get dropped by conntrack, causing timeouts in DNS queries).

This enables configuration of the pure Go resolver to use TCP, either using the GODEBUG env var (GODEBUG=netdns=go+tcp) or using the Resolver.PreferTCP flag.
 
Fixes #29358


jfbus:
Replaced by #29594

Issue #29261, opened 2018-12-14
net/url: misleading error message when url has a leading space


stefanb:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.11.3 darwin/amd64
</pre>

### Does this issue reproduce with the latest release?

Yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/stefanb/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/stefanb/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/Cellar/go/1.11.3/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.11.3/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/z2/3bdl__bs0xxd3kmf5c1pj0gm0000gn/T/go-build212726590=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre></details>

### What did you do?

```go
package main

import (
	"net/http"
)

func main() {
	_, err := http.Get(" http://example.org")
	if err != nil {
		panic(err)
	}
}
```
https://play.golang.org/p/jkcYSD6ZRcO

### What did you expect to see?

Either 
* a generic error message of invalid URL or
* a detailed error of invalid URL scheme or
* a detailed error of leading space in URL

### What did you see instead?

A *misleading detailed* error message:
```
parse  http://example.org: first path segment in URL cannot contain colon
```

In #24246 the proposed solution was to trim the URLs before using them, but from the given error message it is very hard to see what the real problem is. 

I propose to either:
* adjust the error message to eg:
```
[...]

PR #29384, opened 2018-12-21
net/url: improve url parsing error messages by quoting

stefanb:
Current implementation doesn't always make it obvious what the exact
problem with the URL is, so this makes it clearer by consistently quoting
the invalid URL, as is the norm in other parsing implementations, eg.:
strconv.Atoi(" 123") returns an error: parsing " 123": invalid syntax

Updates #29261


stefanb:
I signed it!

vagababov:
This actually caused quite a few errors in our unit tests due to quoting...

ianlancetaylor:
Us too.  See https://golang.org/cl/195978.  We decided to bite the bullet and fix the tests.

vagababov:
Yeah, in the end it wasn't that hard to patch the tests, but it was surprising :)

PR #29460, opened 2018-12-29
net/url: give a proper error message on invalid character in scheme

### LINKED ISSUE #29261: net/url: misleading error message when url has a leading space

cgopalan:
Fixes #29261


cgopalan:
@agnivade I have made changes based on our discussion in the issue, but at the moment I have one failing test:
```
--- FAIL: TestFileServerEscapesNames (0.00s)
    fs_test.go:300: test "simple_name": Get: Get http://127.0.0.1:36757/0: failed to parse Location header "0/": parse 0/: URL scheme has invalid character
FAIL
FAIL	net/http	4.252s
```
Not sure why this is failing since the scheme seems to be fine. Would love to get some pointers on this.

stefanb:
Please note that the negative tests are not actually testing the expected error message, just its mere presence.

cgopalan:
@agnivade @stefanb in the failing test I mentioned above, it seems like the `0` in `http://127.0.0.1:36757/0` url is being parsed as a location header and as a result `0` is a valid beginning character for a url. Also there are tests that count on `/ . ? ;` being valid starting characters for a url so I had to include them.
Is this the right approach? Would like some guidance here.

agnivade:
The code can do with some improvement. Your ` (c < 'A' || (c > 'Z' && c < 'a') || c > 'z') ` is repeated which I wanted to avoid. 

We don't typically discuss CLs in github. Gerrit is the place for that. Please feel free to ping Brad on the CL to get guidance.

PR #29532, opened 2019-01-03
os: clarify meaning of "truncate"

### LINKED ISSUE #29531: os: clarify meaning of "truncate"
For non-native speakers the term "empty" might be easier to understand
eg change `// truncate regular writable file when opened.` to `// truncate (empty) regular writable file when opened.` in the [O_TRUNC](https://github.com/golang/go/blob/0175064e697d3e072573b32173299b7edbda7360/src/os/file.go#L76) constant

UlisseMini:
For non-native speakers the term "empty" might be easier to understand.

Fixes #29531


UlisseMini:
I've signed the CLA

Issue #27215, opened 2018-08-25
cmd/go: go get golang/x/tools/...: no matching versions for query "latest"


wsc1:
Please answer these questions before submitting your issue. Thanks!

#### What did you do?
If possible, provide a recipe for reproducing the error.
A complete runnable program is good.
A link on play.golang.org is best.

go get golang.org/x/tools/...

#### What did you expect to see?

success

#### What did you see instead?
⎣ ⇨ go get -u -v golang.org/x/tools/...
Fetching https://golang.org/x/tools?go-get=1
Parsing meta tags from https://golang.org/x/tools?go-get=1 (status code 200)
get "golang.org/x/tools": found meta tag get.metaImport{Prefix:"golang.org/x/tools", VCS:"git", RepoRoot:"https://go.googlesource.com/tools"} at https://golang.org/x/tools?go-get=1
go: finding golang.org/x/tools/... latest
Fetching https://golang.org/x/tools?go-get=1
Parsing meta tags from https://golang.org/x/tools?go-get=1 (status code 200)
get "golang.org/x/tools": found meta tag get.metaImport{Prefix:"golang.org/x/tools", VCS:"git", RepoRoot:"https://go.googlesource.com/tools"} at https://golang.org/x/tools?go-get=1
go: finding golang.org/x/tools latest
Fetching https://golang.org/x?go-get=1
Parsing meta tags from https://golang.org/x?go-get=1 (status code 200)
Fetching https://golang.org?go-get=1
Parsing meta tags from https://golang.org?go-get=1 (status code 200)
go get golang.org/x/tools/...: no matching versions for query "latest"

#### System details

```
go version go1.11 darwin/amd64
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/scott/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/scott/Dev/zic"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/scott/Dev/zic/src/zikichombo.org/sound/go.mod"
GOROOT/bin/go version: go version go1.11 darwin/amd64
[...]

PR #29606, opened 2019-01-08
cmd/go: fix go get with ... fails in module mode

### LINKED ISSUE #29363: cmd/go: go get with ... fails in module mode
This used to work and should ideally keep working.

```
$ go get -u github.com/google/codesearch/cmd/...
go: finding github.com/google/codesearch/cmd/... latest
go: finding github.com/google/codesearch/cmd latest
go: finding github.com/google/codesearch latest
go: downloading github.com/google/codesearch v0.0.0-20150617151851-a45d81b686e8
go: extracting github.com/google/codesearch v0.0.0-20150617151851-a45d81b686e8
go get github.com/google/codesearch/cmd/...: no matching versions for query "latest"
$ go list github.com/google/codesearch/...
go: warning: "github.com/google/codesearch/..." matched no packages
```

hxzhao527:
There is a mistake about haveGoFiles detection in modload, and it caused go get with ... fails in module mode.

Fixes #29363


hxzhao527:
@bcmills Hi here. I'm sorry, but the Gerrit is too difficult for me to send a comment. 
I doubt about why we need a fix for the general case.

Now `go get` in mod mode is not always failed, just when the package not in module dependencies. 
As the #27215 shows, we can call `go get golang.org/x/perf/cmd/benchstat` and then `go get golang.org/x/perf/cmd/...`, even `go get golang.org/x/perf/...cmd/...`.

After the first `go get` add the package to module dependencies, everything works.
So the first call is the just one needed to fix. Is this correct?

PR #29728, opened 2019-01-14
net/http: useless string replace operation in Cookie.String. Fixes #29135

### LINKED ISSUE #29135: net/http: useless string replace operation in Cookie.String
* isCookieNameValid do not allow "\r" "\n" appear in the cookie name.
https://github.com/golang/go/blob/master/src/net/http/cookie.go#L168
https://github.com/golang/go/blob/master/src/net/http/http.go#L50
https://github.com/golang/go/blob/master/src/internal/x/net/http/httpguts/httplex.go#L15

* sanitizeCookieName replace "\r" "\n" to "-"
https://github.com/golang/go/blob/master/src/net/http/cookie.go#L172

This replace operation should have no effect.So it can be delete to save cpu time and make code easier to read.

bronze1man:
net/http: useless string replace operation in Cookie.String. Fixes #29135

### DIFF
--- a/src/net/http/cookie.go
+++ b/src/net/http/cookie.go
@@ -169,7 +169,7 @@ func (c *Cookie) String() string {
 		return ""
 	}
 	var b strings.Builder
-	b.WriteString(sanitizeCookieName(c.Name))
+	b.WriteString(c.Name)
 	b.WriteRune('=')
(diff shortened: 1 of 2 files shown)


bradfitz:
Closing so I can take this over on Gerrit.

PR #29950, opened 2019-01-26
cmd/go: honor vendor.conf fork repositories in modconv

tiborvass:
With this patch, many projects using vendor.conf format should be able
to transition to go mod.

Example of a problematic line in cmd/go/internal/modconv/testdata/moby.out:

  github.com/go-check/check 4ed411733c5785b40214c70bce814c3a3a689609 https://github.com/cpuguy83/check.git

In vendor.conf format, the third column represents the URL of a fork
repository which was being ignored before this patch.

Because of this, go mod was incorrectly assuming the revision (4ed...609)
exists in the upstream github.com/go-check/check when in fact it only does
in the github.com/cpuguy83/check fork.

One cannot simply expect a user to change github.com/go-check/check to
github.com/cpuguy83/check because the import path in both cases is
github.com/go-check/check.

For that example, this patch would do the equivalent of:
  go mod edit -require=github.com/go-check/check@latest
              -replace=github.com/go-check/check=github.com/cpuguy83/check@4ed411733c5785b40214c70bce814c3a3a689609

### DIFF
--- a/src/cmd/go/internal/modconv/modconv_test.go
+++ b/src/cmd/go/internal/modconv/modconv_test.go
@@ -58,6 +58,9 @@ func Test(t *testing.T) {
 			for _, r := range out.Require {
 				fmt.Fprintf(&buf, "%s %s\n", r.Mod.Path, r.Mod.Version)
 			}
+			for _, r := range out.Replace {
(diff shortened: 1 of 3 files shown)


tiborvass:
@googlebot I signed it!

tiborvass:
Ugh, apologies I just found the corresponding open issue: https://github.com/golang/go/issues/25556 which wants to fix it for all tools. Feel free to close this.

PR #29970, opened 2019-01-29
sync/once: remove unnecessary atomic operation

dragonylffly:
The Lock() of a Mutex is a memory barrier,
so the subsequent lock owners guarantee to see
the new value, the heavyweight atomic operation
is not needed

Signed-off-by: Li Wang <laurence.liwang@gmail.com>


dragonylffly:
I signed it!

PR #29994, opened 2019-01-30
bytes: check Write error before written length

leonklingele:


randall77 (approved):
Sure, this is probably the right pattern in this instance.
(Although often after Write you should record that n bytes were written before checking the error. Not necessary, or even desired, in this case.)

leonklingele:
Overhauled by #29996.

PR #29998, opened 2019-01-30
cmd/compile/internal/ssa: ignore error from second call to MatchString in test

leonklingele:


randall77 (commented):
This function is guaranteed to not return an error.
The previous call to regexp.MatchString didn't fail, so this one can't.
(Errors only depend on the first argument.)

Probably the right fix is to do:
  match, _ = regexp.MatchString(want, got.e)
to indicate we know the error can't happen.

leonklingele:
Updated.

PR #30003, opened 2019-01-30
cmd/internal/obj/mips: use r instead of p.Reg in call to OP_IRR

leonklingele:


randall77 (commented):
This looks like a bug to me. I think line 1278 should be using r instead of p.Reg.
Not sure though, this exact pattern is not used anywhere else.
Needs a mips expert.

randall77:
@minux @cherrymui

cherrymui:
I agree with @randall77. Line 1278 should be using r. Effectively the result was still correct though, because REGZERO's encoding is also 0. Anyway, I think we should change line 1278 to use r for the right fix.

randall77:
REGZERO is not zero. It is
```
REGZERO = REG_R0
REG_R0 = obj.RBaseMIPS + iota // iota is zero here
RBaseMIPS  = 13 * 1024 // range [13k, 14k)
```
I guess 0 and REGZERO do have the same low 5 bits, so the result is still correct.

cherrymui:
Yes, that's what I meant. The low 5 bits are still 0.

PR #30015, opened 2019-01-30
internal/x/text/unicode/bidi: ignore unused return values in test

leonklingele:


leonklingele:
File should be removed altogether. https://github.com/golang/go/issues/30671

PR #30021, opened 2019-01-30
internal/x/text/unicode/bidi: where does the i come from?

leonklingele:


leonklingele:
Where does the `i` come from? I'll provide a fix if someone can give me a hint :)

leonklingele:
I've opened an [issue](https://github.com/golang/go/issues/30671) for discussion.

PR #30022, opened 2019-01-30
net: explain why we ignore the first line of Plan 9 interface-status files

leonklingele:


leonklingele:
Why is `statusf.readLine()` called twice on Plan 9? I'll provide a fix if someone can give me a hint :)

leonklingele:
Updated to explain why we ignore the first line of Plan 9 interface-status files.

PR #30023, opened 2019-01-30
runtime/pprof/internal/profile: do not ignore error

leonklingele:


leonklingele:
`postDecode()` always returned a `nil` error, even if an `err` was defined in the function scope. What's the reason for that? The fixed version (properly returning the error) makes the `TestEmptyProfile` test fail. Is that even a valid test? I'll provide a fix if someone can give me a hint :)

PR #30157, opened 2019-02-10
io/ioutil: add AppendFile

Mottl:
This commit adds AppendFile function to io/ioutil.
It has the same behaviour as WriteFile, except that it doesn't truncate the file if it already exists.


Mottl:
Closed due to #30211 and #30159

PR #30187, opened 2019-02-12
context: allow cancelling a context with custom error

Saluev:
Sometimes it is useful to cancel a context with a more specific error, say,
"cancelled job X after timeout of Y expired" instead of default and not 
very informative "context cancelled". This can provide valuable 
information on what level of execution has actually timed out, if there are
multiple possibilities. Apparently, current implementation of cancellable
context in standard library already supports that, yet for some reason does
not expose this possibility, forcing developers to copy-parse `cancelCtx`
sources to achieve the described functionality. I added a simple function
to do that near the existing WithCancel function. I'm not sure what a 
perfect name would there be for it though.

### DIFF
--- a/src/context/context.go
+++ b/src/context/context.go
@@ -232,6 +232,18 @@ func WithCancel(parent Context) (ctx Context, cancel CancelFunc) {
 	return &c, func() { c.cancel(true, Canceled) }
 }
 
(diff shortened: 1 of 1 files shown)


Saluev:
I signed it!

PR #30200, opened 2019-02-13
time: improve implementation of 3 methods for the type Duration

zrcoder:
Improved the implementation of 3 methods below:
func (d Duration) Seconds() float64
func (d Duration) Minutes() float64
func (d Duration) Hours() float64


DingHub:
I signed it!

Issue #30208, opened 2019-02-13
all: redundant or missing nil checks


octomarat:
Hi! I am writing a static analysis tool to detect problems with `nil` values in Go programs. I have executed it on the source code of the Go itself and found some minor issues.

1) Redundant `err == nil` check: 
https://github.com/golang/go/blob/master/src/cmd/go/internal/work/exec.go#L106
`err` is a local variable created with `nil` value and it is not updated before the check (making it redundant).

2) Redundant `err != nil` check:
https://github.com/golang/go/blob/master/src/crypto/x509/name_constraints_test.go#L2223
`err` is always not `nil` because of the continue statement a couple of lines above.

3) Potential 'nil pointer dereference':
https://github.com/golang/go/blob/master/src/net/textproto/reader_test.go#L335
https://github.com/golang/go/blob/master/src/html/template/escape_test.go#L1873
In both cases, `err` can be `nil` making `err.Error()` call dangerous. `err == nil` check above doesn't stop the execution of the test.

I already have a fork with these issues fixed, will be happy to submit a pull request.

agnivade:
Please go ahead ! Thanks.

dgryski:
The new data-flow based nilness check in `vet` finds the first two cases:

```
/Users/dgryski/go/src/go.googlesource.com/go/src/cmd/go/internal/work/exec.go:106:11: tautological condition: nil == nil
...
/Users/dgryski/go/src/go.googlesource.com/go/src/crypto/x509/name_constraints_test.go:2223:10: tautological condition: non-nil != nil
```

 but I'm not aware of a tool that handles the third one.  Nifty!

mvdan:
@octomarat please send a single PR for all these changes. Separate commits aren't necessary and just clutter the git history, in particular since most of the changes only touch test code.

agnivade:
@octomarat - I think only the cmd/go/internal fix is remaining. Would you like to push ahead and finish this one out ?

octomarat:
Oh. I didn't noticed that some of the commits have been accepted and merged to the master already. Probably I am missing something, but what should I do to make the last one accepted and merged?

PR #30212, opened 2019-02-13
go/cmd: remove redundant check for nil

marat-rkh:
Local variable err is defined with nil value and almost immediately
compared with nil. This comparison is redundant as always evaluates
to true.

Updates #30208


octomarat:
https://github.com/golang/go/pull/30226
Current request can be closed (rejected).

octomarat:
Please, see this comment:

PR #30213, opened 2019-02-13
crypto/x509: remove redundant check for nil in tests

marat-rkh:
Comparing err variable to be not nil is redundant in this case.
The code above ensures that it is always not nil.

Updates #30208


octomarat:
#30226
Current request can be closed (rejected). Doing this as requested here:

PR #30214, opened 2019-02-13
net/textproto: prevent test from failing with nil pointer dereference

marat-rkh:
The variable err could have nil value when we call err.Error(),
because after we check it for nil above we continue the test
(t.Errorf doesn't stop the test execution).

Updates #30208


octomarat:
#30226
Current request can be closed (rejected). Doing this as requested here:

PR #30226, opened 2019-02-14
go/cmd, crypto/x509, net/textproto, html/template: fix minor issues with nil values

### LINKED ISSUE #30208: all: redundant or missing nil checks

marat-rkh:
Remove redundant checks for nil value, add missing nil checks to prevent tests from failing with 'nil pointer dereference'.

Fixes #30208.

### DIFF
--- a/src/cmd/go/internal/work/exec.go
+++ b/src/cmd/go/internal/work/exec.go
@@ -103,9 +103,7 @@ func (b *Builder) Do(root *Action) {
 		var err error
 
 		if a.Func != nil && (!a.Failed || a.IgnoreFail) {
-			if err == nil {
-				err = a.Func(b, a)
-			}
+			err = a.Func(b, a)
 		}
 
 		// The actions run in parallel but all the updates to the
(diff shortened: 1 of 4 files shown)


octomarat:
Removed additional spaces in the pull request name.

PR #30278, opened 2019-02-17
net: add IP.IsLocal

MaxSem:
Fixes #29146


src/net/ip.go
```diff
@@ -130,6 +130,19 @@ func (ip IP) IsMulticast() bool {
...
 }
 
+// IsLocal reports whether `ip' is a local address, according to
+// RFC 1918 (IPv4 addresses) and RFC 4193 (IPv6 addresses).
+func (ip IP) IsLocal() bool {
+	if ip4 := ip.To4(); ip4 != nil {
+		// Local IPv4 addresses are defined in https://tools.ietf.org/html/rfc1918
+		return ip4[0] == 10 ||
+			(ip4[0] == 172 && ip4[1]&0xf0 == 16) ||
+			(ip4[0] == 192 && ip4[1] == 168)
```
msoedov: 1) `IsLocal` name is miss-leading. Did you mean `IsPrivate`? 
2) How to handle cases `0.0.0.0` and `127.0.0.1` and `::1`?

duedal: We do at least have `IsUnspecified()` and `IsLoopback()` to cover those cases with IsLocal you can do `IsLocal() || IsUnspecified() || IsLoopback()` if you want to cover all those cases.

ilyam8: Speaking about [private addresses](https://en.wikipedia.org/wiki/Reserved_IP_addresses) - [`100.64.0.0/10`](https://tools.ietf.org/html/rfc6598) is private, [`198.18.0.0/15`](https://tools.ietf.org/html/rfc5735) is private too.

This implementation is true only for `RFC 1918 (IPv4 addresses) and RFC 4193 (IPv6 addresses).`

duedal: Sure but that doesn't mean it would still be useful too have a group for RFC 1918 and RFC 4193... We're using this to filter webhook calls, so people can setup a webhook that calls a private address in our internal networks. The carrier grade nat scope  100.64.0.0/10 and/or the benchmark test net 198.18.0.0/15 is not really any concern for us, protecting internal networks is.

ilyam8: i was not clear, my points are:
 - `IsLocal` is a bad name, i am with you on it.
 - `IsPrivate` is ok name, but changing the name brings new problem - current implementation includes only part of private ip address blocks.

ilyam8: I have a second thought about it. `IsPrivate` wont do, because there are a lot of other ip block in the private scope. 

IP blocks from `rfc1918` fall into `private` category but they have special purpose. `rfc1918` doesnt name them local tho. But the gist is -  local communication (inside enterprise). So the grouping is ok, naming is a little bit miss-leading. Proof - our comments :smile: 

But `IsLocal` makes more sense then `IsPrivate` and this logical block shouldnt cover

> How to handle cases 0.0.0.0 and 127.0.0.1 and ::1?

Only blocks from `RFC 1918 (IPv4 addresses) and RFC 4193 (IPv6 addresses)` should be there.

PR #30326, opened 2019-02-20
 crypto/tls: Expose maps for cipher suite IDs/names 

JAORMX:
These are utility functions that allow you to get maps that translate from Cipher Suite names towards IDs and vice versa.

The main purpose of this is to be able to easily parse such configurations for HTTP(s) servers without much boiler plate code.

Fixes: #30325


noneymous:
Thank you @JAORMX for this pull request, Having this change would be awesome. 

I often have problems in GO where some go packages holds some information I'd like to access, but I can't because its private and there is no way araound. This really sucks. From a builders point of view I understand, that some information is not necessary and can be unknown. But i do use Go for testing (respectively, building security checks), and such information is elemntary to me. In Python I could just go ahead and grab what I wanted, kind of adult mode,... I love Go, but this hardcore nany mode comes at a cost for me :D

JAORMX:
@noneymous thanks for checking this out! Feel free to jump in  #30325 where the discussion about this is happening.

PR #30352, opened 2019-02-22
net/http: add Dir.Open godoc

Mottl:
This commit add godoc for `http.Dir.Open()` function


Mottl:
Closed to garbage found.

PR #30381, opened 2019-02-25
container/heap: avoid memory leak

quadraticrain:
Set element in slice to nil avoiding memory leak.


WhisperRain:
I signed it!

WhisperRain:
I changed primary email of my account

PR #30545, opened 2019-03-03
cmd/go/internal/test: pass default timeout to test programs if not given from command line

### LINKED ISSUE #28147: cmd/go: when -timeout is the default (10m), pass a nonzero -test.timeout to tests
In https://go-review.googlesource.com/c/go/+/141218/1/src/cmd/go/go_test.go#133, @ianlancetaylor notes (emphasis mine):

> If you run `go test`, the go tool will pass its `-timeout` argument to the test program as the `-test.timeout` flag.  But you will only see that when running `go test`, not when running the test program directly, and **you will only see it if you use a `-timeout` option.  The default value of 10 minutes is not passed to the test program by `go test`.  This is arguably a bug.**  See also #14780.

The fact that `-test.timeout` is not passed for the default `10m` timeout makes it difficult for the test to fail cleanly if that timeout is reached (see also #28135).

okamotoyuki:
Make 'go test' command to pass the default timeout (10m) to test programs if the value is not given from command line.

Fixes #28147

### DIFF
--- a/src/cmd/go/internal/test/test.go
+++ b/src/cmd/go/internal/test/test.go
@@ -484,8 +484,9 @@ var (
 	pkgArgs          []string
 	pkgs             []*load.Package
 
-	testKillTimeout = 10 * time.Minute
-	testCacheExpire time.Time // ignore cached test results before this time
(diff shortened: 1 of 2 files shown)


okamotoyuki:
I signed it!

PR #30704, opened 2019-03-09
syscall: add support for ptrace on FreeBSD

rayrapetyan:
The existing implementation doesn't support ptrace-ing on FreeBSD.

Used by go-delve/delve#213.


rayrapetyan:
The CLA shows signed in 2016.

Issue #21167, opened 2017-07-25
crypto/tls: Export TLS default cipher suites


bdarnell:
```
$ go version
go version go1.8.3 darwin/amd64
```

I would like to set `tls.Config.CipherSuites` to a non-default value (such as [mozilla's "modern" compatibility recommendation](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility)), and prioritize either AES-GCM or ChaCha20 depending on whether hardware-accelerated AES is available (i.e. the same logic as [in tls/common.go](https://github.com/golang/go/blob/df91b8044dbe790c69c16058330f545be069cc1f/src/crypto/tls/common.go#L919)). 

There is no way that I can see to inspect this default certificate list (it is only used in the non-exported `Config.cipherSuites` method), and the `cipherhw` package is internal-only. I could duplicate the `cipherhw` package and make the `CPUID` assembly call myself (and the analogous instructions for other CPUs), but that allows for version skew between my code and `crypto/tls`: what I want is not "does this CPU have AES support", but "will `crypto/tls` *use* an accelerated AES implementation on this CPU". 

It would be nice if `crypto/tls` exposed either some way to inspect its default cipher suite list or to make the same performance-based cipher prioritization decisions as the default implementation.

ALTree:
cc @agl

bradfitz:
Possible hack: write a func to use crypto/tls to handshake with itself over an in-memory net.Pipe (not even localhost TCP) and see what it announces to itself with:
 
https://golang.org/pkg/net/#Pipe +
https://golang.org/pkg/crypto/tls/#Config.GetCertificate + InsecureSkipVerify +
https://golang.org/pkg/crypto/tls/#ClientHelloInfo.CipherSuites

*shrug*

FiloSottile:
I don't think we should commit to the default logic being a simple list of ciphersuites. A nil Config.CipherSuites just means "what the team thinks is best", which might involve custom logic.

For example #21144 proposed making version-based decisions. TLS 1.3 might bring logic along, too.

About what you are trying to do, crypto/tls is just another user of crypto/cipher, so I think replicating the logic is at least as good as inferring it from the defaults.

bdarnell:
> I don't think we should commit to the default logic being a simple list of ciphersuites.

That's not what I'm asking for. It's already not a simple list: the order depends on whether hardware-accelerated AES is available. 

[...]

PR #30738, opened 2019-03-11
crypto/tls: expose implemented cipher suites

noneymous:
Allows to dynamically retrieve a list of available cipher suites. This change allows a variety of applicatoins:
- Enable all available ciphers programmatically, without hardcoding, and always refering to the latest set of implemented ciphers
- Filter ciphers by desired flags and use them dynamically, without hardcoding, and always refering to the latest set of implemented ciphers
- Build application configs for user-decided selection of ciphers to allow

Ciphers are returned as copies, in order to avoid messing up internal stuff.

In contrast to the already discussed solutions, this one reduced the amount of necessary changes. Maintainability will be equal to before. This should not contain any breaking changes, as it only makes former private variables public, respectively, adds functions.

Fixes #30325 and #21167

### DIFF
--- a/src/crypto/tls/cipher_suites.go
+++ b/src/crypto/tls/cipher_suites.go
@@ -14,6 +14,7 @@ import (
 	"crypto/sha1"
 	"crypto/sha256"
 	"crypto/x509"
+	"github.com/test/go-fork/fmt"
 	"hash"
 	"internal/x/crypto/chacha20poly1305"
 )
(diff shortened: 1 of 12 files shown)


noneymous:
I signed it!

noneymous:
I signed it!

agnivade:
Seems like this is the same as https://github.com/golang/go/pull/30739. Closing this.

PR #30877, opened 2019-03-16
net/http: document behaviour of cancellation on writable response bodies

### LINKED ISSUE #30876: net/http: document that context cancellation does not apply to writable bodies
https://github.com/golang/go/issues/26937

To enable http.ReverseProxy to cleanly support websocket connections, http.Transport began returning writable response bodies.

This is documented. But what's undocumented is how this response body deals with the request context's cancellation. 

See https://github.com/golang/go/issues/26101

Right now, the documentation says the cancellation of the request context will cause the response body to close, even if it is writable. However, this is not the case. If a writable response body is returned, cancellation of the request context does not affect it. This is certainly the right behaviour but should be documented.

nhooyr:
Closes #30876

### DIFF
--- a/src/net/http/request.go
+++ b/src/net/http/request.go
@@ -345,6 +345,10 @@ func (r *Request) Context() context.Context {
 // For outgoing client request, the context controls the entire
 // lifetime of a request and its response: obtaining a connection,
 // sending the request, and reading the response headers and body.
+//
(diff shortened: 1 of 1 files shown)


nhooyr:
Please reopen when there is discussion on #30876

PR #31009, opened 2019-03-22
log: expose Writer() method of the standard logger

pjediny:
The Go 1.12 introduced Writer() method for logger objects, but
it was not exposed as log package function for standard logger.
This commit adds such Writer() function.


pjediny:
CLA signed!

PR #31071, opened 2019-03-27
mime: add javascript module mime type (.mjs)

johnbalvin:
There are default mime types in this package for handling static content, but there's
a new one missing ".mjs"  that is "Content-Type: text/javascript".

https://developers.google.com/web/fundamentals/primers/modules#mjs


johnbalvin:
I signed it!

PR #31079, opened 2019-03-27
crypto/tls: add StrictVersionNegotiation config field for TLS 1.2

riraccuia:
This PR exposes a new configuration field 'StrictVersionNegotiation' to customize whether the version field in the record layer of the TLS 1.2 ClientHello should match or not the one advertised in the handshake layer.
The current default behavior just sets the version in the record layer to 1.0 to maintain compatibility with legacy/buggy TLS server implementations.

This problem is described in [RFC 5246 Appendix E.1](https://tools.ietf.org/html/rfc5246#appendix-E.1)

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -563,6 +563,15 @@ type Config struct {
 	// which is currently TLS 1.3.
 	MaxVersion uint16
 
+	// StrictVersionNegotiation only applies to TLS 1.2 ClientHello messages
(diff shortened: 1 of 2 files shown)


riraccuia:
I signed it!

dmagyar:
I would second this PR, the current client sends TLS1.0 in its' ClientHello regardless of what MinVersion/MaxVersion is. We ran into the issue explained in RFC 5246 Appendix E.1 before.

PR #31200, opened 2019-04-02
net/http: add method GetValue for cookie

sxs2473:
This commit add method GetValue for cookie to net/http.
It can easily get cookie value when use Request.Cookie.

### DIFF
--- a/src/net/http/cookie.go
+++ b/src/net/http/cookie.go
@@ -225,6 +225,14 @@ func (c *Cookie) String() string {
 	return b.String()
 }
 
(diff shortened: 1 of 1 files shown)


sxs2473:
I signed it!

agnivade:
Hi @sxs2473 - This adds new public API. Every public API addition needs to have a proposal where its discussed and approved.

I would request you to file an issue first, where we can discuss the problems that this PR is trying to solve and get a change like this approved.

Thanks.

PR #31356, opened 2019-04-09
Remove length check to prevent length guessing

ibudiallo:
in crypto/subtle ConstantTime is not always constant. When the two strings length don't match, the time is always lower.


ibudiallo:
I signed it

randall77:
With your modifications `ConstantTimeCompare` will panic if `len(y) < len(x)`. That seems bad.

ibudiallo:
> With your modifications `ConstantTimeCompare` will panic if `len(y) < len(x)`. That seems bad.

You are correct, I rewrote it to check the length and assign it to the smallest value to prevent a panic.

PR #31383, opened 2019-04-10
crypto/x509: use already parsed fields for certificate verification

kichankwon:
Distinguished Name attributes can be encoded in
various formats such as UTF-8 or PrintableString, and
each string format has their own tag byte.

Because of string type tag byte, verification can
be failed though the Distinguished Name of
child's Issuer is same with the certificate's Subject.

To ensure equality regardless of string format,
this CL now uses already parsed data fields for
the comparisons.


kichankwon:
I signed it!

PR #31418, opened 2019-04-11
expvar: improve Map.addKey for large number of keys

### LINKED ISSUE #31414: expvar: overhead cost of Map.Add when the number of entries grows

<pre>
$ go version
go version go1.12 darwin/amd64
</pre>

Yes.

$ go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/xkwei/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/xkwei/Code/GO"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/Cellar/go/1.12/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.12/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/s2/s4mtcvn50z798btptmrvvg3h0000gp/T/go-build853285173=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre></details>

### What did you do?
I tried to use `Map` in package the `expvar` to collect some metrics but I found it would be very slow to call `Add` method of it when the number of entries in the map grew. As for the concurrent case most of goroutines calling the `Add` of the same `Map` instance would be blocked.

And the source code of the `Add` method tells that in the `Add` method a sorting for all the keys will be executed holding the lock:
```
// addKey updates the sorted list of keys in v.keys.
func (v *Map) addKey(key string) {
	v.keysMu.Lock()
	defer v.keysMu.Unlock()
	v.keys = append(v.keys, key)
	sort.Strings(v.keys)
}

....

// Add adds delta to the *Int value stored under the given map key.
func (v *Map) Add(key string, delta int64) {
	i, ok := v.m.Load(key)
	if !ok {
		var dup bool
		i, dup = v.m.LoadOrStore(key, new(Int))
		if !dup {
			v.addKey(key)
		}
	}

	// Add to Int; ignore otherwise.
	if iv, ok := i.(*Int); ok {
		iv.Add(delta)
	}
}
```

1. It should be low-cost to add a key in the `Map` even if the number of the keys is large.
2. I think it is a waste to do sorting every time a new key is added. (Actually I guess it can be sorted only when output if the sorted outputs is necessary)
3. I think it is not smart to do a high cost operation(sorting) holding a lock. (Actually it is better to search and insert in this case).

### What did you see instead?
Holding the lock do a sort for all the entries.

ShiKaiWi:
The existing implementation has poor performance for large number of
keys because it chooses to append the new key to the sorted keys array
and sort the new array again.

The improvement tries to utilize the sorted keys array by searching the
index and doing an insertion if any. It also fixes an error that
duplicate keys may be inserted in high concurrent scenes.

Fixes #31414

### DIFF
--- a/src/expvar/expvar.go
+++ b/src/expvar/expvar.go
@@ -141,8 +141,14 @@ func (v *Map) Init() *Map {
 func (v *Map) addKey(key string) {
 	v.keysMu.Lock()
 	defer v.keysMu.Unlock()
-	v.keys = append(v.keys, key)
-	sort.Strings(v.keys)
+	// Using insertion sort to place key into the already-sorted v.keys.
+	if i := sort.SearchStrings(v.keys, key); i >= len(v.keys) {
+		v.keys = append(v.keys, key)
+	} else if v.keys[i] != key {
+		v.keys = append(v.keys, "")
+		copy(v.keys[i+1:], v.keys[i:len(v.keys)-1])
+		v.keys[i] = key
+	}
 }
 
 func (v *Map) Get(key string) Var {
--- a/src/expvar/expvar_test.go
+++ b/src/expvar/expvar_test.go
@@ -300,9 +300,9 @@ func BenchmarkMapSetDifferent(b *testing.B) {
 	})
 }
 
-// BenchmarkMapSetDifferentRandom simulates that the concerned keys of
-// Map.Set are generated dynamically and as a result insertion is out
-// of order and the number of the keys may be large.
+// BenchmarkMapSetDifferentRandom simulates such a case where the concerned
+// keys of Map.Set are generated dynamically and as a result insertion is
+// out of order and the number of the keys may be large.
 func BenchmarkMapSetDifferentRandom(b *testing.B) {
 	procKeys := make([][]string, runtime.GOMAXPROCS(0))
 	for i := range procKeys {
@@ -381,9 +381,9 @@ func BenchmarkMapAddDifferent(b *testing.B) {
 	})
 }
 
-// BenchmarkMapAddDifferentRandom simulates that the concerned keys of
-// Map.Add are generated dynamically and as a result insertion is out
-// of order and the number of the keys may be large.
+// BenchmarkMapAddDifferentRandom simulates such a case where that the concerned
+// keys of Map.Add are generated dynamically and as a result insertion is out of
+// order and the number of the keys may be large.
 func BenchmarkMapAddDifferentRandom(b *testing.B) {
 	procKeys := make([][]string, runtime.GOMAXPROCS(0))
 	for i := range procKeys {
--- a/src/expvar/expvar_test.go
+++ b/src/expvar/expvar_test.go
@@ -6,6 +6,7 @@ package expvar
 
 import (
 	"bytes"
+	"crypto/sha1"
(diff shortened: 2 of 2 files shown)


ShiKaiWi:
benchmark(old)
```
goos: darwin
goarch: amd64
pkg: std/expvar
BenchmarkMapSet-4                               20000000                96.4 ns/op            32 B/op          2 allocs/op
BenchmarkMapSetDifferent-4                       5000000               365 ns/op             128 B/op          8 allocs/op
BenchmarkMapSetString-4                         20000000               107 ns/op              32 B/op          2 allocs/op
BenchmarkMapAddSame-4                            2000000               624 ns/op             480 B/op         11 allocs/op
BenchmarkMapAddDifferent-4                       1000000              1585 ns/op            1088 B/op         31 allocs/op
BenchmarkMapAddDifferentRandom-4                   10000            170760 ns/op           20560 B/op        523 allocs/op
BenchmarkMapAddSameSteadyState-4                50000000                28.1 ns/op             0 B/op          0 allocs/op
BenchmarkMapAddDifferentSteadyState-4           20000000               102 ns/op               0 B/op          0 allocs/op
PASS
ok      std/expvar      16.344s
```
benchmark(new)
```
goos: darwin
goarch: amd64
pkg: std/expvar
BenchmarkMapSet-4                               20000000                93.9 ns/op            32 B/op          2 allocs/op
BenchmarkMapSetDifferent-4                       5000000               346 ns/op             128 B/op          8 allocs/op
BenchmarkMapSetString-4                         20000000                93.8 ns/op            32 B/op          2 allocs/op
BenchmarkMapAddSame-4                            3000000               521 ns/op             448 B/op         10 allocs/op
BenchmarkMapAddDifferent-4                       1000000              1526 ns/op             960 B/op         27 allocs/op
BenchmarkMapAddDifferentRandom-4                   50000             33741 ns/op           17361 B/op        423 allocs/op
BenchmarkMapAddSameSteadyState-4                50000000                32.6 ns/op             0 B/op          0 allocs/op
BenchmarkMapAddDifferentSteadyState-4           10000000               130 ns/op               0 B/op          0 allocs/op
PASS
ok      std/expvar      14.856s
```

And there is big improvement for the new added benchamrk `BenchmarkMapAddDifferentRandom`:
```
BenchmarkMapAddDifferentRandom-4(old)                  10000            170760 ns/op           20560 B/op        523 allocs/op
BenchmarkMapAddDifferentRandom-4(new)                  50000             33741 ns/op           17361 B/op        423 allocs/op
```

PR #31425, opened 2019-04-11
strings: fixed comment of TrimSpace function

torniker:
TrimSpace function returns just a string, but the comment had "slice of the string"


adelowo:
See https://go-review.googlesource.com/c/go/+/170104#message-156f9eda4f65da96364fe237de20b864dae90e67

PR #31527, opened 2019-04-17
cmd/go/internal/generate: stop premature variable substitution in commands

### LINKED ISSUE #31608: cmd/go generate prematurely does variable substitution in no-arg commands

<pre>
$ go version
go version go1.12 windows/amd64
</pre>

Yes.

<pre>
$ go env

C:\Users\selliott\Documents\Source\Github\GoLang\go\src>go env
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\xx\AppData\Local\go-build
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=C:\Users\xx\go
set GOPROXY=
set GORACE=
set GOROOT=C:\Tools\Go
set GOTMPDIR=
set GOTOOLDIR=C:\Tools\Go\pkg\tool\windows_amd64
set GCCGO=gccgo
set CC=gcc
set CXX=g++
set CGO_ENABLED=0
set GOMOD=C:\Users\xx\Documents\Source\Github\GoLang\go\src\go.mod
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=C:\Users\xx\AppData\Local\Temp\go-build857656854=/tmp/go-build -gno-record-gcc-switches
</pre></details>

### What did you do?

For a no-arguments command, generate incorrectly performs variable substitutions on the definition of the command, rather than on a copy of the command definition at each use.  This gives incorrect results when the values of the variables changes.

e.g. using bash syntax, write a temp file holding a value set for variable C.
```
//go:generate sh -c "echo export C='22' >setC.tmp"
//go:generate -command A  sh -c ". ./setC.tmp;echo B${C}D"
//go:generate A
//go:generate sh -c "echo export C='66' >setC.tmp"
//go:generate A
```
```
B22D
B66D
```
### What did you see instead?
```
B22D
B22D
```

QuasarSE:
go:generate commands passed no arguments are currently subject 
to premature variable substitution due to mistakenly assuming append 
guarantees a copy.  The change fixes this by forcing a slice copy at 
each invocation of a command.

The previous code assumed that append would always generate a 
copy of its inputs. However, append wouldn't create a copy if there was
no need to increase capacity and it would just return the original
input slice. This resulted in premature variable substitutions in
the "master word list" of generate commands, thus yielding incorrect
results across multiple invocations of the same command when the
body contained substitutions e.g. environment variables, moreover
these can change during the lifetime of go:generate processing a
file.

Note that this behavior would not manifest itself if any arguments were 
passed to the command, because append would make a copy of the slice 
as it needed to increase its capacity.   The "hacky" work-around was to
always pass at least one argument to any command, even if the 
command ignores it.  e.g.,
       //go:generate MyNoArgsCmd ' '

This CL fixes that issue and removes the need for the hack mentioned
above.

Fixes #31608

### DIFF
--- a/src/cmd/go/internal/generate/generate.go
+++ b/src/cmd/go/internal/generate/generate.go
@@ -374,15 +374,10 @@ Words:
 	// Substitute command if required.
 	if len(words) > 0 && g.commands[words[0]] != nil {
 		// Replace 0th word by command substitution.
-		//  If no other words besides the command are on the invocation
-		//  line, e.g.,
-		//       //go:generate MY-COMMAND
-		//  append wouldn't force a copy as no increase in capacity was
-		//  needed.  Instead, it would instead return a reference to
-		//  the actual g.commands value (the original word array).  We
-		//  don't want the original word list to be subjected to os.Expand
-		//  below, since the words in the command could contain changing
-		//  environment variables, etc., so force a copy.
+		//
+		// Force a copy of the command definition to
+		// ensure words doesn't end up as a reference
+		// to the g.commands content.
 		tmpCmdWords := append([]string(nil), (g.commands[words[0]])...)
 		words = append(tmpCmdWords, words[1:]...)
 	}
--- a/src/cmd/go/internal/generate/generate.go
+++ b/src/cmd/go/internal/generate/generate.go
@@ -374,17 +374,16 @@ Words:
 	// Substitute command if required.
 	if len(words) > 0 && g.commands[words[0]] != nil {
 		// Replace 0th word by command substitution.
-	
+
 		// ... If no other words besides the command are on the invocation
 		// ... line, e.g.,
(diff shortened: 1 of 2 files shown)


QuasarSE:
CLA now signed.

QuasarSE:
Review suggestions incorporated.   Tests added -- will fail with existing code, pass with new.

PR #31596, opened 2019-04-21
syscall: avoid zeroing unused syscall arguments

coypoop:


coypoop:
Added all occurrences of the same code, but I only tested on netbsd, freebsd and linux.

Adam8989 (approved):

coypoop:
It's code that serves no functional purpose. It makes the code very slightly smaller and very slightly faster at no cost. free win :-)

agnivade:
As mentioned above, replying to the github thread does not pass on the comment to gerrit. Please comment on the gerrit CL. Otherwise people will not know that you have replied.

PR #31614, opened 2019-04-22
cmd/go: derive executable name from package path in 'go run'

### LINKED ISSUE #31571: cmd/go: go run . uses wrong executable name
If I am in a directory x/y/z (in module mode, but maybe in GOPATH too)
such that the import path of that directory is m/x/y/z, and there is
a single file foo.go in the directory, and I run `go run .`, the name of
the binary that is built and run should be `z`. Right now it is `foo`.
This matters because it shows up in `ps`, etc.

ghost:
Change name of temporary executable on go run . to directory name.
Fixes #31571

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "go.formatTool": "goimports"
-}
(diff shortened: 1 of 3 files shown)


Liberatys:
I signed it!

Liberatys:
I signed it!

Liberatys:
Check

PR #31627, opened 2019-04-23
net: introduce ListenConfig#AfterBind

### LINKED ISSUE #11822: net Listen not support unixsocketperm
when use unix socket Listen:
`net.Listen("unix", addr)`
can't set unixsocketperm 777.

many reverse proxy such as nginx return http 502 when sock file not 777 permission.

Al2Klimov:
fixes #11822

### DIFF
--- a/src/net/dial.go
+++ b/src/net/dial.go
@@ -597,6 +597,14 @@ type ListenConfig struct {
 	// Listen will cause the Control function to be called with "tcp4" or "tcp6".
 	Control func(network, address string, c syscall.RawConn) error
 
+	// If AfterBind is not nil, it is called after creating the network
+	// connection and binding it to the operating system.
+	//
+	// Network and address parameters passed to Control method are not
+	// necessarily the ones passed to Listen. For example, passing "tcp" to
+	// Listen will cause the Control function to be called with "tcp4" or "tcp6".
+	AfterBind func(network, address string, c syscall.RawConn) error
+
 	// KeepAlive specifies the keep-alive period for network
 	// connections accepted by this listener.
 	// If zero, keep-alives are enabled if supported by the protocol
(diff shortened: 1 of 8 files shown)


JeremyRand:
It's not obvious to me why this was closed; could someone please clarify? The two suggestions given in https://github.com/golang/go/issues/11822 are unworkable for a variety of use cases; `syscall.Umask(0077)` will cause side effects in multithreading, while `os.Chmod` after listening will produce a TOCTOU security vulnerability.

If the problem is purely procedural (i.e. someone needs to open a new issue requesting this) then I am happy to do it, but it's not at all obvious from the history of this PR that that's the blocker.

ianlancetaylor:
@JeremyRand We use Gerrit for code review. Very few people see comments on GitHub pull requests. The place to comment is https://go.dev/cl/173359. Thanks.

Also this change introduces new API, which requires going through the proposal process as described at https://go.dev/s/proposal.

I don't entirely understand how this fixes the problem. Isn't there still a TOCTOU race?

JeremyRand:
> I don't entirely understand how this fixes the problem. Isn't there still a TOCTOU race?

@ianlancetaylor The problem with the current standard library behavior is that `bind` and `listen` are done together. If you `bind` but don't `listen`, the socket is in a state where `connect` always gets rejected. So if you `chmod` the socket *between* `bind` and `listen`, that prevents the vulnerability. Which is what I think this PR was attempting to fix.

Would you like me to open a new issue to discuss this, so that there can be a discussion about what changes might or might not be appropriate? I am not particularly attached to the API in this PR, and I'd honestly prefer something that doesn't require as much cleverness on the user's end.

ianlancetaylor:
Thanks for the explanation.

Please do go ahead and open a new issue. Thanks.

PR #31647, opened 2019-04-24
golang proxy request incorrect

uucckk:
when use proxy to request nginx, nginx can not be use correctly, After I rewrite some code in readRequest method then it can be succeed. The code I have already upload. So you can take a look. Waiting for your reply. Thank you

So if I run the following code, ngxin won't report that it can't find "project error"
```javascript
	remote, err := url.Parse(p.path)
			if err != nil {
				panic(err)
			}
			proxy := httputil.NewSingleHostReverseProxy(remote)
			req.URL.Path = Substring(urlPath.Path, StringLen(p.pattern), -1)


			if req.Method == "OPTIONS" {
				w.Header().Set("Access-Control-Allow-Origin", "*") //
				w.Header().Add("Access-Control-Allow-Headers", "content-type, accept, x-auth-token, X-Subject-Token,x-openstack-nova-api-version")
				w.Header().Add("Access-Control-Allow-Methods", "POST")
				w.Write([]byte("{\"test\":\"OPTIONS\"}"))
				return true
			}
			w.Header().Add("Access-Control-Allow-Origin", "*") //x-openstack-nova-api-version
			w.Header().Add("Access-Control-Allow-Headers", "content-type, accept, x-auth-token, X-Subject-Token,x-openstack-nova-api-version")
			w.Header().Add("Access-Control-Expose-Headers", "X-Subject-Token, x-openstack-nova-api-version,X-Auth-Token")
			w.Header().Add("Access-Control-Allow-Methods", "PUT,POST,GET,DELETE,OPTIONS")
			w.Header().Add("Access-Control-Allow-Credentials", "true")
			proxy.ServeHTTP(w, req)
		}
```

the Gerrit change.

  + No trailing period
+ No Markdown

### DIFF
--- a/src/net/http/request.go
+++ b/src/net/http/request.go
@@ -1029,10 +1029,10 @@ func readRequest(b *bufio.Reader, deleteHostHeader bool) (req *Request, err erro
 	//	GET http://www.google.com/index.html HTTP/1.1
 	//	Host: doesntmatter
 	// the same. In the second case, any Host line is ignored.
-	req.Host = req.URL.Host
-	if req.Host == "" {
-		req.Host = req.Header.get("Host")
-	}
+	//req.Host = req.URL.Host
+	//if req.Host == "" {
+	//	req.Host = req.Header.get("Host")
+	//}
 	if deleteHostHeader {
 		delete(req.Header, "Host")
 	}


uucckk:
check

PR #31709, opened 2019-04-27
Add 'EmptyValue' interface, empty if element is struct and all fields within it is also empty.

WWWSpider:
1.Add EmptyValue interface for json and xml to specify if it is a empty value. Useful for struct such as time.Time, empty if IsZero().
2.Empty if element's tag has "omitempty" and it is struct, map, slice and all fields within it is also empty, even like "schools>school,omitempty".

### DIFF
--- a/src/encoding/json/encode.go
+++ b/src/encoding/json/encode.go
@@ -233,6 +233,12 @@ type UnsupportedTypeError struct {
 	Type reflect.Type
 }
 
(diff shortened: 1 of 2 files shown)


WWWSpider:
I signed it!

seanb-c:
@WWWSpider, thanks for adding this. I was looking into doing the same thing and found this PR.

Issue #13192, opened 2015-11-09
cmd/link:  ppc64 (big endian)  cgo errors 


laboger:
I've built golang from master with the patches from issue 11184 to get external linking to work with ppc64le.  That seems to work well on ppc64le with external linking and cgo.

In cmd/dist/build.go, linux/ppc64le is in the cgoEnabled map but linux/ppc64 is not.  So the build of golang with the latest patches on ppc64 does not quite work with cgo.

If I build the golang toolchain on ppc64 with CGO_ENABLED=1, I first get this error:

cannot use dynamic imports with -d flag

I made a change to cmd/link/internal/ppc64/obj.go to get rid of this message, but then hit this error:
/home/boger/golang/gitsrc/latest/go/pkg/linux_ppc64/net.a(_all.o): unknown relocation type 51; compiled without -fpic?
......
runtime/cgo(.opd): unexpected relocation type 307
.......
and then too many errors

I can see that relocation type 51 is R_PPC64_TOC and that is not handled by the code.  I tried adding those defines but not sure what should be generated for this relocation type.  If there are suggestions on what to do I can try them out.

laboger:
I've been doing this on a RHEL7.2 system.

mwhudson:
Those errors suggest that you're not actually linking externally. Try
adding -ldflags=-linkmode=external to your go tool invocation.

I'm pretty sure that things still won't work though.

laboger:
These errors happen during the build of golang on ppc64:

Under ##### Building packages and commands for linux/ppc64.
.....
cmd/pprof
net/rpc/jsonrpc
# cmd/trace
/home/boger/golang/gitsrc/latest/go/pkg/linux_ppc64/net.a(_all.o): unknown relocation type 51; compiled without -fpic?
/home/boger/golang/gitsrc/latest/go/pkg/linux_ppc64/net.a(_all.o): unknown relocation type 51; compiled without -fpic?
/home/boger/golang/gitsrc/latest/go/pkg/linux_ppc64/net.a(_all.o): unknown relocation type 51; compiled without -fpic?
/home/boger/golang/gitsrc/latest/go/pkg/linux_ppc64/net.a(_all.o): unknown relocation type 51; compiled without -fpic?
/home/boger/golang/gitsrc/latest/go/pkg/linux_ppc64/net.a(_all.o): unknown relocation type 51; compiled without -fpic?
[...]

PR #31738, opened 2019-04-29
cmd/link, runtime: reserve 48 bytes in stack frame for ppc64

zte-majiang:
C functions requires 48 bytes of space on caller stack frame for ppc64.  Without this fix, a simple cgo program(build with CGO_CFLAGS="-O0") could mess the stack and finally trigger a segmentation fault.

Updates #13192


laboger:
Should this be linux only?

Helflym:
I don't think you need to change the whole Go stack. The Go stack is relatively independent from the C stack as far as I know. You should be able to directly use `cgoCalleeStackSize` and `asmcgocallSaveOffset` constantes as it's done for AIX. 
Changing the FIXED_FRAME seems a bit risky and if it's needed anyway, I'll have to check on AIX. But I'm not sure it will work. 
@zte-majiang could you try without changing FIXED_FRAME and keep us updated ?

zte-majiang:
@Helflym Thanks for the advice. I have tried the plan(only adjust the parts of go stack) some days ago, it seemed not working. 
A quick search show that there are more than 3 pieces of code that try to set the minimum frame size. So I gave up the plan (which adjust go stack only), and turn to make sure the minimum frame size is always 48 on both side.
In my opinion, it's more clear to keep C and GO stack frames consistent. Or else, there will be two minimum frame sizes, one for GO which is 32, and one for C which is 48. This might make things complex(for example, stack backtrace? I'm not sure).
I will have a try as you suggested tomorrow.

Issue #31855, opened 2019-05-06
net/http: allow ListenAndServeTLS to omit PEM file values if TLSConfig.GetConfigForClient set


TheTesla:
If I set GetConfigForClient and start the server using ListenAndServeTLS the arguments of ListenAndServeTLS, the server certificate and its key are lneeded to be given and they are also loaded. I have set the server certificate and its key in the configuration TLSConfig served by GetConfigForClient before. So using GetConfigForClient ListenAndServeTLS should ignore its arguments and they should also be optional.

mikedanese:
What config should be used if the client doesn't send a hello?

seankhliao:
Done in [CL 578396](https://go.dev/cl/578396)

PR #31867, opened 2019-05-06
GetConfigForClient replaces cert and key arguments

TheTesla:
This addresses my issue #31855. This is not tested yet and not guarantied be fully complient.


TheTesla:
I signed it!

TheTesla:
I wait for feedback and test. Do I have to do something else?

PR #32098, opened 2019-05-17
encoding/json: Fix time is empty and has value

guonaihong:
```golang
package main

import (
	"encoding/json"
	"fmt"
	"time"
)

type Object struct {
	X int
	Y int
}

func (o *Object) IsZero() bool {
	return o.X == 0 && o.Y == 0
}

type Test struct {
	StartTime time.Time `json:"startTime,omitempty"`
	Obj       Object    `json:"obj,omitempty"`
	Z         int
}

func main() {
        //all, err := json.Marshal(&Test{})
	all, err := json.MarshalFilterStruct(&Test{}, func(v interface{}) bool {

		if v, ok := v.(Object); ok {
			if v.IsZero() {
				return true
			}
		}

		return false
	})

	fmt.Printf("err= %v, %s\n", err, all)

}
```
Marshal:
err= <nil>, {"startTime":"0001-01-01T00:00:00Z","obj":{"X":0,"Y":0},"Z":0}

MarshalFilterStruct:
err= <nil>, {"Z":0}

### DIFF
--- a/src/encoding/json/encode.go
+++ b/src/encoding/json/encode.go
@@ -171,6 +171,22 @@ func Marshal(v interface{}) ([]byte, error) {
 	return buf, nil
 }
 
+//todo
+func MarshalFilterStruct(v interface{}, filterFunc FilterFunc) ([]byte, error) {
+	e := newEncodeState()
+
+	err := e.marshal(v, encOpts{escapeHTML: true, filterFunc: filterFunc})
+	if err != nil {
+		return nil, err
+	}
+	buf := append([]byte(nil), e.Bytes()...)
+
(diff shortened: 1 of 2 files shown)


guonaihong:
Thank you for reminding me, I will try to satisfy  in the next commit [#11939]

guonaihong:
Thank you Daniel Martí for review.
I am new to an issue or to discuss it in #11939. What is your suggestion?

agnivade:
I think it is good to comment on #11939 with your new approach as that proposal still hasn't decided on any fixed approach, so alternatives are open.

Issue #31913, opened 2019-05-08
cmd/go: GOPROXY list should not terminate on non 404/410 errors


amnonbc:
The new GOPROXY list feature allows the specification of a series of proxies. If the first one returns a 404/410 status, we will try to fetch from the next one.

But if any other error is returned, e.g. a 500, then the search will terminate and the download will fail.
This limits our ability to use multiple proxies to improve make importing more robust and allow downloads to succeed even if a subset of the proxies are down or misbehaving.

We also can not rely on proxies to be well behaved in the errors they return. The athens proxy returns a 500 status if it can not authenticate against the upstream repository. This should of course be fixed in athens. But it would be useful if this feature allows us to work around such problems.

bcmills:
A 500 error indicates that the server for some reason didn't know how to handle the request. We have no way to determine whether it would have rejected the module or simply didn't know about it. In that case, there is a conservative (privacy-preserving) option, which is to assume that it would have rejected it, and an aggressive (error-tolerant) option, which is to assume that the proxy simply didn't know about it.

For a comma-separated list of proxies in `GOPROXY`, we have chosen the option that biases toward preserving the privacy of paths. That was discussed and decided on #26334, and absent new information (such as empirical observations of the reliability of generally-available proxy implementations), I don't think we should revisit that decision.

If you want to compose multiple proxies using some other policy, it should be fairly trivial to write a local `GOPROXY` server that implements whatever error behavior you like.

bcmills:
See [related discussion](https://go-review.googlesource.com/c/go/+/173441/4#message-5eb2aa3e9f0d98b60705706b169b943cbc531d93) on CL 173441 (#26334).

beoran:
In our proxy I made it so that it returns 404 if the upstream repo couldn't be checked out, exactly because of this way go works. So the current behavior of go helps proxy developers to do the Right Thing (tm). I think  Athens should make the same fix, it's really normally just small a 15 minutes fix.

PR #32199, opened 2019-05-23
goproxy-forward-util-succuss-or-last

storyicon:
`GOPROXY` on the master node promises developers [this](https://github.com/golang/go/blob/master/src/cmd/go/internal/modfetch/proxy.go#L43)：
```
For each request, the go command tries each proxy in sequence,
only moving to the next if the current proxy returns a 404 or 410 HTTP response.
```
But after a rough look at the code, I think the code on the current master node may not work as expected.       
       
What's more, I think this logic is not in line with the developer's expectations, `GOPROXY` should try all the proxies in its list until one of them does not return an error.                  

Even if any of the proxies in the `GOPROXY` list have an error, then golang’s attempt to other proxies should not be interrupted.         

The fact is that most goproxy providers (such as `athens`, `goproxy.io`) return 500 when the server fails to pull locally.            

More about it: https://github.com/storyicon/gos/blob/master/docs/on-goproxy.md

### DIFF
--- a/src/cmd/go/internal/modfetch/proxy.go
+++ b/src/cmd/go/internal/modfetch/proxy.go
@@ -40,7 +40,7 @@ control systems. Setting GOPROXY to "off" disallows downloading modules from
 any source. Otherwise, GOPROXY is expected to be a comma-separated list of
 the URLs of module proxies, in which case the go command will fetch modules
 from those proxies. For each request, the go command tries each proxy in sequence,
-only moving to the next if the current proxy returns a 404 or 410 HTTP response.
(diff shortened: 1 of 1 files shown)


storyicon:
@bcmills  Now I think that only when an proxy in the `GOPROXY` list returns `403`/`200`, golang's attempt on the next proxy should be interrupted. Do you have any ideas?

bcmills:
The behavior introduced by this change does not maintain the privacy-preserving characteristics of GOPROXY as designed — see #31913 for more detail.

PR #32224, opened 2019-05-24
unicode: turn script names into public constants

jeremieSimon:
The idea here is to be able to reuse the script names outside of the unicode package and avoid potential typos.


jeremieSimon:
I signed it!

jeremieSimon:
I signed it!

PR #32315, opened 2019-05-29
net/http: never cancel r.Context() after Hijack

### LINKED ISSUE #32314: net/http: reading from hijacked bufio.Reader should not cause r.Context to be cancelled
I was working on https://nhooyr.io/websocket when I noticed that when the connection returned from Hijack is closed, r.Context() is cancelled. This is bizarre behaviour given after Hijack, there should be no way for net/http to cancel the request context. After all, how could it know the connection is closed?

I tracked down the issue to the bufio.Reader returned from Hijack. When a read errors, it causes r.Context to become cancelled.

In my library I'm using the returned bufio.Reader for all reads. When a goroutine would close the connection, another goroutine would read from the connection and so due to this [line](https://github.com/golang/go/blob/cd24849e8b6c8a8079613cdb8a61fcc3e24f2154/src/net/http/server.go#L790), the context would be cancelled.

I think its best that reads on the bufio.Reader not cause r.Context to be cancelled as after Hijack, net/http should not be involved at all, especially not magical cancellation.

nhooyr:
Fixes #32314

### DIFF
--- a/src/net/http/serve_test.go
+++ b/src/net/http/serve_test.go
@@ -6166,6 +6166,46 @@ func fetchWireResponse(host string, http1ReqBody []byte) ([]byte, error) {
 	return ioutil.ReadAll(conn)
 }
 
+func TestServerContextAfterHijack(t *testing.T) {
+	done := make(chan struct{})
+	s := httptest.NewServer(HandlerFunc(func(w ResponseWriter, r *Request) {
+		defer close(done)
+
+		w.WriteHeader(StatusOK)
+
+		hj, ok := w.(Hijacker)
+		if !ok {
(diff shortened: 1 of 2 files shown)


nhooyr:
@odeke-em feel free to take over, I don't have the time to respond to the reviews

odeke-em:
Hey @nhyor, I think you copied the test that I provided verbatim, it won't
have the trybot failures and also matches the style for code review :) Just
that change and it'll be perhaps good to go.

odeke-em:
*I think if you copy the test that I provided verbatim

Sorry was typing from my phone

PR #32522, opened 2019-06-10
encoding/xml: use a depth counter rather than recursion in Decoder.Skip()

dbolkensteyn:
The `depth` counter technique is already used in the
`unmarshalTextInterface()` function.

My expectation is for the recursion removal to yield
a marginal performance gain, especially on deeply nested
XML documents, however I did not verify this.

The behavior of the `Skip()` function should not be
altered in any way by this change.

### DIFF
--- a/src/encoding/xml/read.go
+++ b/src/encoding/xml/read.go
@@ -738,18 +738,18 @@ Loop:
 // It returns nil if it finds an end element matching the start
 // element; otherwise it returns an error describing the problem.
 func (d *Decoder) Skip() error {
-	for {
-		tok, err := d.Token()
+	depth := 1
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
Should the doc comment be adjusted as well?

dbolkensteyn:
That's a good question @AlexanderYastrebov and I was asking myself the same:
* Nested structures are still skipped, so the previous description is not entirely wrong
* However descriptions that don't reveal implementation details seem preferable to me

I don't have a strong opinion myself on this. I tried to rephrase that sentence a couple time but failed to come up with something better, so I ended up leaving it unchanged for the time being.

PR #32788, opened 2019-06-26
misc/wasm: make wasm_exec.js compatible with create-react-app

DeedleFake:
I've been using some Go-generated WASM in a React project, so I modifed `wasm_exec.js` to be possible to import using Webpack without warnings.

Fatal warnings fixed:

  Line 19:   Unexpected use of 'self'  no-restricted-globals
  Line 20:    Unexpected use of 'self'  no-restricted-globals
  Line 20:   Unexpected use of 'self'  no-restricted-globals
  Line 248:   'fs' is not defined       no-undef
  Line 515:  'Go' is not defined       no-undef
  Line 519:  'fs' is not defined       no-undef

Non-fatal warnings fixed:

  Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

  Line 40:   Expected '!==' and instead saw '!='  eqeqeq
  Line 163:   Expected a default case              default-case
  Line 188:   Expected a default case              default-case
  Line 258:  Missing '()' invoking a constructor  new-parens

### DIFF
--- a/misc/wasm/wasm_exec.js
+++ b/misc/wasm/wasm_exec.js
@@ -11,22 +11,21 @@
 	// - Node.js
 	// - Electron
 	// - Parcel
+	// - Webpack
 
 	if (typeof global !== "undefined") {
 		// global already exists
 	} else if (typeof window !== "undefined") {
 		window.global = window;
+		// eslint-disable-next-line no-restricted-globals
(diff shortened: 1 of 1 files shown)


DeedleFake:
I signed it.

PR #32924, opened 2019-07-03
changed the direction of split of the format string when parsing name…

kyson3:
**encoding/json: changed the direction of split of the format string when parsing namespace and tag.**

For xml where the namespace contained a space character, functions marshal/unmarshal returned an error.

Sample of problem xml and struct:

<Authenticate  xmlns="ORION PRO">
</Authenticate>

type Authenticate struct {
     XMLName   xml.Name `xml:"ORION PRO Authenticate"`
}

Out namespace == "ORION", tag == "PRO Authenticate"


mvdan:
The CLA was never signed, and this is lacking tests and a bug report. Closing for now.

PR #32947, opened 2019-07-04
net/http: stop ExampleServer_Shutdown from hanging on error

eest:
Running the example code when not having permissions
to bind to port 80 will cause the program to hang after
printing the error message.


eest:
I signed it!

PR #32954, opened 2019-07-05
net/url: fail TestParseErrors test when getting an unwanted error

stefanb:
The TestParseErrors test function was not strict with unwanted errors 
received from url.Parse(). It was not failing in such cases, now it does.


stefanb:
I noticed this when adding test cases for #29384

PR #32998, opened 2019-07-09
prototype try builtin

ccbrown:
This is a prototype implementation of the `try` builtin proposal:

https://github.com/golang/proposal/blob/master/design/32437-try-builtin.md

As far as I know it's fully functional (all tests pass), but I've made this strictly for "academic" purposes and can't make any guarantees about it.


ccbrown:
Ah, crap. I meant to open this PR on my own fork, not upstream Golang. My bad.

PR #33172, opened 2019-07-18
cmd/go: look for .go.buildinfo section when looking for ELF version

### LINKED ISSUE #31861: cmd/go: go version <binary> fails with -buildmode=pie Linux ELF binaries
$ go build -o /tmp/go.exe
$ go version /tmp/go.exe
/tmp/go.exe: go version not found
$ 

It's possible I am using too old a tree but I think this is still broken at head.

Infinoid:
Fixes #31861


Infinoid:
I signed a CLA.

PR #33239, opened 2019-07-23
io: add error check on pipe close functions to avoid error overwriting

### LINKED ISSUE #24283: io: Close and CloseWithError overwrite the error stored in a pipe
Calling `Close` after `CloseWithError` (or `CloseWithError` multiple times) overwrites the stored error in `io.pipe`. 

Opening an issue for discussion first, since there is test coverage verifying this behavior, though it seems to only be there to check the `atomic.Value`, and it does subtly change the API. 

Since pipes are often used concurrently, and multiple calls to `Close` _usually_ have no effect, users would often expect the following code to be deterministic:

	pr, pw := io.Pipe()        
	go func() {
		defer pw.Close()
		_, err := io.Copy(pw, reader)
		if err != nil {
			pw.CloseWithError(err)
		}
		// more code
	}()

But in this case, if there is an error and the `PipeReader` executes `Read` immediately after `CloseWithError` it will get the stored error from the `Copy`, but once `defer pw.Close` executes the error is set back to `EOF`. 

Moving the error assignment into the `pipe.once` would make the above work, ~and probably also allow the removal of the `sync.Value` fields, since the errors are protected by the done chan and the `sync.Once`.~

jmartin82:
The current implementation allows multiple calls `Close` and `CloseWithError` in every side of the pipe, as a result, the original error can be overwritten.

This CL fixes this behavior adding an error existence check on `atomicError` type
and keeping the first error still available.

Fixes #24283

### DIFF
--- a/src/io/pipe.go
+++ b/src/io/pipe.go
@@ -10,28 +10,26 @@ package io
 import (
 	"errors"
 	"sync"
-	"sync/atomic"
 )
 
-// atomicError is a type-safe atomic value for errors.
-// We use a struct{ error } to ensure consistent use of a concrete type.
-type atomicError struct {
-	v atomic.Value
-	sync.Mutex
+// onceError is an object that will only store an error once.
+type onceError struct {
+	sync.Mutex // guards following
+	err        error
 }
 
-func (a *atomicError) Store(err error) {
+func (a *onceError) Store(err error) {
 	a.Lock()
 	defer a.Unlock()
-	prvErr := a.Load()
-	if prvErr != nil {
+	if a.err != nil {
 		return
 	}
-	a.v.Store(struct{ error }{err})
+	a.err = err
 }
-func (a *atomicError) Load() error {
-	err, _ := a.v.Load().(struct{ error })
-	return err.error
+func (a *onceError) Load() error {
+	a.Lock()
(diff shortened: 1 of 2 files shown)


av86743:
@jmartin82 

https://github.com/golang/go/blob/master/src/io/pipe.go#L73
https://github.com/golang/go/blob/master/src/io/pipe.go#L112

1. Do you really need to keep `p.erMu` locked during `p.once.Do(func() { close(p.done) })`?

2. Do you need `p.once.Do(...)` at all?  As you seem to disallow repeated `p.close()`.

jmartin82:
@av86743 Thanks for your fast review.

> 1. Do you really need to keep `p.erMu` locked during `p.once.Do(func() { close(p.done) })`?
No, you are right I can unlock it before.

> 2. Do you need `p.once.Do(...)` at all?  As you seem to disallow repeated `p.close()`.
It's a bit tricky because you can close the pipe from both sides (read/write) but you only have to close the channel once. if I remove the `once` I have to find an alternative (with a flag for example) to keep the same behavior or only allow to close one time the pipe, without matter if is the read or write side.

jmartin82:
@av86743 Solved both things, thanks!

PR #33248, opened 2019-07-23
runtime: fix gdb pretty print for slices

elbeardmorez:
gdb version sha:e86ca25fd6cc06488c9dda936cf83f0fcdf6b58b (17-Apr-2018)
removed TYPE_TAG_NAME in favour of TYPE_NAME to 'simplify' usage. the
fallout due to the new implementation in 'gdb/python/py-type.c'
resulting in premature removal of the 'struct' identifier that the go
'SliceTypePrinter' was using to match on

fix adds a trivial 'TypePrinter' base class to allow for a specialised
'match()' in the 'SliceTypePrinter' class which considers the type's
structure in addition to its (now simplified) type name

-----

Bisecting gdb resulted in this being the pivot:
```
commit e86ca25fd6cc06488c9dda936cf83f0fcdf6b58b (refs/bisect/bad)
Author: Tom Tromey <tom@tromey.com>
Date:   Tue Apr 17 13:51:24 2018 -0600

    Remove TYPE_TAG_NAME
```
I believe the relevant part is as follows:
```diff
iff --git a/gdb/python/py-type.c b/gdb/python/py-type.c
index 80055b05e0..dd8c0aa9dc 100644
--- a/gdb/python/py-type.c
+++ b/gdb/python/py-type.c
@@ -416,10 +416,16 @@ static PyObject *
 typy_get_tag (PyObject *self, void *closure)
 {
   struct type *type = ((type_object *) self)->type;
+  const char *tagname = nullptr;
 
-  if (!TYPE_TAG_NAME (type))
+  if (TYPE_CODE (type) == TYPE_CODE_STRUCT
+      || TYPE_CODE (type) == TYPE_CODE_UNION
+      || TYPE_CODE (type) == TYPE_CODE_ENUM)
+    tagname = TYPE_NAME (type);
+
+  if (tagname == nullptr)
     Py_RETURN_NONE;
-  return PyString_FromString (TYPE_TAG_NAME (type));
+  return PyString_FromString (tagname);
 }
 /* Return the type, stripped of typedefs. */
```
### good (pre-breaking change)
```
(gdb) ptype
type = struct []int32 {
    int32 *array;
    int len;
    int cap;
}
(gdb) p arr
$3 = []int32 = {1, 2, 4, 5, 7, 8, 10}

(gdb) python import pdb
(gdb) python pdb.run('gdb.execute("print arr")')
> <string>(1)<module>()
..
(Pdb) n
> /usr/lib/go/src/runtime/runtime-gdb.py(225)matcher()
-> if klass.pattern.match(str(val.type)):
(Pdb) p str(val.type)
'struct []int32'
(Pdb)
```
### bad (e86ca25fd6cc06488c9dda936cf83f0fcdf6b58b)
```
(gdb) ptype arr
type = struct []int32 {
    int32 *array;
    int len;
    int cap;
}
(gdb) p arr
$1 = {array = 0xc0000142a0, len = 7, cap = 8}

(gdb) python import pdb
(gdb) python pdb.run('gdb.execute("print arr")')
...
(Pdb) n
-> def matcher(val):
(Pdb) p val
<gdb.Value object at 0x7f7e8b9ba070>
(Pdb) print(val)
{array = 0xc00001a2a0, len = 7, cap = 8}
...
(Pdb) n
> /usr/lib/go/src/runtime/runtime-gdb.py(225)matcher()
-> if klass.pattern.match(str(val.type)):
(Pdb) p klass
<class '__main__.SliceTypePrinter'>
(Pdb) p klass.pattern
re.compile('^struct \\[\\]')
(Pdb) p str(val.type)
'[]int32'
(Pdb) gdb.execute('ptype arr')
type = struct []int32 {
    int32 *array;
    int len;
    int cap;
}
```
The difference is in the `p str(val.type)` result, the `"struct "` prefix has gone. Following this patch the original behaviour is restored.

Testing environment used go v1.12.6, and dozens of gdb versions up to HEAD.

### DIFF
--- a/src/runtime/runtime-gdb.py
+++ b/src/runtime/runtime-gdb.py
@@ -98,8 +98,16 @@ class SliceValue:
 #  Pretty Printers
 #
 
+class TypePrinter:
 
-class StringTypePrinter:
+	pattern = None
+
+	@classmethod
+	def match(cls, type_):
+		return cls.pattern.match(str(type_)) if cls.pattern else False
+
+
+class StringTypePrinter(TypePrinter):
 	"Pretty print Go strings."
 
 	pattern = re.compile(r'^struct string( \*)?$')
@@ -115,10 +123,11 @@ class StringTypePrinter:
 		return self.val['str'].string("utf-8", "ignore", l)
 
 
-class SliceTypePrinter:
+class SliceTypePrinter(TypePrinter):
 	"Pretty print slices."
 
-	pattern = re.compile(r'^struct \[\]')
+	pattern = re.compile(r'^(?:struct )?\[\]')
+	strip = re.compile(r'^struct ')
 
 	def __init__(self, val):
 		self.val = val
@@ -127,7 +136,7 @@ class SliceTypePrinter:
 		return 'array'
 
 	def to_string(self):
-		return str(self.val.type)[6:]  # skip 'struct '
+		return self.strip.sub('', str(self.val.type))  # strip any 'struct ' prefix
 
 	def children(self):
 		sval = SliceValue(self.val)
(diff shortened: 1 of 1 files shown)


elbeardmorez:
I signed it!

PR #33291, opened 2019-07-26
reflect/DeepEqual: remove redundant type equality check

darshanime:
This change modifies Go to remove type check in [`DeepEqual`](https://github.com/golang/go/blob/919594830f17f25c9e971934d825615463ad8a10/src/reflect/deepequal.go#L193) since it also being done in [`deepValueEqual`](https://github.com/golang/go/blob/919594830f17f25c9e971934d825615463ad8a10/src/reflect/deepequal.go#L28)


darshanime:
I signed it!

PR #33331, opened 2019-07-28
runtime: add race check for buf in slicebytetostring.

anjiawei1991:
I noticed there's no race check for `buf` in function `slicebytetostring`.

reusing `rawstringtmp` can fix it. And by this way, we simplified the codes.


anjiawei1991:
I signed it!

PR #33337, opened 2019-07-29
cmd/compile: output err when append only has only one argument

jingyugao:
Signed-off-by: jingyugao <1121087373@qq.com>

This change modify go to output err when "append" has only one argument.
Before:
image
After Change:
image

I think append with only one argument has no meaning.I think no one will use it to replace "assignment".
Sometimes we will forget to type the second argument by mistake,but it is still an legal gram.
So I suggest to make it an error.


jingyugao:
I signed it!

PR #33450, opened 2019-08-03
log: modify `Logger` struct, make it supports multi output

kerbalwzy:
Modify the `out` of the `Logger` structure, reset it to `outs`, type `[]io.Writer`, and the relevant places involving the operation `outs`.
The goal is to support multiple outputs of log information, such as simultaneous output to files and consoles, and more.

### DIFF
--- a/src/log/log.go
+++ b/src/log/log.go
@@ -60,8 +60,8 @@ type Logger struct {
 // destinations to which log data will be written.
 // The prefix appears at the beginning of each generated log line.
 // The flag argument defines the logging properties.
-func New(outs []io.Writer, prefix string, flag int) *Logger {
-	return &Logger{outs: outs, prefix: prefix, flag: flag}
+func New(out io.Writer, prefix string, flag int) *Logger {
+	return &Logger{outs: []io.Writer{out}, prefix: prefix, flag: flag}
 }
 
 // SetOutput sets the output destinations for the logger.
@@ -71,7 +71,7 @@ func (l *Logger) SetOutput(w io.Writer) {
 	l.outs = append(l.outs, w)
 }
 
-var std = New([]io.Writer{os.Stderr}, "", LstdFlags)
+var std = New(os.Stderr, "", LstdFlags)
 
 // Cheap integer to fixed-width decimal ASCII. Give a negative width to avoid zero-padding.
 func itoa(buf *[]byte, i int, wid int) {
--- a/src/log/log.go
+++ b/src/log/log.go
@@ -51,26 +51,27 @@ type Logger struct {
 	mu     sync.Mutex // ensures atomic writes; protects the following fields
 	prefix string     // prefix to write at beginning of each line
 	flag   int        // properties
-	out    io.Writer  // destination for output
+	//out    io.Writer  // destination for output
+	outs   []io.Writer// destinations for ouput
 	buf    []byte     // for accumulating text to write
 }
 
-// New creates a new Logger. The out variable sets the
-// destination to which log data will be written.
+// New creates a new Logger. The outs variable sets the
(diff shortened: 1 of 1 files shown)


agnivade:
This PR is incorrect as you are trying to merge it into boringcrypto branch. Please file an issue explaining what you are trying to achieve.

kerbalwzy:
This is my first time to submit RP here. I have just signed CLA.
If I want to contribute some ideas and source code, what should I do or what is the process?
In particular, which branch should I merge with? Thanks.

kerbalwzy:
@googlebot I signed it!

agnivade:
> If I want to contribute some ideas and source code, what should I do or what is the process?
In particular, which branch should I merge with? Thanks.

Welcome to Go :)

If you want to contribute source code, submitting a PR against the master branch is the way to go, as is  usually done in all software projects.

But unfortunately, this is a non-trivial change since you have changed the API of a public function. For all non-trivial changes, please file an issue so that people can discuss and approve it before sending code. Please see - https://golang.org/doc/contribute.html.

For submitting a proposal, please see - https://golang.org/s/proposal

This PR is unlikely to be accepted until the API change has been discussed and approved. I would suggest you to file an issue for that. Thanks.

kerbalwzy:
> Welcome to Go :)
> 
> If you want to contribute source code, submitting a PR against the master branch is the way to go, as is usually done in all software projects.
> 
> But unfortunately, this is a non-trivial change since you have changed the API of a public function. For all non-trivial changes, please file an issue so that people can discuss and approve it before sending code. Please see - https://golang.org/doc/contribute.html.
> 
> For submitting a proposal, please see - https://golang.org/s/proposal
> 
> This PR is unlikely to be accepted until the API change has been discussed and approved. I would suggest you to file an issue for that. Thanks.

Thank you for your advice. I'll do it later.

src/log/log.go
```diff
@@ -51,26 +51,27 @@ type Logger struct {
...
 
-// New creates a new Logger. The out variable sets the
-// destination to which log data will be written.
+// New creates a new Logger. The outs variable sets the
+// destinations to which log data will be written.
 // The prefix appears at the beginning of each generated log line.
 // The flag argument defines the logging properties.
-func New(out io.Writer, prefix string, flag int) *Logger {
-	return &Logger{out: out, prefix: prefix, flag: flag}
+func New(outs []io.Writer, prefix string, flag int) *Logger {
```
kerbalwzy: This make not backwards compatible, May be I need to fix it

src/log/log.go
```diff
@@ -51,23 +51,24 @@ type Logger struct {
...
 	buf    []byte     // for accumulating text to write
 }
 
-// New creates a new Logger. The out variable sets the
-// destination to which log data will be written.
+// New creates a new Logger. The outs variable sets the
+// destinations to which log data will be written.
 // The prefix appears at the beginning of each generated log line.
 // The flag argument defines the logging properties.
 func New(out io.Writer, prefix string, flag int) *Logger {
```
kerbalwzy: Return the original parameter form to make the code backward compatible

src/log/log.go
```diff
@@ -51,23 +51,24 @@ type Logger struct {
...
-	return &Logger{out: out, prefix: prefix, flag: flag}
+	return &Logger{outs: []io.Writer{out}, prefix: prefix, flag: flag}
 }
 
-// SetOutput sets the output destination for the logger.
+// SetOutput sets the output destinations for the logger.
```
leaxoy: Create a new function named `AddOutput`?

kerbalwzy: No, I just forgot to restore the notes.

PR #33517, opened 2019-08-07
net/http: modify variables style

wangzhione:
Minor style change.


src/net/http/server.go
```diff
@@ -2205,9 +2205,7 @@ type muxEntry struct {
 func NewServeMux() *ServeMux { return new(ServeMux) }
 
 // DefaultServeMux is the default ServeMux used by Serve.
-var DefaultServeMux = &defaultServeMux
-
-var defaultServeMux ServeMux
+var DefaultServeMux = NewServeMux()
```
AlexMapley: Might want to check in with https://github.com/bradfitz on this

AlexMapley (approved):
Would double check that `defaultServeMux` isn't referenced anywhere else in the `http` package, but as far as I can tell looks good 👍

PR #33525, opened 2019-08-07
math/big: add big.Int methods Inc() and Dec()

lmittmann:
Adding a sample implementation for big.Int methods `Inc()` and `Dec()`. #29951

Incrementing and decrementing big.Int's by one can be done with less comparisons and allocations opposed to `x.Add(x, one)` and `x.Sub(x, one)`.

Benchmark results:

```
goos: linux
goarch: amd64
pkg: math/big
BenchmarkInc-4         70917309                16.6 ns/op
BenchmarkIncByAdd-4    55380994                18.7 ns/op
```

```
goos: linux
goarch: amd64
pkg: math/big
BenchmarkDec-4         63753597                17.3 ns/op
BenchmarkDecBySub-4    57247021                21.3 ns/op
```

### DIFF
--- a/src/math/big/int.go
+++ b/src/math/big/int.go
@@ -133,6 +133,17 @@ func (z *Int) Add(x, y *Int) *Int {
 	return z
 }
 
+// Inc sets z to the sum z+1 and returns z.
+func (z *Int) Inc() *Int {
+	if z.neg {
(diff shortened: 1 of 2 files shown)


mvdan:
The CLA was never signed, and this needs an issue to propose the API changes first. Closing for now.

PR #33730, opened 2019-08-20
added otp package to generate one time password

shrikar007:
This is the package to generate the one-time password

### DIFF
--- a/src/otp/otp.go
+++ b/src/otp/otp.go
@@ -1,0 +1,14 @@
+package otp
+
+import (
+	"math/rand"
+	"strconv"
+	"strings"
+)
+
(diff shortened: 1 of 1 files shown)


agnivade:
Hello @shrikar007 - thanks for the PR. Unfortunately, I don't think this package needs to exist in the standard library. Please see https://golang.org/doc/faq#x_in_std. 

However, if you think such a function is important enough to be included, we need to discuss this in a proposal and it needs to be approved first. Please file a [proposal](http://golang.org/s/proposal) if you think so. I will close the PR for now. We can reopen this if the proposal gets approved. Thanks.

PR #33732, opened 2019-08-20
Added otp package to generate otp 

shrikar007:
otp package to import inbuilt library to generate one time password


agnivade:
Hi @shrikar007 - Please have a look at my comment [here](https://github.com/golang/go/pull/33730#issuecomment-522919282). I would suggest you to file a proposal issue first and let that be approved. And _then_ open a PR or reopen this one.

PR #33810, opened 2019-08-24
net/http: fix a typo in comments

jrevillas:
HTTP is an initialism, not an acronym, where you pronounce each letter as a
word. It's "an H", not "a H".

Running `find src/net/http -type f | xargs grep -n 'an HTTP' | wc -l` shows
that the "an HTTP" form is used 67 times across the `net/http` package.
Furthermore, `find src/net/http -type f | xargs grep -n 'a HTTP' | wc -l`
yields only 4 results.


DisposaBoy:
@jrevillas Unless I'm missing something... that isn't a typo because `H` isn't a vowel?

jrevillas:
Probably "typo" is not the word. I have updated the description accordingly. Thank you, @DisposaBoy.

PR #33833, opened 2019-08-26
net/http: make docs refer to Context.Value as a getter instead of context.WithValue

extemporalgenome:
The doc comments of both ServerContextKey and LocalAddrContextKey both suggest that context.WithValue can be used to access (get) properties of the server or connection. This PR fixes those comments to refer to Context.Value instead.


extemporalgenome:
@googlebot I signed it!

Issue #29098, opened 2018-12-04
net/url: URL.Parse Multiple Parsing Issues


wir3less:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.11.2 windows/amd64
</pre>

### Does this issue reproduce with the latest release?

Yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\wir3less\AppData\Local\go-build
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=C:\Users\wir3less\go
set GOPROXY=
set GORACE=
set GOROOT=C:\Go
set GOTMPDIR=
set GOTOOLDIR=C:\Go\pkg\tool\windows_amd64
set GCCGO=gccgo
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -mthreads -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=C:\Users\wir3less\AppData\Local\Temp\go-build829182294=/tmp/go-build -gno-record-gcc-switches
</pre></details>

### What did you do?

While playing around with URL.Parse I found a few problems I'd like to share.
I'll gladly share more details if anything is unclear or if someone is interested.

Normally, `javascript:alert(1)` when parsed by `url.parse` has no `Hostname()`
But `javascript://alert(1)` has a hostname of `alert(1)`
This can be taken further...
`javascript://%250aalert(1)+'aa@google.com/a'a` has a hostname of `google.com` and will pop an alert if relocated to by a browser (after decoding)

IPV6 support also has it's issues...
this URL `http://[google.com]:80` has the hostname of `google.com`
But also do all of these:
`http://google.com]:80`
`http://google.com]:80__Anything_you'd_like_sir`
`http://[google.com]FreeTextZoneHere]:80`

Even without thinking about how this would interact with other systems and parsers,
[...]

PR #33876, opened 2019-08-27
net/url: fail TestParseErrors test when getting an unwanted error

### LINKED ISSUE #33646: net/url: mysql://signer@tcp(mysql:3306)/notarysigner no longer parses after 1.12.8

<pre>
$ go version
1.12.8 or 1.11.13

</pre>

Yes

Not relevant

### What did you do?

Updated to Go 1.12.8

CI to pass after updating to Go 1.12.8 😅 https://github.com/theupdateframework/notary/pull/1485

### What did you see instead?

```
error: parse mysql://signer@tcp(mysql:3306)/notarysigner: invalid port ":3306)" after host
```

This looks a similar regression as https://github.com/golang/go/issues/12023

t.b.h., I don't know if such DSN's _should be_ parseable by net/url (https://github.com/golang/go/issues/12023#issuecomment-127976935), but it's a change in behaviour, so I thought I'd report it

Reproducer: https://play.golang.org/p/cPiVWDRZ3G4

Looks like it only fails if there's a port specified:

```go
package main

import (
	"fmt"
	"net/url"
)

func main() {
	_, err := url.Parse("mysql://boulder@tcp(localhost)/boulder_test?parseTime=true")
	if err != nil {
		fmt.Println(err)
	}
	_, err = url.Parse("mysql://boulder@tcp(localhost:3306)/boulder_test?parseTime=true")
	if err != nil {
		fmt.Println(err)
	}
}
```

Produces:

```
parse mysql://boulder@tcp(localhost:3306)/boulder_test?parseTime=true: invalid port ":3306)" after host
```

stefanb:
The TestParseErrors test function was not strict with unwanted errors
received from url.Parse(). It was not failing in such cases, now it does

Fixes #33646
Updates #29098

### DIFF
--- a/src/net/url/url_test.go
+++ b/src/net/url/url_test.go
@@ -1437,7 +1437,7 @@ func TestParseErrors(t *testing.T) {
 		{"http://[%10::1]", true},        // no %xx escapes in IP address
 		{"http://[::1]/%48", false},      // %xx in path is fine
 		{"http://%41:8080/", true},       // not allowed: % encoding only for non-ASCII
-		{"mysql://x@y(z:123)/foo", true}, // not well-formed per RFC 3986, github.com/golang/go/issues/33646
+		{"mysql://x@y(z:123)/foo", true}, // not well-formed per RFC 3986, golang.org/issue/33646
 		{"mysql://x@y(1.2.3.4:123)/foo", true},
 
 		{"http://[]%20%48%54%54%50%2f%31%2e%31%0a%4d%79%48%65%61%64%65%72%3a%20%31%32%33%0a%0a/", true}, // golang.org/issue/11208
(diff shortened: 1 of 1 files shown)


stefanb:
@av86743 no surprise there, this change fixes the TEST to detect the bug #33646 and rightfully fail.

stefanb:
Because this needs to get in as soon as the  #33646 is fixed. In the same
https://go-review.googlesource.com/c/go/+/185080 I was kindly asked by maintainers to re-submit it for a later merge.

stefanb:
> So, basically, you don't care that `.Errorf()` message contains `"...; want no error"`.

@av86743 That is a testing construct to fail a test if the tested `Parse` function returned an error that was not wanted.

Issue #27128, opened 2018-08-21
net/http: FileServer on WSL fails with wget and curl for files over ~4MB


upsampled:
Please answer these questions before submitting your issue. Thanks!

### What version of Go are you using (`go version`)?
`go version go1.10.3 linux/amd64`

### Does this issue reproduce with the latest release?
yes

### What operating system and processor architecture are you using (`go env`)?
GOARCH="amd64"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build493773858=/tmp/go-build -gno-record-gcc-switches"

using WSL

Ran the same test on a CentOS Machine and I did not see this problem

### What did you do?
Attempted to server files via HTTP that an application using curl can download.

```
package main

import (
        "log"
        "net/http"
)

func main() {
        fs := http.FileServer(http.Dir("./"))
        http.Handle("/repo/", http.StripPrefix("/repo/", fs))

        log.Println("Listening...")
        http.ListenAndServe(":80", nil)
}
```

### What did you expect to see?
Curl to download the files.  

### What did you see instead?
The file server worked for file that were roughly less than 5M, but failed for larger files with `curl error 18`. The curl command line tool also reproduces this error:
```
curl http://<IP>/repo/<FILE> --output <FILE>
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                                    Dload  Upload   Total   Spent     Left  Speed
 56 7257k   56 4096k    0     0      4970k      0       0:00:01 --:--:--  0:00:01 4970k
curl: (18) transfer closed with 3237034 bytes remaining to read
```

Wget (busybox version) also has the same problem:
```
wget http://<IP>/repo/<FILE> -O <FILE>
Connecting to <IP> (<IP>:80)
[...]

PR #33882, opened 2019-08-27
net/http: improve FileServer error logging

hullarb:
The errors returned during copying the file content in ServeContent,
ServeFile and FileServer were ignored to avoid excessive, meaningless
logging. This commit introduces error filtering and logging to ensure
that the errors occurring during reading the content are logged.

Updates #27128


hullarb:
@googlebot I signed it!

PR #33930, opened 2019-08-29
database/sql: refactor err returning value from Scan to if stmt

arthurkushman:
database/sql: refactor err returning value from Scan to if stmt

This change modifies Go to satisfy it's own code-style rules and readability + compiler parsing boosting


arthurkushman:
@googlebot I fixed it..

PR #34127, opened 2019-09-05
encoding/json: fix and optimize marshal for quoted string

### LINKED ISSUE #34154: encoding/json: string option (struct tag) on string field with SetEscapeHTML(false) escapes anyway
If using `encoding/json.NewEncoder` with `SetEscapeHTML(false)` an a field, that has the `string` option in the struct tags set, the HTML content gets escaped anyways.

<pre>
$ go version
go version go1.12.5 linux/amd64
</pre>

Yes

$ go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/lubr/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/lubr/go"
GOPROXY=""
GORACE=""
GOROOT="/home/lubr/.gvm/versions/go1.12.5.linux.amd64"
GOTMPDIR=""
GOTOOLDIR="/home/lubr/.gvm/versions/go1.12.5.linux.amd64/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build368255501=/tmp/go-build -gno-record-gcc-switches"
</pre></details>

### What did you do?

```Go
package main

import (
	"encoding/json"
	"os"
)

type Foo struct {
	Bar string `json:"bar,string"`
}

func main() {
	j := Foo{
		Bar: `<html>foobar</html>`,
	}
	e := json.NewEncoder(os.Stdout)
	e.SetEscapeHTML(false)
	_ = e.Encode(j)
}
```

```plain
$  go run main.go
{"bar":"\"<html>foobar</html>\""}
```

### What did you see instead?

```plain
$  go run main.go
{"bar":"\"\\u003chtml\\u003efoobar\\u003c/html\\u003e\""}
```

#34127 does fix this issue

breml:
Since Go 1.2 every string can be marshaled to JSON without error even if it
contains invalid UTF-8 byte sequences. Therefore there is no need to use
Marshal again for the only reason of enclosing the string in double quotes.
Not using Marshal here also removes the error check as there has not been a
way for Marshal to fail anyway.

name          old time/op    new time/op    delta
Issue34127-4     360ns ± 3%     200ns ± 3%  -44.56%  (p=0.008 n=5+5)

name          old alloc/op   new alloc/op   delta
Issue34127-4     56.0B ± 0%     40.0B ± 0%  -28.57%  (p=0.008 n=5+5)

name          old allocs/op  new allocs/op  delta
Issue34127-4      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.008 n=5+5)

Fixes #34154

### DIFF
--- a/src/encoding/json/bench_test.go
+++ b/src/encoding/json/bench_test.go
@@ -297,6 +297,22 @@ func BenchmarkIssue10335(b *testing.B) {
 	})
 }
 
+func BenchmarkIssue34127(b *testing.B) {
+	b.ReportAllocs()
+	j := struct {
+		Bar string `json:"bar,string"`
+	}{
+		Bar: `foobar`,
+	}
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			if _, err := Marshal(&j); err != nil {
+				b.Fatal(err)
+			}
+		}
+	})
+}
+
 func BenchmarkUnmapped(b *testing.B) {
 	b.ReportAllocs()
 	j := []byte(`{"s": "hello", "y": 2, "o": {"x": 0}, "a": [1, 99, {"x": 1}]}`)
--- a/src/encoding/json/encode.go
+++ b/src/encoding/json/encode.go
@@ -601,11 +601,11 @@ func stringEncoder(e *encodeState, v reflect.Value, opts encOpts) {
 		return
 	}
 	if opts.quoted {
-		sb, err := Marshal(v.String())
-		if err != nil {
-			e.error(err)
-		}
-		e.string(string(sb), opts.escapeHTML)
+		b := make([]byte, 0, v.Len()+2)
+		b = append(b, '"')
+		b = append(b, []byte(v.String())...)
+		b = append(b, '"')
+		e.stringBytes(b, opts.escapeHTML)
 	} else {
 		e.string(v.String(), opts.escapeHTML)
 	}
--- a/src/encoding/json/stream_test.go
+++ b/src/encoding/json/stream_test.go
@@ -118,6 +118,11 @@ func TestEncoderSetEscapeHTML(t *testing.T) {
 		Ptr    strPtrMarshaler
 	}{`"<str>"`, `"<str>"`}
 
+	// https://golang.org/issue/34154
+	stringOption := struct {
+		Bar string `json:"bar,string"`
+	}{`<html>foobar</html>`}
+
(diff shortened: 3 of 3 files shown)


breml:
@mvdan Thanks for your hint about the benchmarks. From my point of view, the change was not primary driven by performance optimization in mind, this came more as a welcome side effect.

My motivation for the proposed change is:

1. remove the unnecessary error case
2. fix #34154

Nevertheless I will try to have a look at `benchstat` and provide some benchmark statistics for the improved performance as well.

breml:
@mvdan I had a look at the benchmarks in `encoding/json` and non of them contains struct fields with the string option set in the struct tags.

The change affects the encoding of struct like:

```Go
type Foo struct {
	Bar string `json:"bar,string"`
}
```

While looking through the commit messages for the recent changes on `encoding/json` I found this [one](https://github.com/golang/go/commit/ae68a912725e5a3a0482bc5945687663f2ddafe3#diff-cbb71488de8e250b59cee0de7af128f6) from you:

> ...
> This change should slightly reduce the amount of work when decoding into
> json.Number, though that isn't very common nor part of any current
> benchmarks.
> ...

I would argue, that this change also slightly improves encoding performance and that it is not very common nor part of any existing benchmarks. Would this be fine as well or do you insist on benchmarks for this particular case?

mvdan:
Please reply on Gerrit. I'd rather not keep the discussion in two places.

andig:
For some reason I cannot comment on the CL. I have a stupid suggestion that might be worth trying. In https://go-review.googlesource.com/c/go/+/194338#message-ce99b8cb629480e1273232d92051ffd7b815922a an encodeBuffer from Pool is used to optimize allocations. Being in the same package it might also lend itself here instead of the bytes array. However, it contains overhead and cannot be allocated to the desired size. May not be worth it in the end or not even the cleanest thing to do.

PR #34319, opened 2019-09-16
net/http: fix reading cookie mistake if ";;" exists, most browsers like Chrome, Safari can handle such a cookie 

qyb225:
If cookies is something like this: "a=xxx1;; b=xxx2", func readCookies just ignore "b=xxx2" because of this mistake. This PR can fix it.

### DIFF
--- a/src/net/http/cookie.go
+++ b/src/net/http/cookie.go
@@ -246,7 +246,7 @@ func readCookies(h Header, filter string) []*Cookie {
 
 		var part string
 		for len(line) > 0 { // continue since we have rest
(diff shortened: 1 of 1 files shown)


qyb225:
In the source code, I find that readCookies funciton doesn't return any error, it handles some error cookie as excepted. 

For example, if cookie is `"a=1; ;b=2"` (noticed that there is a space between two ';'), readCookies can recognize both a=1 and b=2 as expected. But if cookie is `"a=1;;b=2"`, readCookies just ignored b=2. It's confusing.

PR #34386, opened 2019-09-19
net/url: add tohex

zhangyunhao116:
Use tohex instead of string like "0123456789ABCDEF", the former is more faster and readable.


ZYunH:
@googlebot I signed it!

PR #34479, opened 2019-09-23
testing: WIP: Adding mean and CI to benchmarks with high repeat count

voronaam:
This PR adds mean and 95 confidence interval to benchmark output with high enough run count is specified.

The idea is that when a user supplied high `testing.count` parameter, they want to look at the statistics.

Please let me know if there is interest in accepting this PR in. I will work with the reviewers to make this more parameterized. Otherwise I will keep this patch as a useful tool for myself.

Sample output:
```
$ go test -bench=. -count=20
goos: darwin
goarch: amd64
BenchmarkAtomic-12      99183856                11.9 ns/op
BenchmarkAtomic-12      99689896                12.4 ns/op
BenchmarkAtomic-12      100000000               11.9 ns/op
BenchmarkAtomic-12      98967619                12.1 ns/op
BenchmarkAtomic-12      99485732                12.2 ns/op
BenchmarkAtomic-12      96931659                12.2 ns/op
BenchmarkAtomic-12      100000000               12.1 ns/op
BenchmarkAtomic-12      96437036                12.4 ns/op
BenchmarkAtomic-12      96977478                12.3 ns/op
BenchmarkAtomic-12      97329852                12.1 ns/op
BenchmarkAtomic-12      99378790                12.1 ns/op
BenchmarkAtomic-12      98713808                12.0 ns/op
BenchmarkAtomic-12      100000000               12.0 ns/op
BenchmarkAtomic-12      100000000               12.1 ns/op
BenchmarkAtomic-12      95132949                12.0 ns/op
BenchmarkAtomic-12      100000000               12.3 ns/op
BenchmarkAtomic-12      99645378                12.1 ns/op
BenchmarkAtomic-12      100000000               12.5 ns/op
BenchmarkAtomic-12      100000000               13.0 ns/op
BenchmarkAtomic-12      100000000               11.9 ns/op
--- Overall: BenchmarkAtomic-12         12.2±0.12 ns/op
```

### DIFF
--- a/src/runtime/export_test.go
+++ b/src/runtime/export_test.go
@@ -734,6 +734,16 @@ func (p *PageAlloc) Scavenge(nbytes uintptr, locked bool) (r uintptr) {
 	})
 	return
 }
+func (p *PageAlloc) InUse() []AddrRange {
+	ranges := make([]AddrRange, 0, len(p.inUse.ranges))
+	for _, r := range p.inUse.ranges {
+		ranges = append(ranges, AddrRange{
+			Base:  r.base,
+			Limit: r.limit,
+		})
+	}
+	return ranges
+}
 
 // Returns nil if the PallocData's L2 is missing.
 func (p *PageAlloc) PallocData(i ChunkIdx) *PallocData {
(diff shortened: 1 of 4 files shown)


rsc:
go test should not be in the business of statistical analysis.
There are so many analyses you might want to do on the numbers.
Benchstat is one set; keeping them outside go test means you can write your own just as easily.

If the problem is that go test does not print enough significant figures for you to write your own analysis tool, let's talk about that instead.

It sounds like you want to see at least four significant digits instead of three,
so that instead of 12.3 ns/op we'd print, say, 12.35 ns/op. I'm skeptical that the 0.05 ns/op has any accuracy behind it, but I also wouldn't object to adding a digit for these small numbers.

/cc @aclements

rsc:
Sorry, meant to reply on #34626. Please comment there instead.

PR #34551, opened 2019-09-26
syscall/js: improve documentation of js.FuncOf

### LINKED ISSUE #34324: syscall/js: unclear behavior of js-triggered and wrapped functions

<pre>
go version go1.13 darwin/amd64
</pre>

yes

go env
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/tschinke/Library/Caches/go-build"
GOENV="/Users/tschinke/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/tschinke/go"
GOPRIVATE=""
GOPROXY="direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/tschinke/repos/wdy_ausbildung/trainiety/app/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/bv/6rt5ck194ls704dz9p4c0hlh0000gn/T/go-build951955169=/tmp/go-build -gno-record-gcc-switches -fno-common"

</pre></details>

### What did you do?
```
<div id="outer" class="fullscreen">
  <div id="inner" class="dialog">
  </div>
</div>

var outerDiv js.Value = ...
var innerDiv js.Value = ...

outerDiv.Call("addEventListener", "click", js.FuncOf(func(this js.Value, args []js.Value) interface{} {
		log.Println("listener called: ", this, this.Get("id"))
		return nil
	}), once)
[...]
innerDiv.Call("addEventListener", "click", js.FuncOf(func(this js.Value, args []js.Value) interface{} {
                 args[0].Call("stopPropagation")
		log.Println("listener called: ", this, this.Get("id"))
		return nil
	}), once)
```

Naturally I expected that the listener to the inner div should always be called first. However looking at  the documentation of js.FuncOf, I don't know what to expect:

> Invoking the JavaScript function will synchronously call the Go function fn with the value of JavaScript's "this" keyword and the arguments of the invocation.
> [...]
> A wrapped function triggered by JavaScript's event loop gets executed on an extra goroutine.
> Blocking operations in the wrapped function will block the event loop.
>[...]
>A blocking function should therefore explicitly start a new goroutine.

The logic of these statements is contradictory: 
* What exactly is meant by a  "synchronous call" in the first statement?
* If a wrapped and triggered function is always executed in a goroutine, why should it block the event loop?

### What did you see instead?
I tried to use `stopPropagation` but the calling order is always wrong (outer-div first), independent of registration order. Also if a callback is always executed within a new goroutine, it sounds like `stopPropagation` et. al cannot be used properly. 

The documentation of js.FuncOf does not help me much to understand the actual calling behavior or at least if this works as "expected".

torbenschinke:
The existing documentation is improved to be more
explicit about the lifecycle and its consequences.

Fixes #34324

### DIFF
--- a/src/syscall/js/func.go
+++ b/src/syscall/js/func.go
@@ -22,19 +22,24 @@ type Func struct {
 	id    uint32
 }
 
-// FuncOf returns a wrapped function.
-//
-// Invoking the JavaScript function will synchronously call the Go function fn with the value of JavaScript's
-// "this" keyword and the arguments of the invocation.
-// The return value of the invocation is the result of the Go function mapped back to JavaScript according to ValueOf.
-//
-// A wrapped function triggered during a call from Go to JavaScript gets executed on the same goroutine.
-// A wrapped function triggered by JavaScript's event loop gets executed on an extra goroutine.
-// Blocking operations in the wrapped function will block the event loop.
-// As a consequence, if one wrapped function blocks, other wrapped funcs will not be processed.
-// A blocking function should therefore explicitly start a new goroutine.
+// FuncOf registers a function to be used by JavaScript.
 //
 // Func.Release must be called to free up resources when the function will not be used any more.
+//
+// The Go function fn is called with the value of JavaScript's "this" keyword and the
+// arguments of the invocation. The return value of the invocation is
+// the result of the Go function mapped back to JavaScript according to ValueOf.
+// 
+// Invoking the wrapped Go function from JavaScript will
+// pause the event loop and spawns a new go routine.
+// Other wrapped functions which are triggered during a call from Go to JavaScript
+// gets executed on the same goroutine.
+//
+// As a consequence, if one wrapped function blocks, JavaScript's event loop
+// is blocked until that func returns.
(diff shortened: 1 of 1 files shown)


torbenschinke:
I haven't heard from you for while now. I updated everything to the reviewers favor. Is there something else I need to do?

PR #34667, opened 2019-10-02
sort: be more specific about string sorting order in docs

### LINKED ISSUE #34666: sort: Strings sort specifies increasing order (not lexicographic)
I believe 'increasing order' can be too ambiguous for official documentation. Since that could be interpreted as increasing in lengths of strings instead.
`Strings sorts a slice of strings in increasing order. `

I propose for it to state `...in lexicographical order` instead.  (e.g. [Oracle Java docs do that](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#compareTo(java.lang.String)))

JurisMajors:
Simple changes in documentation regarding sorting strings. Instead of "increasing order" it's "lexicographic order"
Fixes #34666

### DIFF
--- a/src/sort/sort.go
+++ b/src/sort/sort.go
@@ -291,7 +291,7 @@ func isNaN(f float64) bool {
 // Sort is a convenience method.
 func (p Float64Slice) Sort() { Sort(p) }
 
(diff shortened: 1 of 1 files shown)


JurisMajors:
@googlebot I signed it!

PR #34704, opened 2019-10-04
bufio: simplify bufio.Reader.ReadBytes to avoid an extra loop over a slice

alex:


av86743:
@ianlancetaylor

Notice how identical [`len(frag)`](https://github.com/golang/go/blob/5f4aa5d79fba270d6a14f2a55999b5b6ccec2326/src/bufio/bufio.go#L448) and [`len(buf)`](https://github.com/golang/go/blob/5f4aa5d79fba270d6a14f2a55999b5b6ccec2326/src/bufio/bufio.go#L451) are not CSE-d. This is not trivial for compiler to do as it requires knowledge of properties of `copy()`.

PR #34741, opened 2019-10-07
testing/iotest: add ErrReader

caarlos0:
Adds an io.Reader that always returns 0 and a non-nil error.

Fixes #38781

### DIFF
--- a/src/testing/iotest/reader.go
+++ b/src/testing/iotest/reader.go
@@ -68,6 +68,7 @@ func (r *dataErrReader) Read(p []byte) (n int, err error) {
 	return
 }
 
+// ErrTimeout is a fake timeout error.
 var ErrTimeout = errors.New("timeout")
 
 // TimeoutReader returns ErrTimeout on the second read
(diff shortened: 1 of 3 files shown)


caarlos0:
It's not clear if I need to do anything for this to be reviewed 🤔

src/testing/iotest/reader.go
```diff
@@ -68,6 +68,7 @@ func (r *dataErrReader) Read(p []byte) (n int, err error) {
 	return
 }
 
+// ErrTimeout is a fake timeout error
```
carnott-snap, line 71: doc comments should end with `.`s:
```suggestion
// ErrTimeout is a fake timeout error.
```

src/testing/iotest/reader.go
```diff
@@ -86,3 +87,17 @@ func (r *timeoutReader) Read(p []byte) (int, error) {
 	}
 	return r.r.Read(p)
 }
+
+// ErrIO is a fake IO error
```
carnott-snap, line 91: doc comments should end with `.`s:
```suggestion
// ErrIO is a fake IO error.
```

PR #34761, opened 2019-10-08
Add a cancel function which can cancel a context due to an error

fanfuxiaoran:
Sometimes, we run mutilple go routines in a context, if one go routine
failed due to an error, we want to cancel the context and record the
error in it. So I added a function called WithErrorCancel function which
returns both CancelFunc and ErrorCancelFunc. CancelFunc is the same as
the function which is returned by WithCancel, ErrorCancelFunc is used to
cancel a context with an error.

### DIFF
--- a/src/context/context.go
+++ b/src/context/context.go
@@ -552,3 +552,12 @@ func (c *valueCtx) Value(key interface{}) interface{} {
 	}
 	return c.Context.Value(key)
 }
+
+// A ErrorCancelFunc tells an opertion to abandon due to the error.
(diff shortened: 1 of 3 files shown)


fanfuxiaoran:
@googlebot I signed it!

fanfuxiaoran:
Have created a issue for this pr:
https://github.com/golang/go/issues/34764

fanfuxiaoran:
#34764

bradfitz:
Closing per comments on https://go-review.googlesource.com/c/go/+/199505

PR #34803, opened 2019-10-09
strconv/quote: refactor - moved quotes to constants

proggga:
in strconv/quote.go moved to constants '\'" and '"'
small refactoring


proggga:
@googlebot I signed it!

proggga:
Okey, so what you think If i change to private const? I made public by mistake

PR #34843, opened 2019-10-11
debug/pe: limit the maximum size of the string table buffer to 1MB

niallnsec:
Fixes an issue where Go may panic if an otherwise valid PE file has an invalid or corrupt string table due to allocation of a data controlled buffer bound to the maximum size of a uint32.


niallnsec:
@googlebot I signed it!

niallnsec:
@googlebot I fixed it.

Issue #17036, opened 2016-09-09
net/smtp: possible typo in Client.Rcpt


dtelyukh:
https://github.com/golang/go/blob/master/src/net/smtp/smtp.go#L256
`c.cmd(25, "RCPT TO:<%s>", to)`

I guess it must be 250, but not 25. Am I right?

bradfitz:
Yes, looks like it.

Why did nobody (or no tests) notice this before?

Does Rcpt return an error now, but nobody checks the error?

mattn:
Maybe, no one check the error?

https://github.com/search?utf8=%E2%9C%93&q=%22client.Rcpt%22+language%3AGo&type=Code&ref=searchresults

bradfitz:
Paging Mr. @campoy for analysis! :)

mattn:
https://github.com/golang/go/blob/0104a31b8fbcbe52728a08867b26415d282c35d2/src/net/textproto/reader.go#L196-L200

https://play.golang.org/p/YTS3qTb_PE

odeke-em:
Hmm, as alluded by @mattn, the signature of Client.cmd https://github.com/golang/go/blob/master/src/net/smtp/smtp.go#L103
```go
func (c *Client) cmd(expectCode int, format string, args ...interface{}) (int, string, error) {
```

takes in an expectCode which is used in verifying the status of normalized message.

Perhaps we could improve the readability of that code by defining a table/small function of expect codes and then changing that to something like
```diff
+ 	_, _, err := c.cmd(expectCode(250), "RCPT TO:<%s>", to)
-	_, _, err := c.cmd(25, "RCPT TO:<%s>", to)
```

or set it to 250
```diff
+ 	_, _, err := c.cmd(250, "RCPT TO:<%s>", to)
-	_, _, err := c.cmd(25, "RCPT TO:<%s>", to)
```

 following suit with other usages:
https://github.com/golang/go/blob/master/src/net/smtp/smtp.go#L155
https://github.com/golang/go/blob/master/src/net/smtp/smtp.go#L184
https://github.com/golang/go/blob/master/src/net/smtp/smtp.go#L220
https://github.com/golang/go/blob/master/src/net/smtp/smtp.go#L248

mattn:
I guess the code is trying to handle codes like 221/251. But seems not good.

odeke-em:
So I was curious about the acceptable status codes for RCPT and looked at a few resources on the 25X code intentions so found some hits off Google:


[...]

PR #34903, opened 2019-10-14
net/smtp: change expected command for RCPT to 250

xhit:
According to RFC5321 https://tools.ietf.org/html/rfc5321 the RCPT command returns 250 when result is OK


xhit:
@googlebot I signed it!

xhit:
Closed by conversation about expectCode in https://github.com/golang/go/issues/17036

expectCode works in a range, if expectCode is 25 then a valid code is in range 250-259

Issue #12073, opened 2015-08-07
text/tabwriter: add cell width function


jszwedko:
### What version of Go are you using (go version)?

1.4.2
### What operating system and processor architecture are you using?

Linux / AMD64
### What did you do?

``` go
package main

import (
        "fmt"
        "os"
        "text/tabwriter"
)

func main() {
        t := tabwriter.NewWriter(os.Stdout, 0, 8, 0, '\t', 0)
        fmt.Fprintf(t, "\x1b[31mfoo\x1b[0m\tbar\n")
        fmt.Fprintf(t, "foo\tbar\n")
        t.Flush()
}
```
### What did you expect to see?

```
foo     bar
foo     bar
```

Where the first foo is red.
### What did you see instead?

```
foo     bar
foo             bar
```

Where the first foo is red due to non-printable ANSI escape sequences being included in cell width calculations.
### Proposal:

I realize that printing to the terminal may not be `text/tabwriter`s intended purpose, but I think it would be nice to be able to configure how the width of a cell is calculated as I can imagine further cases (similar to the special casing of HTML tags and entities currently) that would also benefit from this.

Suggestion: Exposing an additional field on the `tabwriter.Writer` struct so as to maintain the function signature of `Init` and `NewWriter`:

``` go
type Writer struct {
    CellWidth func([]byte) int
}
```

If this function was `nil`, the existing width calculation could be used, but otherwise this function could be called with the contents of the cell.

I can take a stab at the implementation of this, but I first wanted to see if such a change would be welcome.

ianlancetaylor:
CC @griesemer

griesemer:
@jszwedko I think this is a reasonable suggestion, and in retrospect perhaps this would have been a better way to customize cell measurement than with the specific flags we have now. That said, I'm not sure how easy it would be to make this work given that part of the current implementation is using a state-machine like mechanism (but I haven't looked into it in detail).

Specifically, my concern is how such a function would interact or possibly interfere with what we have already.

Alternatively, an additional flag could be provided that would exclude escaped sequences of characters from the width computation. Or perhaps another escape character. Both these might be more in the current spirit of things even though perhaps less flexible than the function.

If you want to give it a shot and come up with a concrete implementation, please feel free to go ahead. I'm happy to review but will push back if the result doesn't fit nicely with what we have.

jszwedko:
[...]

PR #35017, opened 2019-10-20
text/tabwriter: add ANSI Graphics Rendition format

kindermoumoute:
Updates #12073


kindermoumoute:
@googlebot I signed it!

PR #35038, opened 2019-10-21
Add Getline in fmt

0delta0:
Add Getline in fmt :)


mvdan:
The CLA wasn't signed in a month. Also, all API changes need to go through a proposal first.

Issue #14808, opened 2016-03-13
cmd/compile: unnecessary bounds checks are not removed


brtzsnr:
Please answer these questions before submitting your issue. Thanks!

1. What version of Go are you using (`go version`)?
go version devel +8ec8017 Sun Mar 13 22:12:01 2016 +0000 linux/amd64

2. What operating system and processor architecture are you using (`go env`)?
GOARCH="amd64"
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"

3. What did you do?
4. What did you expect to see?
5. What did you see instead?
I composed a list of [bound checks](https://docs.google.com/document/d/1vdAEAjYdzjnPA9WDOQ1e4e05cYVMpqSxJYZT33Cqw2g/edit#) a that the compiler at tip can and cannot eliminate. For some I provided an alternative implementation.

btracey:
Is this a duplicate of #5364 ? Both seem like tracking issues.

martisch:
A case that maybe could be added to the doc above and loopbce.go is when writing to an array/slice from back to front e.g. for integer formatting:

```
for i = len(buf) - 1; i >= 0; i-- {
    if u < 10 {
        buf[i] = byte('0' + u)
        break
    }
    next := u / 10
    buf[i] = byte('0' + u - next*10)
    u = next
}
```
This currently generates bounds checks but when writing the above with i = 0; i < len(buf); i++ the bounds checks are removed.

seebs:
https://play.golang.org/p/epAcCz9W-Uu
=>
https://godbolt.org/z/vl_rTS

If you have four slice indexing operations in a single expression, with constant indexes, it should be possible to combine them.

PR #35040, opened 2019-10-21
encoding/binary: removal of unnecessary boundary checks.

ghost:
When creating a slice, e.g.: a 4 byte array can become optimized to 3 bytes only, resulting in a panic from the boundary check, which was experienced during decoding of LoRa packets.

Fixes issue #14808.


lemmerelassal:
> @googlebot I signed it!

PR #35077, opened 2019-10-22
crypto/x509: applied go fmt to verify.go

evalevanto:


evalevanto:
@googlebot I fixed it.

PR #35108, opened 2019-10-23
gofmt/utils: and fix/utils: moved same functions to common file like src/cmd/gofmt/internal.go

proggga:
Moved function to private utils.go, that is copypaste
We should improve this somehow, to use it normally without copying,

for example, we have 6 or 8 copies of isGoFIle and they are all the same but with small differences


proggga:
Close it, i know the better solution

PR #35110, opened 2019-10-23
crypto/elliptic: implement unmarshal for compressed format

### LINKED ISSUE #34105: crypto/elliptic: handle compressed formats in new MarshalCompressed, UnmarshalCompressed

<pre>
$ go version
go version go1.12.9 linux/amd64
</pre>

I think so.

play.golang.org

### What did you do?
https://play.golang.org/p/ugkQeobnYRP

Successful unmarshaling.

### What did you see instead?
Failed unmarshaling.

### The real problem
`elliptic.Unmarshal()` only supports uncompressed format (prefix `0x04`) at the moment, as can be seen [in the source code](https://golang.org/src/crypto/elliptic/elliptic.go?s=8747:8803#L312), which probably is fine given that it's supposed to be a companion for `elliptic.Marshal()` that only uses that format to serialize the point. But at the same time it would be very nice to have support for compressed (prefixes `0x02` and `0x03`) format that is being used in the wild, it doesn't cost much in terms of code (there is a [nice snippet on SO](https://stackoverflow.com/questions/46283760)) and this proposed change is certainly backwards-compatible. At the moment I see this (or similar) code being duplicated by several projects that use compressed format, thus this issue.

Of course it would be very nice if `elliptic` could also serialize into this format, but that's a bit more complicated, because you can't change `Marshal()` without breaking backwards compatibility (the only option I see at the moment is introducing another function like `MarshalCompressed()`).

im-kulikov:
First byte can be 0x2 or 0x3 (compressed form).
To unmarshal used formula y² = x³ - 3x + b.

Fixes golang/go#34105

### DIFF
--- a/src/crypto/elliptic/elliptic.go
+++ b/src/crypto/elliptic/elliptic.go
@@ -52,11 +52,8 @@ func (curve *CurveParams) Params() *CurveParams {
 	return curve
 }
 
-func (curve *CurveParams) IsOnCurve(x, y *big.Int) bool {
-	// y² = x³ - 3x + b
-	y2 := new(big.Int).Mul(y, y)
-	y2.Mod(y2, curve.P)
-
+// polynomial returns x³ - 3x + b
+func (curve *CurveParams) polynomial(x *big.Int) *big.Int {
 	x3 := new(big.Int).Mul(x, x)
 	x3.Mul(x3, x)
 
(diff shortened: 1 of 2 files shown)


im-kulikov:
@dvyukov I believe this is quite useful change, could you please suggest what should be changed in order to merge this request? Thanks.

dvyukov:
> @dvyukov I believe this is quite useful change, could you please suggest what should be changed in order to merge this request? Thanks.

This should answer your questions:
https://go-review.googlesource.com/c/go/+/202819/3#message-f07276cc31f521e20bb5c68f910b8de8c4de4580

im-kulikov:
> During May-July and Nov-Jan the Go project is in a code freeze, during which
little code gets reviewed or merged. 

@dvyukov oh, it's my mistake

im-kulikov:
Any updates?

im-kulikov:
As I see, failed in another part (go mod)

im-kulikov:
any updates?

FiloSottile:
Closing the PR hoping gerritbot will then let me make changes to it in Gerrit, so I can land it today.

PR #35123, opened 2019-10-23
doc/play: match the hello world snippet in x/website

### LINKED ISSUE #35099: x/website: golang.org homepage has two different Hello worlds, one is not gofmt-ed
The "Try Go" element on the golang.org homepage has two different _Hello world_, one of which is not gofmt-ed. 

When you load golang.org for the first time, the "Try Go" element on the right look like this:

hello1

The "Open in Playground" element links to https://play.golang.org/p/MbPUiI4y-IE, which contains a non-gofmt-ed Hello world (it uses 2 spaces to indent the `fmt` line).

If you use the selector below to display another code snippet (say, _Concurrent Pi_), and then you re-select _Hello world_, the Hello that is displayed looks like this:

hello2

Note how this is a _different_ Hello world (gofmt-ed, and the comment at the top is missing). The "Open in Playground" element links to a different program (https://play.golang.org/p/7vin2BK8_A6).

We should probably settle on one Hello world (gofmt-ed). 

If we want to keep two (one with the "// You can edit this code!" comment at the top, and one without), we should at least gofmt the default one. It looks bad to have a non-gofmt-ed program on the homepage.

rhnvrm:
Fixes #35099

### DIFF
--- a/doc/play/hello.go
+++ b/doc/play/hello.go
@@ -1,3 +1,5 @@
+// You can edit this code!
+// Click here and start typing.
 package main
 
 import "fmt"


rhnvrm:
cc @dmitshur

PR #35168, opened 2019-10-25
crypto/rsa: correct the RIPEMD160 hash prefix per RFC 4880

ragansa:
https://tools.ietf.org/html/rfc4880#section-5.2.2
I validated the fix using 3rd party software (openssl) to ensure interoperability; I was unable to find an algorithm validation test


ragansa:
Yeah, this is difficult... I'm not finding a good source.  My assumption is that it's always been incorrect.
I've noticed in the community in general that the "go" version is more or less limited to golang projects:
https://github.com/search?q=%220x30%2C+0x20%2C+0x30%2C+0x08%2C+0x06%2C+0x06%2C+0x28%2C+0xcf%2C+0x06%2C+0x03%2C+0x00%2C+0x31%2C+0x04%2C+0x14%22&type=Code

And the "corrected" version is more widespread:
https://github.com/search?q=%220x30%2C+0x21%2C+0x30%2C+0x09%2C+0x06%2C+0x05%2C+0x2B%2C+0x24%2C+0x03%2C+0x02%2C+0x01%2C+0x05%2C+0x00%2C+0x04%2C+0x14%22&type=Code

I realize this is hardly enough to advocate for changing :)
I'll keep looking for something more definitive.

PR #35322, opened 2019-11-03
cmd/go/internal/get: automatically remove https:// prefixes

ruudk:
I often want to `go get` a Go package that I'm browsing on Github.

Nowadays all browsers hide https:// prefixes from the address bar.

After I manually select github.com/user/repo and copy it, I go the terminal and type `go get <paste>` and press enter. The browser automatically prefixed the text I copied with https://.

The go get program immediately gives an error:
```
go get https:/github.com/user/repo: malformed module path "https:/github.com/user/repo": invalid char ':'
```
With this change, we improve the UX, by quickly removing the https:// prefix and just continue.

I need to add a test to verify that this is working. What's the best practice here?


bradfitz:
Sorry, perhaps you can send a change to return a more helpful error message instead?

PR #35357, opened 2019-11-04
cmd/go/internal/modfetch: add GOINSECURE

witchard:
Enables insecure fetching of dependencies whos path matches those specified in
the enironment variable GOINSECURE.

Fixes #32966


witchard:
I think I have signed the CLA now.

Also, I was very confused about how the script tests work to actually test this - are there some pointers somewhere on how I would run a script test with a custom webserver (that isn't secure!)?

witchard:
@googlebot I signed it!

PR #35405, opened 2019-11-06
syscall: fix the wrong comment

tanin47:
The length is less than or equal to 3, not less than 3.


tanin47:
@googlebot I signed it!

tanin47:
Wait, I'm wrong. Sorry for the false alarm.

PR #35540, opened 2019-11-12
crypto/rand: detect biases in rand.Int(), clarify internals

paragonie-scott:
cc @FiloSottile


paragonie-scott:
A more comprehensive test would be to fill up an array of size N with M samples (where M >= X^2 * N) and ensure that every element in the array has at least X samples.

So long as X isn't very small, you should always see a roughly even distribution, so if something ever violates this condition, you have a potential bias. If it's reproducible, you definitely have a bias.

PR #35659, opened 2019-11-18
bytes: except "slice bounds out of range" error

kuwa72:
Currently bytes.Buffer do not check error before use read length.
If io.Reader return invalid read length with error, raise "slice bounds out of range".


kuwa72:
@googlebot I signed it!

PR #35681, opened 2019-11-19
net/http: add additional doc regarding http request context cancel behaviour

novalagung:
for incoming request which have nil body, the request context will be cancelled when client's connection closes.

However for incoming request which have non-nil body, the request context will **never** be cancelled unless the application starts reading the body.

I think it would be good if we can have this non-nil request context cancel behaviour documented.

Related discussion on StackOverflow https://stackoverflow.com/a/57247158/1467988


novalagung:
@googlebot I signed it!

PR #35736, opened 2019-11-21
net/url: update net/url split comment

### LINKED ISSUE #35735: net/url: comment of split function is wrong
```
// split slices s into two substrings separated by the first occurrence of
// sep. If cutc is true then sep is excluded from the second substring.
// If sep does not occur in s then s and the empty string is returned.
func split(s string, sep byte, cutc bool) (string, string) {
	i := strings.IndexByte(s, sep)
	if i < 0 {
		return s, ""
	}
	if cutc {
		return s[:i], s[i+1:]
	}
	return s[:i], s[i:]
}
```

### What did you see instead?
```
// split slices s into two substrings separated by the first occurrence of
// sep. If cutc is true then sep is included with the second substring.
// If sep does not occur in s then s and the empty string is returned.
func split(s string, sep byte, cutc bool) (string, string) {
	i := strings.IndexByte(s, sep)
	if i < 0 {
		return s, ""
	}
	if cutc {
		return s[:i], s[i+1:]
	}
	return s[:i], s[i:]
}
```
https://github.com/golang/go/blob/39a9cb4b5dbf1e518b0c66fa3a7b4175f90226fc/src/net/url/url.go#L454-L466

jinmiaoluo:
Fixes #35735

### DIFF
--- a/src/net/url/url.go
+++ b/src/net/url/url.go
@@ -452,7 +452,7 @@ func getscheme(rawurl string) (scheme, path string, err error) {
 }
 
 // split slices s into two substrings separated by the first occurrence of
-// sep. If cutc is true then sep is included with the second substring.
+// sep. If cutc is true then sep is excluded from the second substring.
 // If sep does not occur in s then s and the empty string is returned.
 func split(s string, sep byte, cutc bool) (string, string) {
 	i := strings.IndexByte(s, sep)


jinmiaoluo:
@googlebot I signed it!

jinmiaoluo:
@googlebot I fixed it.

PR #35805, opened 2019-11-24
syscall/js: make wasm_exec.js compatible with Webpack

Lesmiscore:
In Webpack, require("fs") will always be empty. This behavior throws an error: "fs.writeSync is not function". It happens when you did "fmt.Println".
This PR avoids such problem and use polyfill in wasm_exec.js on Webpack.


nao20010128nao:
@googlebot done

PR #36065, opened 2019-12-10
`runtime:Modify runtime.Stack issu`

thotf:
sorry，My English is very poor.But I have to try to be clear again.
Examples：
	buf:=make([]byte,200,1200)
	runtime.Stack(buf,false)
Question one：
The call stack trace is formatted and written to buf。
Suppose the stack trace data is 500
Then only 200 of the data can be saved in buf。But cap (buf) = 1200

Question two：
If stack trace data is 100，
Then output to the terminal，As shown：
`E_4ABKGHOZ)%4H%2`RSMOR
The other 100 extra squares, all replaced by spaces。
This is because we are not sure about the size of the stack trace data.
so
	buf:=make([]byte,0,1200)
Let him automatically adjust the buffer space he needs.
If it's what you used to do,
if len(buf)=0, Then no data will be written to buf.

### DIFF
--- a/src/runtime/mprof.go
+++ b/src/runtime/mprof.go
@@ -783,7 +783,7 @@ func Stack(buf []byte, all bool) int {
 	}
 
 	n := 0
-	if len(buf) > 0 {
+	if cap(buf) > 0 {
 		gp := getg()
 		sp := getcallersp()
 		pc := getcallerpc()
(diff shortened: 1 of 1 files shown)


ThotfLi:
@googlebot I signed it!

ThotfLi:
Please read it again. if you can't understand it, I will ask an English expert to translate it.

agnivade:
Hi @ThotfLi, like Ian said, we can't do such a change.

Unlike other projects, we do not use the issue tracker for questions such as these. It is only used for bugs and feature proposals. I will close this, but please feel free to ask it in any of these forums below:

* [Stack Overflow](https://stackoverflow.com/questions/tagged/go) with questions tagged "go"

* [The Go Forum](https://forum.golangbridge.org/), a web-based forum

* [Gophers Slack](https://gophers.slack.com), use the [invite app](https://invite.slack.golangbridge.org/) for access. The `#general` channel is a good starting point.

* [Go Community on Hashnode](https://hashnode.com/n/go) with questions and posts tagged with "go"

* [The golang-nuts mailing list](https://groups.google.com/d/forum/golang-nuts)

* **IRC** channel #go-nuts on Freenode

Thanks

PR #36066, opened 2019-12-10
cmd/go: add -Wl,-E to linker flag whitelist (shortcut of --export-dynamic)

SilverRainZ:
According to https://linux.die.net/man/1/ld, `-E` is a shortcut of
`--export-dynamic`, it will be better to be added in to whitelist for the
later one has been added in https://golang.org/cl/134016.


SilverRainZ:
/comments I just update the pull request description, hope it can be sync to commit messsage by @gopherbot.

PR #36092, opened 2019-12-11
testing: print extra precision in the benchmark output

voronaam:
Makes benchmarks output extra precision of recorded performance
numbers, if the -v flag is used.

This change is backwards compatible and conforms to the 
specification in [1], which requires that the value be
parseable by strconv.ParseFloat.

[1]: https://go.googlesource.com/proposal/+/master/design/14313-benchmark-format.md

Fixes #34626

### DIFF
--- a/src/testing/benchmark.go
+++ b/src/testing/benchmark.go
@@ -320,7 +320,7 @@ func (b *B) launch() {
 			b.runN(int(n))
 		}
 	}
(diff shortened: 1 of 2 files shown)


voronaam:
Note that this PR uses an already existing `Chatty` flag to ask for higher precision.

Alternatively I can add a new flag to `initBenchmarkFlags` to do just that. Please let me know if this is desirable.

voronaam:
Thank you. I have updated the commit message.

PR #36152, opened 2019-12-15
net/http: Add Equal for HardwareAddr

dedalqq:
I just thought this method would be very useful.

At least I really need it. =)


dedalqq:
@googlebot I signed it!

PR #36198, opened 2019-12-18
path/filepath: fix readDirNames to not unconditionally read a directory

### LINKED ISSUE #36197: filepath: filepath.Walk readDirName will cause large memory leak
[https://github.com/golang/go/blob/master/src/path/filepath/path.go#L363](https://github.com/golang/go/blob/master/src/path/filepath/path.go#L363)

```
  // walk recursively descends path, calling walkFn.
  func walk(path string, info os.FileInfo, walkFn WalkFunc) error {
	if !info.IsDir() {
		return walkFn(path, info, nil)
	}
	names, err := readDirNames(path)
	err1 := walkFn(path, info, err)
	// If err != nil, walk can't walk into this directory.
	// err1 != nil means walkFn want walk to skip this directory or stop walking.
	// Therefore, if one of err and err1 isn't nil, walk will return.
	if err != nil || err1 != nil {
		// The caller's behavior is controlled by the return value, which is decided
		// by walkFn. walkFn may ignore err and return nil.
		// If walkFn returns SkipDir, it will be handled by the caller.
		// So walk should return whatever walkFn returns.
		return err1
	}
```

`
	names, err := readDirNames(path)
`
the readDirNames will lead large memory if directory has large number of files

which conflict skipDir purpose.

wcc526:
readDirNames was always called first to read a directory. And then depending on the value of skipDir,
we would stop walking further down. This defeated the purpose of skipDir and can possibly hold on to
a large amount of memory if the directory has a large number of files.

Fixes #36197

### DIFF
--- a/src/path/filepath/path.go
+++ b/src/path/filepath/path.go
@@ -360,17 +360,20 @@ func walk(path string, info os.FileInfo, walkFn WalkFunc) error {
 		return walkFn(path, info, nil)
 	}
 
-	names, err := readDirNames(path)
-	err1 := walkFn(path, info, err)
+	err1 := walkFn(path, info, nil)
 	// If err != nil, walk can't walk into this directory.
 	// err1 != nil means walkFn want walk to skip this directory or stop walking.
 	// Therefore, if one of err and err1 isn't nil, walk will return.
-	if err != nil || err1 != nil {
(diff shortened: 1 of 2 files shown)


wcc526:
I signed it!

PR #36228, opened 2019-12-20
syscall: change mips64 architecture Stat_t struct item to uint64 and int64

Robin-Mao:
The Dev and Pad1 item in Stat_t struct should be uint64 in 64-bits mips architecture

### DIFF
--- a/src/syscall/syscall_linux_mips64x.go
+++ b/src/syscall/syscall_linux_mips64x.go
@@ -137,8 +137,8 @@ func Iopl(level int) (err error) {
 }
 
 type stat_t struct {
-	Dev        uint32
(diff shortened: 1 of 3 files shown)


Robin-Mao:
@googlebot I signed it!

duchuanLX:
> The Dev and Pad1 item in Stat_t struct should be uint64 in 64-bits mips architecture

golang.org/x/sys@v0.0.0-20191005200804-aed5e4c7ecf9 
golang.org/x/sys@v0.0.0-20190813064441-fde4db37ae7a

i test fsnotify demo on this sys package version， cannot recv any events。

PR #36270, opened 2019-12-24
net: fix typo in dns client unix test

isfanazha:
Fix a typo DNS name, it should be [onion.vegetables.com](http://onion.vegetables.com/)


isfanazha:
@googlebot I signed it!

PR #36271, opened 2019-12-24
database/sql: support db filed value is NULL with  variable type is string

viile:
support db filed value is NULL with  struct type is string

### DIFF
--- a/src/database/sql/convert.go
+++ b/src/database/sql/convert.go
@@ -295,6 +295,8 @@ func convertAssignRows(dest, src interface{}, rows *Rows) error {
 		}
 	case nil:
 		switch d := dest.(type) {
+		case *string:
+			return nil
 		case *interface{}:
 			if d == nil {
 				return errNilPtr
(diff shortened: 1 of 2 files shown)


viile:
as follow is error code
```golang
package main

import (
	"database/sql"
	_ "github.com/go-sql-driver/mysql"
	"log"
)

func main()  {
	db, err := sql.Open("mysql", "root:@tcp(127.0.0.1:3306)/test")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()
	
	err = db.Ping()
	if err != nil {
		log.Fatal(err)
	}
	
	rows, err := db.Query("select name from test where id = 1")
	if err != nil {
		log.Fatal(err)
	}
	defer rows.Close()
	
	var name string
	for rows.Next() {
		err := rows.Scan(&name)
		if err != nil {
			log.Fatal(err)
		}
		log.Println( name)
	}
	
	err = rows.Err()
	if err != nil {
		log.Fatal(err)
	}
}
```
```html
sql: Scan error on column index 0, name "name": unsupported Scan, storing driver.Value type <nil> into type *string
```

PR #36294, opened 2019-12-28
runtime: converge duplicate calls to netpollBreak into one

panjf2000:
There might be some concurrent (maybe not concurrent, just sequential but in a short time window) and duplicate calls to `netpollBreak`, trying to wake up a net-poller. If one has called `netpollBreak` and that waking event hasn't been received by epollwait/kevent/..., then the subsequent calls of `netpollBreak` ought to be ignored or in other words, these calls should be converged into one.

Benchmarks go1.13.5 darwin/amd64:

benchmark-func           time/op (old)  time/op (new)  delta
BenchmarkNetpollBreak-4  29668ns ±1%    3131ns ±2%     -89.45%

mem/B (old)  mem/B (new)  delta
154B ±13%    0B ±0%       -100%

### DIFF
--- a/src/runtime/netpoll_epoll.go
+++ b/src/runtime/netpoll_epoll.go
@@ -6,7 +6,10 @@
 
 package runtime
 
-import "unsafe"
+import (
+	"runtime/internal/atomic"
+	"unsafe"
+)
 
 func epollcreate(size int32) int32
 func epollcreate1(flags int32) int32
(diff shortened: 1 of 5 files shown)


zhiqiangxu:
Maybe do this optimization after something like `Native.eventFdWrite` is available in go?  So that it will be done in  one place  instead of being baked into N architectures... Trading maintainability for little performance may not worth it.

PR #36318, opened 2019-12-30
runtime: refine netpollGenericInit by replacing mutex with CAS

panjf2000:
Leverage CAS to refine the sync process in `netpollGenericInit`.


proyb6:
@panjf2000 What CAS stands?

PR #36382, opened 2020-01-04
net/http: allow upgrading non keepalive connections

### LINKED ISSUE #36381: net/http: Allow upgrading connections even if DisableKeepAlives is set
If the http transport has DisableKeepAlives enabled, then connection upgrading does not work. See https://github.com/nhooyr/websocket/issues/185

If DisableKeepAlives is set, then the client sets `Connection: close` on every request which is completely fine.

The problem is that when the server encounters such a request, it prevents any http.Handler from setting the `Connection` header.

See https://github.com/golang/go/blob/3a6cd4c72d5c8fc6c86f074eadaaf3972e591973/src/net/http/server.go#L1293

If that is changed to:

```go
	} else if !w.req.ProtoAtLeast(1, 1) ||
		// Only close if the request indicates the client wants to close
		// and we are not upgrading the connection.
		(w.wantsClose && header.get("Connection") != "Upgrade") {
		w.closeAfterReply = true
	}
```

Then everything works fine.

nhooyr:
If one was using http.Transport with DisableKeepAlives and trying
to upgrade a connection against net/http's Server, the Server
would not allow a "Connection: Upgrade" header to be written
and instead override it to "Connection: Close" which would
break the handshake.

This change ensures net/http's Server does not override the
connection header for successful protocol switch responses.

Fixes #36381.

### DIFF
--- a/src/net/http/response.go
+++ b/src/net/http/response.go
@@ -352,10 +352,16 @@ func (r *Response) bodyIsWritable() bool {
 	return ok
 }
 
-// isProtocolSwitch reports whether r is a response to a successful
-// protocol upgrade.
+// isProtocolSwitch reports whether the response code and header
+// indicate a successful protocol upgrade response.
 func (r *Response) isProtocolSwitch() bool {
-	return r.StatusCode == StatusSwitchingProtocols &&
-		r.Header.Get("Upgrade") != "" &&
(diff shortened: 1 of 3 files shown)


nhooyr:
@bradfitz From whom should I request further reviews to get this merged?

PR #36491, opened 2020-01-10
crypto/x509: this change modifies C.CopyPEMRoots to static function

bxq2011hust:


bxq2011hust:
> @googlebot I signed it!

@googlebot I signed it!

PR #36493, opened 2020-01-10
Update go_spec.html

AlphaHot:
Translated from French to English


mvdan:
This is on purpose. The CLA was never signed either.

PR #36500, opened 2020-01-10
crypto/tls: create certs w/o KeyEncipherment KU for non-RSA keys in generate_cert.go

### LINKED ISSUE #36499: crypto/tls: generate_cert.go sets KeyEncipherment KU for non-RSA keys [freeze exception]

<pre>
go version go1.13.4 linux/amd64
</pre>

Yes

N/A

### What did you do?

```
$ go run ./src/crypto/tls/generate_cert.go -host threeletter.agency -ecdsa-curve=P256

$ openssl x509 -in cert.pem -noout -text | grep -A1 "X509v3 Key Usage:"
```

```
            X509v3 Key Usage: critical
                Digital Signature
```

### What did you see instead?

```
            X509v3 Key Usage: critical
                Digital Signature, Key Encipherment
```

### Summary:

The `crypto/tls/generate_cert.go` utility should only set the template `x509.Certificate`'s `KeyUsage` field to a value with the `x509.KeyUsageKeyEncipherment` bits set when the certificate subject public key is an RSA public key, not an ECDSA or ED25519 public key.

Presently it sets the `KeyUsage` to `KeyUsageKeyEncipherment` and `KeyUsageDigitalSignature` no matter what type of public key is used: https://github.com/golang/go/blob/79ccbe1b67b9c481c73bd6b310849bbcf13954a5/src/crypto/tls/generate_cert.go#L110

[RFC 5480](https://tools.ietf.org/html/rfc5480) describes the usage of ECDSA elliptic curve subject keys with X509. Unfortunately while Section 3 "Key Usages Bits" indicates which key usage bits MAY be used with a certificate that indicates `id-ecPublicKey` in the `SubjectPublicKeyInfo` field it doesn't provide guidance on which usages should not be included (e.g. the `keyEncipherment` bit, which is particular to RSA key exchange). The same problem is present in [RFC 8410](https://tools.ietf.org/html/rfc8410) Section 5 describing Key Usage Bits for ED25519 elliptic curve subject keys.

There's an update to RFC 5480 in last call stage within the IETF LAMPS WG, [draft-ietf-lamps-5480-ku-clarifications-00](https://tools.ietf.org/html/draft-ietf-lamps-5480-ku-clarifications-00). This update is meant to clarify the allowed Key Usages extension values for certificates with ECDSA subject public keys by adding: 

> If the keyUsage extension is present in a certificate that indicates id-ecPublicKey as algorithm of AlgorithmIdentifier [RFC2986] in SubjectPublicKeyInfo, then following values MUST NOT be present:
>
> keyEncipherment; and
> dataEncipherment.

I don't believe there is an update for RFC 8410 in the works but I suspect it will be clarified similarly in the future (I will follow up with the LAMPS WG).

The current behaviour of `generate_cert.go` won't comply with the updated RFC 5480 requirement.

cpu:
Summary

The crypto/tls/generate_cert.go utility should only set the template
x509.Certificate's KeyUsage field to a value with the
x509.KeyUsageKeyEncipherment bits set when the certificate subject
public key is an RSA public key, not an ECDSA or ED25519 public key.

Background

RFC 5480 describes the usage of ECDSA elliptic curve subject keys with
X.509. Unfortunately while Section 3 "Key Usages Bits" indicates which
key usage bits MAY be used with a certificate that indicates
id-ecPublicKey in the SubjectPublicKeyInfo field it doesn't provide
guidance on which usages should *not* be included (e.g. the
keyEncipherment bit, which is particular to RSA key exchange). The same
problem is present in RFC 8410 Section 5 describing Key Usage Bits for
ED25519 elliptic curve subject keys.

There's an update to RFC 5480 in last call stage within the IETF LAMPS
WG, draft-ietf-lamps-5480-ku-clarifications-00. This update is meant
to clarify the allowed Key Usages extension values for certificates with
ECDSA subject public keys by adding:

> If the keyUsage extension is present in a certificate that indicates
> id-ecPublicKey as algorithm of AlgorithmIdentifier [RFC2986] in
> SubjectPublicKeyInfo, then following values MUST NOT be present:
>
> keyEncipherment; and
> dataEncipherment.

I don't believe there is an update for RFC 8410 in the works but I
suspect it will be clarified similarly in the future.

This commit updates generate_cert.go to ensure when the certificate
public key is ECDSA or ED25519 the generated certificate has the
x509.Certificate.KeyUsage field set to a value that doesn't include KUs
specific to RSA. For ECDSA keys this will adhere to the updated RFC 5480
language.

Fixes #36499

### DIFF
--- a/src/crypto/tls/generate_cert.go
+++ b/src/crypto/tls/generate_cert.go
@@ -81,6 +81,16 @@ func main() {
 		log.Fatalf("Failed to generate private key: %v", err)
 	}
 
+	// Both ECDSA and RSA subject keys should have the DigitalSignature KeyUsage
+	// bits set in the x509.Certificate template
+	keyUsage := x509.KeyUsageDigitalSignature
+	// Only RSA subject keys should have the KeyEncipherment KeyUsage bits set. In
+	// the context of TLS this KeyUsage is particular to RSA key exchange and
+	// authentication.
+	if _, isRSA := priv.(*rsa.PrivateKey); isRSA {
+		keyUsage |= x509.KeyUsageKeyEncipherment
+	}
+
 	var notBefore time.Time
 	if len(*validFrom) == 0 {
 		notBefore = time.Now()
@@ -107,7 +117,7 @@ func main() {
 		NotBefore: notBefore,
 		NotAfter:  notAfter,
 
-		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		KeyUsage:              keyUsage,
 		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
 		BasicConstraintsValid: true,
 	}
--- a/src/crypto/tls/generate_cert.go
+++ b/src/crypto/tls/generate_cert.go
@@ -81,8 +81,8 @@ func main() {
 		log.Fatalf("Failed to generate private key: %v", err)
 	}
 
-	// Both ECDSA and RSA subject keys should have the DigitalSignature KeyUsage
-	// bits set in the x509.Certificate template
+	// ECDSA, ED25519 and RSA subject keys should have the DigitalSignature
+	// KeyUsage bits set in the x509.Certificate template
 	keyUsage := x509.KeyUsageDigitalSignature
 	// Only RSA subject keys should have the KeyEncipherment KeyUsage bits set. In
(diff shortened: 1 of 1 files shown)


cpu:
_Quick note: This is my first Go PR. I've tried to adhere to the guidance in the CONTRIBUTING.md but would appreciate some extra patience from maintainers if I've made any mistakes_

PR #36537, opened 2020-01-13
fix of #36535 issue

### LINKED ISSUE #36535: net/http: unexpected behavour of client.Do() with cookie jar

<pre>
$ go version go1.12.7 linux/amd64
</pre>

Yes, stable 1.13.6 version has same problem too.

Any platform with any enviroment

### What did you do?
code is here: https://play.golang.org/p/M7q1GhoW1wV

So, for my opinion, the problem is in r.AddCookie() function, which don't check request header is nil and forcely call Headers methods. :thinking: 

My projects must not panicing for situations like this :+1: 

i think, that request initializing with `&http.Request{}` instead `http.NewRequest()` is more convenient, cause `http.NewRequest()` does too many things. Declaration of request object looks more readable and simple.

### What did you see instead?

They are panicing! :open_mouth:

quenbyako:
fix #36535

### DIFF
--- a/src/net/http/request.go
+++ b/src/net/http/request.go
@@ -431,6 +431,10 @@ func (r *Request) Cookie(name string) (*Cookie, error) {
 // means all cookies, if any, are written into the same line,
 // separated by semicolon.
 func (r *Request) AddCookie(c *Cookie) {
+	if r.Header == nil {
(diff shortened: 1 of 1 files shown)


ololosha228:
@googlebot I signed it!

PR #36674, opened 2020-01-21
 doc/articles/wiki: add missing log import to net/http tutorial

ChiragSukhala:
The log package is used with the net/http but was not in the import clause.


ChiragSukhala:
@googlebot I signed it!

PR #36678, opened 2020-01-21
net/http/httputil: add support for X-Forwarded-Proto, X-Forwarded-Host and an option to not trust forwarded headers in ReverseProxy

dunglas:
In addition to X-Forwarded-For, which is already supported but undocumented
(#36672), this patch adds support for the X-Forwarded-Proto (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)
and X-Forwarded-Host (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host)
headers in ReverseProxy.
It also introduces a new option, ReverseProxy.TrustForwardedHeaders
that must be explicitly set to true to trust forwarded headers coming from
the client or previous proxies.

Setting these headers is a standard behavior expected from reverse
proxies, especially when X-Forwarded-For is set (which is already the
case).

The new ReverseProxy.TrustForwardedHeaders option fixes a potential
common security issue in user land code. It forces the user to explicitly
opt-in to trust reverse proxies in front of the Go application using
ReverseProxy.

The current behavior of blindly trusting the X-Forwarded-For header
provided by the client is very dangerous.
For instance, because of the current (undocumented) behavior, the Vulcain
gateway server (https://vulcain.rocks) was sensible to a potential security
issue.

If this flag is not set to true, previous values set by the client or
untrusted proxies will be discarded (replaced by the values computed
by ReverseProxy).
Technically, this is a BC break, but, in my opinion, it is a necessary one
to improve the security of the whole Go ecosystem using this module.

Edit: added a backward compatible implementation, see https://github.com/golang/go/pull/36678#issuecomment-577163525

### DIFF
--- a/src/net/http/httputil/reverseproxy.go
+++ b/src/net/http/httputil/reverseproxy.go
@@ -24,23 +24,37 @@ import (
 	"golang.org/x/net/http/httpguts"
 )
 
+type ForwardedHeaderBehavior uint8
+
+const (
+	// Overwrite existing X-Forwarded-For, X-Forwarded-Proto
+	// and X-Forwarded-Host HTTP headers with values extracted
+	// from the current connection with the client.
+	// As a special case, headers explicitly set to nil
+	// are never modified.
+	ForwardedHeaderOverwrite ForwardedHeaderBehavior = iota
+
+	// Add client IP to the existing IPs in X-Forwarded-For.
+	// Preserve X-Forwarded-Proto and X-Forwarded-Host if set.
+	// Ensure that the previous proxies in the chain are trusted
+	// As a special case, headers explicitly set to nil
+	// are never modified.
+	// when using this value, or it will lead to security issues.
+	ForwardedHeaderAdd
+
+	// Preserve all existing forwarded HTTP headers.
+	// Ensure that the previous proxies in the chain are trusted
+	// when using this value, or it will lead to security issues.
+	ForwardedHeaderPreserve
+)
+
 // ReverseProxy is an HTTP Handler that takes an incoming request and
 // sends it to another server, proxying the response back to the
 // client.
 //
 // ReverseProxy by default sets the client IP as the value of the
-// X-Forwarded-For header.
-//
-// If an X-Forwarded-For header already exists, the client IP is
-// appended to the existing values. As a special case, if the header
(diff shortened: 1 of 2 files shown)


artyom:
Can this be a documentation update, maybe with an example of httputil.ReverseProxy.Director function handling headers in question, instead of a backwards-incompatible change?

Note that even if this change is accepted, at least some ReverseProxy users would need to support pre-change and post-change Go versions, which adds maintenance overhead.

dunglas:
@artyom Regarding `TrustForwardedHeaders` it's also possible to make the change backward compatible by setting `TrustForwardedHeaders` to true by default, or by renaming it `OverrideForwardedHeaders`, but IMHO it's  less secure (as secure than currently actually), and it's always better to have to opt-in for potentially dangerous features (the current behavior as already be proven dangerous). Anyway, your call. I'm perfectly ok to make it backward compatible if you think it's better for the ecosystem.

Regarding setting `X-Forwarded-Proto` and `X-Forwarded-Host`, I think that it's a must have feature (especially because `X-Forwarded-For` is already set). It's the de facto standard, the behavior of most reverse proxies and the expected behavior of most web application frameworks, including Symfony, Laravel and Gorilla to name a few.

dunglas:
@artyom I just pushed a new patch containing a backward compatible implementation. WDYT about this one?

src/net/http/httputil/reverseproxy.go
```diff
@@ -244,6 +259,17 @@ func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		outreq.Header.Set("Upgrade", reqUpType)
 	}
 
+	if p.OverwriteForwardedHeaders {
+		proto := "https"
+		if req.TLS == nil {
```
ruudk, line 264: Not sure about this. `req.TLS` is only set when the current http server is handling the TLS. But if TLS is done by a load balancer in front of this server, it will say `http`, how can the backend know if it was an `https` request originally?

dunglas: Then you'll have to handle this manually. If the TLS termination is handled by a non-HTTP server, the proxy cannot guess it.

PR #36740, opened 2020-01-24
cmd/go: give error when specifying non *.go files in run

ghost:
When running `go run a.go b.cc c.go`, go will correctly only run `a.go`.
After this change, it will give an error `go run: cannot run non *.go files (b.cc)`.


cheng-chang:
@googlebot I signed it!

PR #37228, opened 2020-02-14
errors: add new function Temporary(error) bool

aka-rider:
This is proposal to add the Temporary function into errors package. I tend to rewrite it over and over again in every project.

Mainly, this function will be used in conjunction with the net package. 
Nevertheless, I think Temporary function belongs to the erros package, in this case user code and other packages may follow the same convention.

### DIFF
--- a/src/errors/errors.go
+++ b/src/errors/errors.go
@@ -67,3 +67,11 @@ type errorString struct {
 func (e *errorString) Error() string {
 	return e.s
 }
+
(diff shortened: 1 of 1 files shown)


aka-rider:
@googlebot I signed it!

aka-rider:
For tracking [https://github.com/golang/go/issues/37250](https://github.com/golang/go/issues/37250)

PR #37251, opened 2020-02-16
sort: Fixes IntSlice Less function if duplicate values

decanus:
This PR fixes IntSlice less function to contain a smaller than or equal to. This should be the case in case an array contains duplicate values, currently an array with duplicate values is never considered sorted.

I am not 100% certain this change will work due to how sorting functions work. If `Less` is used to sort it may be caught in an endless loop.

### DIFF
--- a/src/sort/sort.go
+++ b/src/sort/sort.go
@@ -269,7 +269,7 @@ func IsSorted(data Interface) bool {
 type IntSlice []int
 
 func (p IntSlice) Len() int           { return len(p) }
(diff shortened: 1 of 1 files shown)


decanus:
@googlebot I signed it!

DisposaBoy:
Maybe add a test that demonstrates the issue?

It seems to work correctly as-is: https://play.golang.org/p/WWXtJ7MdOFL

Issue #35867, opened 2019-11-27
x/build/cmd/gerritbot: didn't close a PR that was submitted


dmitshur:
We should investigate why gerritbot didn't close a PR that was imported into Gerrit as a CL and submitted. 

- PR #35751
    - imported as [CL 208318](https://golang.org/cl/208318)

See https://github.com/golang/go/pull/35751#issuecomment-558859722 for more details.

Thanks @odeke-em for noticing and reporting.

dmitshur:
This seems to be happening again with PR #37328. It has been [submitted on Gerrit](https://go-review.googlesource.com/c/go/+/220259/3#message-eacca2c63120bee7580861a608a0d804d5bfe42a) 30 minutes ago, yet `gerritbot` isn't closing it.

From its logs:

```
2020/02/22 15:49:57 Retrieving PR golang/go#37328 from GitHub using Etag "W/\"54eea6bb1cf5f438654ec4e769f55b80\"" ...
2020/02/22 15:49:57 GitHub: 3852/5000 calls remaining; Reset in 23m32.142630856s
2020/02/22 15:49:57 Returning cached version of golang/go#37328
2020/02/22 15:49:57 Processing PR https://github.com/golang/go/pull/37328 ...
2020/02/22 15:49:57 Changes for PR golang/go#37328 have yet to be mirrored in the maintner corpus. Skipping for now.
```

So `gerritbot` thinks that the PR hasn't yet been mirrored in the maintner corpus.

But querying maintner directly, I can see that it has latest data. That PR description was never edited.

There's likely some sort of logic problem in `gerritbot`.

dmitshur:
Another case affected PR #58779. It was submitted on Gerrit 5 days ago and the PR wasn't marked as closed.

It once again got stuck on `Changes for PR golang/go#58779 have yet to be mirrored in the maintner corpus. Skipping for now.` indefinitely. There seems to be a race. The relevant code in `bot.processPullRequest` is:

```Go
cl := b.importedPRs[shortLink]

if cl != nil && b.pendingCLs[shortLink] == cl.Commit.Msg {
	delete(b.pendingCLs, shortLink)
}
if b.pendingCLs[shortLink] != "" {
	log.Printf("Changes for PR %s have yet to be mirrored in the maintner corpus. Skipping for now.", shortLink)
	return nil
}
```

[...]

PR #37328, opened 2020-02-20
hash/maphash: don't discard data on random seed init

### LINKED ISSUE #37315: hash/maphash: first call of the Reset method doesn't reset internal state
The new `hash/maphash` package has a [`Reset` method](https://tip.golang.org/pkg/hash/maphash/#Hash.Reset) documented as:

> Reset discards all bytes added to h. (The seed remains the same.) 

Moreover, the [`Sum64` doc](https://tip.golang.org/pkg/hash/maphash/#Hash.Sum64) says:

> Sum64 returns h's current 64-bit value, which **depends on h's seed and the sequence of bytes added to h since the last call to Reset**

So it appears it would be reasonable to expect this to print two times the same value:

```
package main

import (
	"fmt"
	"hash/maphash"
)

func main() {
	h := new(maphash.Hash)

	h.WriteString("hello")
	fmt.Printf("%#x\n", h.Sum64())

	h.Reset()

	h.WriteString("hello")
	fmt.Printf("%#x\n", h.Sum64())
}
```

But it doesn't:

```
$ gotip run hash.go
0xdf1aabcc36e56272
0x5a04b54a5a2174d6
```

This is because even if the seed remains the same and "all the previously added bytes are discarded", *the internal state of the Hash is maintained*, even after the `h.Reset()` call.

Is this the desired behaviour? If it is, the documentation should probably be amended to make it clear what to expect after a call to the `Reset` method.

cc @alandonovan @randall77

vovapi:
Hash initializes seed on the first usage of seed or state with initSeed.
initSeed uses SetSeed which discards accumulated data.
This causes hash to return different sums for the same data in the first use
and after reset.
This CL fixes this issue by separating the seed set from data discard.

Fixes #37315

### DIFF
--- a/src/hash/maphash/maphash.go
+++ b/src/hash/maphash/maphash.go
@@ -69,7 +69,7 @@ type Hash struct {
 // which does call h.initSeed.)
 func (h *Hash) initSeed() {
 	if h.seed.s == 0 {
-		h.SetSeed(MakeSeed())
+		h.setSeed(MakeSeed())
 	}
 }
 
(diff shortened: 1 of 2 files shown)


dmitshur:
Closing because this PR has been merged via [CL 220617](https://golang.org/cl/220617) as commit 638df87fa4f927763f99ebf0c6bc9c4a5380d1f9.

`gerritbot` should've done this; we need to investigate why it didn't. See https://github.com/golang/go/issues/35867#issuecomment-589971216.

PR #37403, opened 2020-02-24
cmd/internal/objabi: enable frame-pointer for iOS arm64

gawen:
This improves stack unwinding of Go code running on iOS arm64.


steeve:
We (Zenly) have been using this patch successfully in production for the last few months.

PR #37530, opened 2020-02-27
flag: exit 0 when -h or -help invoked but undefined

sding3:
flag treats -h or -help as a special case to print a nice help
message, but exit with a status code of 2. This update makes
that status code 0.

Fixes #37533


sding3:
@googlebot I signed it!

PR #37563, opened 2020-02-28
doc: update Effective Go with 1.14 language changes

### LINKED ISSUE #37560: doc: update Effective Go with 1.14 language changes
In the doc of Effective Go it still states that embedded interfaces must provide disjoint sets of methods which has been changed with Go 1.14 and is no longer true.
https://github.com/golang/go/blob/2cf3ebaf3db5b23a56e2ee62ecd76748dfbb5b8e/doc/effective_go.html#L2698-L2699

fcornelius:
Fixes #37560


fcornelius:
@googlebot I signed it!

PR #37580, opened 2020-02-29
path/filepath: return the correct extensions when using filepath.Ext(path)

Zarathustra2:
Previously filepath.Ext(path) would return incorrect values for paths such as _path/.config_ and files such as _.gitignore_ as both of them do not have an extension.

This PR addresses this issue.


Zarathustra2:
Closed as it would break backwards compatibility as stated by Ian.

Issue #37650, opened 2020-03-04
net/http/httptest: using Recorder.HeaderMap contains headers that were written after http.ResponseWriter.WriteHeader is called


elgohr:
https://github.com/golang/go/issues/14914#issue-142720976
Reproducible in go1.14.

elgohr:
Workaround, use:
```
res.Result().Header.Get("Content-Type")
```
Having two ways for accessing the header feels very confusing to me.

odeke-em:
Hello @elgohr, thank you for filing this issue!

Could you please explain what the problem is and provide sample code instead of perhaps a link to a comment in a bug that was fixed?

Did you mean net/http/httptest.ResponseRecorder or net/http/httptest.Server? I can't seem to reproduce this problem and also we have functioning tests that ensure that this doesn't happen.

However, if you declared a Trailer before invoking w.WriteHeader, that trailer will be sent along,
when set later on.

Given that you didn't mention where you are getting the problem but I can allude to the usage of net/http/httptest.ResponseRecorder, I shall provide you with 2 different scenarios, and please tweak the code to see if it reproduces your issue.

### 1. net/http/httptest.ResponseRecorder
```go
package main

import (
        "net/http"
        "net/http/httptest"
        "net/http/httputil"
        "testing"
)

func handler(w http.ResponseWriter, r *http.Request) {
        // Declare this Trailer that will be permissible after
        // after w.WriteHeader is invoked.
        w.Header().Set("Trailer", "AB")
        w.WriteHeader(200)
        w.Write([]byte("Aloha"))
        // Throw in this stray header that was not declared as a Trailer.
        w.Header().Set("Stray", "Bar")
        w.Header().Set("AB", "CD")
}

func TestHeaderWriteInResponseRecorder(t *testing.T) {
        req := httptest.NewRequest("GET", "https://example.com/", nil)
        rec := httptest.NewRecorder()

        handler(rec, req)
        resp := rec.Result()
        defer resp.Body.Close()

        blob, _ := httputil.DumpResponse(resp, true)
        want := "HTTP/1.1 200 OK\r\nConnection: close\r\n\r\nAloha"

        if g, w := string(blob), want; g != w {
                t.Fatalf("Response mismatch\nGot:\n%q\n\nWant:\n%q", g, w)
        }
        if g, w := resp.Trailer.Get("AB"), "CD"; g != w {
                t.Fatalf("Trailer mismatch\nGot:  %q\nWant: %q", g, w)
        }
[...]

PR #37684, opened 2020-03-05
net/http/httptest: adds test for #37650

elgohr:
This PR adds tests for https://github.com/golang/go/issues/37650.
It does not add a fix, so please take this as a starting point.


elgohr:
Sorry, please ignore as this was not really testing the issue.

PR #37723, opened 2020-03-06
cmd/cgo: avoid extra newline in doc comment of exported function

### LINKED ISSUE #37722: cmd/cgo: CGo Exported Function has an empty line between it and its documentation block
I'm writing a library in Go that I'm compiling to a native library using CGo, however when the functions are exported they have this annoying empty line between the function declaration and the documentation comment.

As an example, let's say I have this in my go file

```go
/**
 * Prints an integer.
 *
 * @param p0 an integer to print.
 */
//export PrintInteger
func PrintInteger(val C.int) {
    fmt.Printf("%v", val);
}
```

Once it's exported in my C header file, it looks like this

```c
/**
 * Prints an integer.
 *
 * @param p0 an integer to print.
 */

extern void PrintInteger(int p0);
```

That empty line for the most part isn't a problem, but on some IDEs when parsing that header file they have difficulty properly displaying the intellisense, making consumption of the header slightly more difficult. Especially when you factor in the way that cgo [renames function parameters](https://stackoverflow.com/questions/57511799/how-do-i-get-proper-parameter-names-in-cgo-exported-functions)

I've found _what i believe_ is the area in the go source code that causes this:

https://github.com/golang/go/blob/master/src/cmd/cgo/out.go#L922

```go
if len(exp.Doc) > 0 {
    fmt.Fprintf(fgcch, "\n%s", exp.Doc)
}
fmt.Fprintf(fgcch, "\nextern %s;\n", s)
```

I feel like perhaps that `\n` before the `extern` keyword shouldn't be there. Unless this is intentional?

nathan-fiscaletti:
Fixes #37722

### DIFF
--- a/src/cmd/cgo/out.go
+++ b/src/cmd/cgo/out.go
@@ -921,8 +921,11 @@ func (p *Package) writeExports(fgo2, fm, fgcc, fgcch io.Writer) {
 
 		if len(exp.Doc) > 0 {
 			fmt.Fprintf(fgcch, "\n%s", exp.Doc)
+			if !strings.HasSuffix(exp.Doc, "\n") {
+				fmt.Fprint(fgcch, "\n")
+			}
 		}
-		fmt.Fprintf(fgcch, "\nextern %s;\n", s)
+		fmt.Fprintf(fgcch, "extern %s;\n", s)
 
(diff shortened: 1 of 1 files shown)


nathan-fiscaletti:
@googlebot I signed it!

sdadas11:
orz

PR #37730, opened 2020-03-07
go/doc: fix detection of whole file examples

flyingmutant:
After CL 211357 (commit 499dc1c),
hasTests and numDecl were not updated properly for function
declarations with parameters, which affected the whole file
example detection logic. This caused examples like

	package foo_test

	func Foo(x int) {
	}

	func Example() {
		fmt.Println("Hello, world!")
		// Output: Hello, world!
	}

to not be detected as whole file ones.


flyingmutant:
What are the steps to get this reviewed? It has been almost a month with no updates.

PR #37812, opened 2020-03-12
net/http/httputil: improve compatibility with server sent events

sadag:
According to https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model
MIME type parameters should be ignored when Content-Type` header value is `text/event-stream`.

Related https://github.com/spring-projects/spring-framework/pull/24632


neild:
Gerrit isn't letting me comment on https://golang.org/cl/223118. ("Error 409: Change 223118 may not exceed 5000 updates.") I'm not sure why the change has 5000 updates on it, but I'm going to force abandon it since it's broken.

This change should use mime.ParseMediaType to parse the Content-Type header, rather than a prefix match.

I've created https://golang.org/cl/340529 with that change. If you'd like to make the change yourself, please feel free to create a new CL and send it to me for review and I'll abandon 340529.

PR #37932, opened 2020-03-18
database/sql: fix variable name in example

alicebob:
It's a very minor error, but it's a bad copy/paste example.


alicebob:
@googlebot I signed it!

PR #37981, opened 2020-03-21
math/big: fix typo in Lucas chain

sukop:
The top bit of `nat` is at `nat.bitLen()-1` (and not `nat.bitLen()`), I assume it is just a typo.

### DIFF
--- a/src/math/big/prime.go
+++ b/src/math/big/prime.go
@@ -250,7 +250,7 @@ func (n nat) probablyPrimeLucas() bool {
 	vk := nat(nil).setWord(2)
 	vk1 := nat(nil).setWord(p)
 	t2 := nat(nil) // temp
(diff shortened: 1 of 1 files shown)


sukop:
@griesemer Starting from `nat.bitLen()`, i.e. from zero, does not advance the chain, it keeps the initial `vk` and `vk1` unchanged. If we trace the execution by hand:

    vk1 = p = V(k) V(k+1) - P = 2*p - p = p
    vk = 2 = V(k)² - 2 = 2² - 2 = 2

odeke-em:
@sukop thank you for this change and for the reply! Unfortunatley Gerrit doesn't mirror comments from here, so please go to reply to Robert's comments at https://go-review.googlesource.com/c/go/+/224584#message-2fd1fc0863b9c3549863813c65183d91d85e15cd.

PR #38021, opened 2020-03-23
net/http/httputil: make Switching Protocol requests (e.g. Websockets) cancelable

### LINKED ISSUE #35559: net/http/httputil: ReverseProxy - websocket connections cannot be canceled

<pre>
$ go version
go version go1.13.4 linux/amd64
</pre>
(also checked master)

Yes

Up to date Arch Linux amd64
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/pierre/.cache/go-build"
GOENV="/home/pierre/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/pierre/go"
GOPRIVATE=""
GOROOT="/usr/lib/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build415071953=/tmp/go-build -gno-record-gcc-switches"
</pre></details>

### What did you do?

* I used `httputil.NewSingleHostReverseProxy` to create a proxy to a backend server which serves websockets.
* I made a request to a backend websocket service through the proxy. The request is cancelable using `context.WithCancel(...)`.
* I canceled the request

I expected the reverse proxy to close the connection to the backend server.

### What did you see instead?

The connection between the reverse proxy and the backend server was preserved.

---

Hi All,

I quickly modified the `TestReverseProxyWebSocket` test to reproduce my issue https://github.com/eshard/go/commit/73147b8fc8f9cb23b71b84a9a9b8af5d39d12e5b (I modified TestReverseProxyWebSocket in place to make the diff more readable)
I also did a dirty fix: https://github.com/eshard/go/commit/98e746e9b583414fe73d622ea4c775641faa6744 which propagates the cancelation to the `handleUpgradeResponse` function. It creates an additional goroutine so it's not ideal.

I'd be happy to improve my patch with your inputs

Best regards
Pierre

piec:
Ensures that a canceled client request for Switching Protocols
(e.g. h2c, Websockets) will cause the underlying connection to
be terminated.

Adds a goroutine in handleUpgradeResponse in order to select on
the incoming client request's context and appropriately cancel it.

Fixes #35559

### DIFF
--- a/src/net/http/httputil/reverseproxy.go
+++ b/src/net/http/httputil/reverseproxy.go
@@ -526,7 +526,20 @@ func (p *ReverseProxy) handleUpgradeResponse(rw http.ResponseWriter, req *http.R
 		p.getErrorHandler()(rw, req, fmt.Errorf("internal error: 101 switching protocols response with non-writable body"))
 		return
 	}
-	defer backConn.Close()
+
+	backConnCloseCh := make(chan bool)
+	go func() {
+		// Ensure that the cancelation of a request closes the backend.
+		// See issue https://golang.org/issue/35559.
+		select {
+		case <-req.Context().Done():
+		case <-backConnCloseCh:
+		}
+		backConn.Close()
+	}()
+
+	defer close(backConnCloseCh)
+
 	conn, brw, err := hj.Hijack()
 	if err != nil {
 		p.getErrorHandler()(rw, req, fmt.Errorf("Hijack failed on protocol switch: %v", err))
--- a/src/net/http/httputil/reverseproxy_test.go
+++ b/src/net/http/httputil/reverseproxy_test.go
@@ -1158,6 +1158,128 @@ func TestReverseProxyWebSocket(t *testing.T) {
 	}
 }
 
+func TestReverseProxyWebSocketCancelation(t *testing.T) {
+	n := 5
+	triggerCancelCh := make(chan bool, n)
+	nthResponse := func(i int) string {
+		return fmt.Sprintf("backend response #%d\n", i)
+	}
+	terminalMsg := "final message"
+
+	cst := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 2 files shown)


piec:
Hi @odeke-em, I think I've treated more or less all of your review :)

piec:
Is the goroutine scheduling different on the wasm runtime?
I'll think about making the test less sensitive to how coroutines are scheduled

piec:
Hi @odeke-em,
I think the current implementation of the test is sensitive to race conditions, we have no guarantee that the coroutine that handles the cancelation actually gets some time to run. I don't know how the go runtime schedules coroutines on wasm but I suppose it's quite different from other targets and shows this issue even more.
I originally implemented the test with a timeout the same way as [`TestReverseProxyCancelation`]( https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy_test.go#L389) to be sure that all coroutines are given a chance to run to exhaustion and the connection is actually canceled. If the implementation is buggy then we'll hit the timeout otherwise the test runs in a fraction of second as usual.
I think I'll push a modified version of the test with the timeout, what do you think?

piec:
Tested locally with (GOARCH=amd64, GOOS=linux) and (GOOS=js, GOARCH=wasm).
It seems fine to me :)
(It's also rebased on master)

PR #38027, opened 2020-03-23
path: enforce glob pattern syntax validation on Match function

### LINKED ISSUE #28614: path/filepath: always validate full Glob patterns
The functions `filepath.Match(pattern, name)` and `path.Match(pattern, name)` return `ErrBadPattern` if the pattern is invalid, but due to lazy evaluation the error may or may not occur depending on what you match.

Example:

```golang
_, err := filepath.Match("a[", "a") // err is nil
_, err := filepath.Match("a[", "abc") // err is "syntax error in pattern"
```

As patterns are often user input, it would be handy to be able to validate the pattern. If the validation is successful, it should be guaranteed that `filepath.Match()` will not return an error.

I created a PR for this [here](https://go-review.googlesource.com/c/go/+/143477) and was asked to open a proposal issue first.

Other programming languages seem to either not validate the pattern at all (like Python's `glob.glob('a[')`), or always validate the pattern (like Java's `FileSystems.getDefault().getPathMatcher("glob: a["))`). I couldn't find an example where validation depends on what name is matched.

jmartin82:
The current implementation of the glob in Match(pattern, name) function
is evaluating the pattern syntax in a lazy way, this can lead to some mistakes,
due to a call with an invalid pattern can end up without match or
returning ErrBadPattern error depending on the input.

This change forces the pattern syntax validation, avoiding inconsistent error
return depending on the input, this change shouldn't break the compatibility
even without the correct error handling but can imply a performance penalty
when the input is empty.

Fixes #28614

Change-Id: I055a3b707b93f8ddda87551effb90f5e767a4117


mvdan:
CL was abandoned at the request of the author.

PR #38041, opened 2020-03-24
sql/driver: ensure driver not nil before lock

panguncle:
chore: ensure driver not nil before lock


panguncle:
@googlebot I signed it!

PR #38430, opened 2020-04-14
container/list: modify the move function

GeorgeSen:


GeorgeSen:
@googlebot I signed it!

PR #38534, opened 2020-04-20
net/http: fix infinite 301 redirects of http.FileServe

### LINKED ISSUE #13996: net/http: ServeFile bug when directory address ends without "/"
go version 1.5.3

Suppose a directory Documents/ has sub-directories A/, B/, C/, and files a, b, c

URL in address bar: **http://localhost/Documents/**

Everything is okay

URL in address bar: **http://localhost/Documents**

Sub-directory's href is "**http://localhost/A/**", NOT "**http://localhost/Documents/A/**".

So do other sub-directories and files.

-
https://github.com/golang/go/blob/master/src/net/http/fs.go#L458-L461

Why is the **redirect** flag in **serveFile(w, r, Dir(dir), file, false)** set to be false?

What's the difference between ServeFile and FileServer except that FileServer will add "/" for directories and remove "/" for files?

As suggestion, I hope that ServeFile won't show directory structure. It only serves files.

## Edited

Code: http://130.211.241.67:3000/test.go

The following 2 URLs has the same html output. But as two URLs have different endings, broswer generates different links.
http://130.211.241.67:3000/test

http://130.211.241.67:3000/test/

Just click "testFile"

The ServeFile function doesn't redirect directory URLs ends without "/" to right place.

If ServeFile redirects, there are no difference between ServeFile and FileServer.

@bradfitz tested 

https://github.com/golang/go/issues/13996#issuecomment-172754548

So I wonder how to modify it.

## Edited Jan 20, 2016, 13:40:00 UTC+8
@bradfitz 's test
> If I change ServeFile's false to true and re-run the net/http tests, there are failures

https://github.com/golang/go/issues/13996#issuecomment-172754548

medcl:
This change modifies If the url is already root "/", don't cause additional redirect. Fixes:  #13996

### DIFF
--- a/src/net/http/fs.go
+++ b/src/net/http/fs.go
@@ -579,7 +579,7 @@ func serveFile(w ResponseWriter, r *Request, fs FileSystem, name string, redirec
 				return
 			}
 		} else {
-			if url[len(url)-1] == '/' {
+			if url[len(url)-1] == '/' && url !="/" {
 				localRedirect(w, r, "../"+path.Base(url))
 				return
 			}


medcl:
@googlebot I signed it!

PR #38553, opened 2020-04-21
add generics

kennethreitz:
add generics


kennethreitz:
@googlebot add generics

kennethreitz:
Go is already great without generics !

PR #38741, opened 2020-04-29
net/http: document when (RoundTripper).RoundTrip may return a nil *Response

### LINKED ISSUE #37599: net/http: document when (RoundTripper).RoundTrip may return a nil *Response
(Factored out from #37598.)

The `net/http.RoundTripper` interface is currently ambiguous about what happens in case of errors.

Its documentation says:
> ```go
>     // RoundTrip executes a single HTTP transaction, returning
>     // a Response for the provided Request.
> ```

but:
> ```go
>     // … RoundTrip must return err == nil if it obtained
>     // a response, regardless of the response's HTTP status code.
>     // A non-nil err should be reserved for failure to obtain a
>     // response.
> ```

Unfortunately, it does not specify whether “failure to obtain a response” may correspond to a nil `*Response`, or only one that is only partially populated.

CC @bradfitz @rsc

abemotion:
Fixes #37599

### DIFF
--- a/src/net/http/client.go
+++ b/src/net/http/client.go
@@ -122,9 +122,10 @@ type RoundTripper interface {
 	// particular, RoundTrip must return err == nil if it obtained
 	// a response, regardless of the response's HTTP status code.
 	// A non-nil err should be reserved for failure to obtain a
-	// response. Similarly, RoundTrip should not attempt to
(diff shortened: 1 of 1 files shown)


abemotion:
@bradfitz @bcmills I add a commit to this PR if smaller change is better.

PR #38779, opened 2020-04-30
crypto/x509: subject/issuer hint fallback for unknown-authority errors

wking:
Because errors like:

    certificate signed by unknown authority

make it difficult to distinguish between "certificate is unexpected"
and "my local trust store is missing something I expected".  This
commit adds a fallback with summaries for the subject and issuer when
hintErr is missing (e.g. because nothing in the local trust store
matched).  That should also help figure out which of many possible
certificates need fixing when trust-management breaks down.

### DIFF
--- a/src/crypto/x509/verify.go
+++ b/src/crypto/x509/verify.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"iter"
 	"maps"
+	"math/big"
 	"net"
 	"net/netip"
 	"net/url"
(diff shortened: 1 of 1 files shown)


wking:
@heschi, details about why this was closed?  Looks like it still applies, and seems like useful detail as I pointed out in my initial comment here.  Anything I can do to help get some movement in this direction landed?

ianlancetaylor:
Reopening.

CloneOfAlex (approved):

wking:
@seankhliao, do you have more context on why you closed this?  I'm happy to rebase if that makes review easier whenever there's capacity to review :)

ianlancetaylor:
Reopening.

wking:
After Ian re-opened, GitHub did detect a conflct in the `import` section.  I've rebased onto the current dev-tip with 62c3b8c1b2 -> 89b1047e5c.

PR #38817, opened 2020-05-02
net/http: include status code in comments

bnfinet:
By showing the status code in the comment the code will be shown
in various editors as additional information.  As a developer, I often
know that I need to return a specific http code but I'm uncertain of
the semantic mapping that golang is using.  By including the code in
the comment it avoids having to either view the RFC or opening
the status.go source.

the Gerrit change.

  + No trailing period
+ No Markdown


bnfinet:
@googlebot I signed it!

bnfinet:
I'm using VSCode.

TBH I'm not certain this fix really gives me what I want.  As I type I'd like to see what the code (the `const int` value) is as I pick from the list of presented terms.

The current behavior using vscode with [gopls](https://github.com/golang/tools/blob/master/gopls/doc/user.md) is to show...
```
int
```

If I type/complete `http.StatusOK` and then hover with my mouse over I see...
```
const http.StatusOK untyped int = 200
http.StatusOK on pkg.go.dev
```

hovering + ctrl is better (shows the line of code and RFC reference)
```
StatusOK                   = 200 // RFC 7231, 6.3.1
StatusCreated              = 201 // RFC 7231, 6.3.2
const http.StatusOK untyped int = 200
http.StatusOK on pkg.go.dev
```

I'm tempted to offer an updated PR with individual comments for each code..
```
	// StatusOK 200 - RFC 7231, 6.3.1
	StatusOK = 200
```

which offers while I type what I'd like to see...
```
int

StatusOK 200 - RFC 7231, 6.3.1
```

I agree this is certainly not a language issue.  Perhaps this is best fixed in gopls.

bnfinet:
This sed line was used to modify the file (then formated by vscode/gofmt).

`cat status.go | sed -r 's|(.*(Status[[:alpha:]]*)[[:space:]]*= ([[:digit:]]*)) // (.*)|// \2 \3 - \4\n\1\n|'`

PR #38834, opened 2020-05-03
testing: reduce memory allocation in Helper

bboreham:
Store the PC instead of the string name of the function, and defer
that conversion until we need it.

Helper is still relatively expensive in CPU time (few hundred ns),
but memory allocation is now constant for a test rather than linear in
the number of times Helper is called.

benchstat:
name        old time/op    new time/op    delta
TBHelper-4    1.30µs ±27%    0.53µs ± 1%   -59.03%  (p=0.008 n=5+5)

name        old alloc/op   new alloc/op   delta
TBHelper-4      216B ± 0%        0B       -100.00%  (p=0.008 n=5+5)

name        old allocs/op  new allocs/op  delta
TBHelper-4      2.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)


bboreham:
Stems from me discovering that calling `testify/require.NoError()` in the middle of a tight loop caused more memory allocation than the code it was testing.

Some discussion and a benchmark at https://github.com/stretchr/testify/pull/933#issuecomment-620217067

PR #38938, opened 2020-05-07
crypto/rand: use BCryptGenRandom instead of CryptGenRandom on Windows

### LINKED ISSUE #33542: crypto/rand: Currently using deprecated API for random number generation on Windows
N/A

Yes

Any release of Windows supported by Microsoft

### What did you do?
Reviewing cryptographic protocols for a downstream project

Use of a modern Windows API to retrieve entropy, namely BCryptGenRandom from CryptoNG, which is compliant with CTR_DRBG from NIST SP800-90.

### What did you see instead?
Go [calls CryptGenRandom](https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L206) from [a deprecated API](https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptgenrandom). The algorithm is not fully documented and is [only known in part](https://en.wikipedia.org/wiki/CryptGenRandom).

neolit123:
The existing function that is used is CryptGenRandom. This function
and the whole underling API is deprecated.

Use the function BCryptGenRandom from the new recommended
API called "Cryptography API: Next Generation (CNG)".

Fixes #33542

### DIFF
--- a/src/crypto/rand/rand.go
+++ b/src/crypto/rand/rand.go
@@ -14,7 +14,7 @@ import "io"
 // On Linux and FreeBSD, Reader uses getrandom(2) if available, /dev/urandom otherwise.
 // On OpenBSD, Reader uses getentropy(2).
 // On other Unix-like systems, Reader reads from /dev/urandom.
-// On Windows systems, Reader uses the CryptGenRandom API.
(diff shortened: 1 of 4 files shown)


neolit123:
https://github.com/golang/go/issues/33542#issuecomment-715642267

PR #38997, opened 2020-05-11
net/http: remove redundant numeric StatusCode from Status message

### LINKED ISSUE #38996: net/http: Status shouldn't include StatusCode
In parsing: `HTTP/1.1 200 OK`

`StatusCode` is correctly set to `200` but `Status` is parsed to `200 OK` when it should only be `OK`. This leads to inconsistency with `http.StatusText` which only returns `OK` instead of `200 OK`

https://github.com/golang/go/blob/master/src/net/http/response.go#L176
https://github.com/golang/go/blob/master/src/net/http/status.go#L85

kkroo:
In the case of parsing HTTP status with a message, eg: `HTTP/1.1 200 OK`,  `StatusCode` is correctly set to `200` but `Status` is parsed to `200 OK` when it should only be `OK`. This leads to inconsistency with `http.StatusText` which only returns `OK` instead of `200 OK`.

In the case `Status` includes a space, it includes a numeric status code we can trim out the numeral and move it to `StatusCode`

Fixes #38996

### DIFF
--- a/src/net/http/response.go
+++ b/src/net/http/response.go
@@ -174,6 +174,7 @@ func ReadResponse(r *bufio.Reader, req *Request) (*Response, error) {
 	statusCode := resp.Status
 	if i := strings.IndexByte(resp.Status, ' '); i != -1 {
 		statusCode = resp.Status[:i]
+		resp.Status = resp.Status[i:]
 	}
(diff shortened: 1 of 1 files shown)


kkroo:
@googlebot I signed it!

PR #39015, opened 2020-05-12
container/list: check the list state before mutation

### LINKED ISSUE #39014: container/list: zero List value breaks after reset

<pre>
$ go version
go version go1.14.2 darwin/amd64
</pre>

Yes.

$ go env
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
</pre></details>

### What did you do?
```go
var l list.List
el := l.PushBack(42)
l = list.List{} // Reset the list.
l.Remove(el)
fmt.Println(l.Len())
```

```bash
0
```

### What did you see instead?
```bash
-1
```

gobwas:
It is not enough to check if the Element's list pointer is equal to some
particular List pointer to answer whether an Element belongs to that List.
List might be zeroed and then mutated, which makes it completely different,
so that previous elements must not be connected with the new generation of
the list.

Fixes #39014

### DIFF
--- a/src/container/list/list.go
+++ b/src/container/list/list.go
@@ -22,6 +22,8 @@ type Element struct {
 
 	// The list to which this element belongs.
 	list *List
+	// Generation of the list to which this element belongs.
+	gen *byte
 
 	// The value stored with this element.
 	Value interface{}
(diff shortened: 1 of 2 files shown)


src/container/list/list.go
```diff
@@ -22,22 +22,24 @@ type Element struct {
...
+	gen *byte
 
 	// The value stored with this element.
 	Value interface{}
 }
 
 // Next returns the next list element or nil.
 func (e *Element) Next() *Element {
-	if p := e.next; e.list != nil && p != &e.list.root {
+	if p := e.next; e.list.has(e) && p != &e.list.root {
```
lsytj0413, line 34: why drop **e.list != nil**？

PR #39224, opened 2020-05-23
net/http: adds ability to marshall cookiejar to struct

infalmo:


infixint943:
@googlebot I signed it!

PR #39245, opened 2020-05-25
syscall: support O_SYNC flag for os.OpenFile on windows

### LINKED ISSUE #35358: os: O_SYNC not utilized in os.OpenFile() on Windows
While Linux systems are able to pass their low level flags directly into `os.OpenFile` on Windows `os.OpenFile` takes "invented values"

https://github.com/golang/go/blob/5d1a95175e693f5be0bc31ae9e6a7873318925eb/src/syscall/types_windows.go#L34-L48

These invented values are then checked against a subset of the features that Windows actually supports in syscall.Open

https://github.com/golang/go/blob/03aca99f476c34bad927410741251162181b6e16/src/syscall/syscall_windows.go#L273

Despite Windows supporting FILE_FLAG_WRITE_THROUGH (very close to O_SYNC on Linux) the Open function does not check for O_SYNC. This is unexpected behavior as developers would expect the O_SYNC flag on Windows to work since it can perform synchronous writes.

Some ways to resolve this unexpected behavior include:
- Adding a check for O_SYNC in syscall.Open in syscall_windows.go
- Having syscall.Open perform on Windows the same way it does on Linux, by passing in file flags directly
   - Note: This solution would also help with https://github.com/golang/go/issues/34681

jarifibrahim:
The current implementation of os.OpenFile function for windows does
not use the O_SYNC flag. This means that even if the user has set the
O_SYNC flag, the os.OpenFile call will ignore the SYNC flag. This
change adds _FILE_FLAG_WRITE_THROUGH, the equivalent of
O_SYNC flag on Linux.

Fixes #35358.


jarifibrahim:
I couldn't find any of the existing tests and I don't know how to write a test for a syscall. I'm looking for help in writing the tests for this PR.

src/syscall/syscall_windows.go
```diff
@@ -360,6 +360,11 @@ func Open(path string, mode int, perm uint32) (fd Handle, err error) {
 			}
 		}
 	}
+
+	if mode&O_SYNC != 0 {
+		attrsAndFlags != FILE_FLAG_WRITE_THROUGH
```
BrianLeishman, line 365: I don't believe this line actually does anything, just checks if those two values are not equal, returning a bool that isn't being captured

jarifibrahim: Thank you @BrianLeishman. It was a silly typo. I've fixed it :+1:

Galecoco:
能不能退订？？？

GopherBot ***@***.***> 于 2021年4月22日周四 07:53写道：

PR #39338, opened 2020-06-01
crypto/tls: re-order cipher suites

xunzhaocnm:
TLS1.3 suites should be listed on top in client hello.


odeke-em:
Please sign the CLA then report here back, but also please provide the motivation/citation for why they should be listed at the front and even perhaps add a test to ensure that we don't regress.

mvdan:
I'm closing this old PR, since it's been a while and the CLA wasn't signed. Feel free to submit another PR or Gerrit CL in the future, but please do sign the CLA if so.

PR #39416, opened 2020-06-05
mime: look up mime types on Arch Linux

andydotxyz:
Some systems use "httpd" directory structure instead of "apache"


andydotxyz:
https://www.archlinux.org/packages/extra/x86_64/apache/files/

PR #39418, opened 2020-06-05
Add option to Limit maximum number of go routines launched from serve()

redhug1:
In Container with limited amount of RAM, heavy HTTP load causes to many go routines to be created and the call back handler routines some time later do their memory allocations that cause container to be killed.
This simple extension whose use is optional allows the setting of maximum go routines launched from serve().
It is up to the programmer to determine what this limit should be through load testing their web site.
Printing out information from calling runtime.ReadMemStats() at a reduced rate to a log file for plotting and analysis
can help to determine this limit.

the Gerrit change.

  + No trailing period
+ No Markdown


redhug1:
@googlebot I signed it!

redhug1:
Please see PDF's in:
  https://github.com/redhug1/go-ListenAndServeLimits
that show the effectiveness of limiting the number of go routines on the memory allocations.

PR #39494, opened 2020-06-10
fix stderr to stdout

brettKK:
fix stderr to stdout in pprof.go


brettKK:
@googlebot I signed it!

PR #39556, opened 2020-06-12
encoding/csv: allow for interpretation of empty lines

### LINKED ISSUE #39119: encoding/csv: skipping of empty rows leads to loss of data in single-column datasets

<pre>
$ go version
go version go1.14 darwin/amd64
</pre>

Yes.

$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/ondrej/Library/Caches/go-build"
GOENV="/Users/ondrej/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/ondrej/go"
GOPRIVATE=""
GOROOT="/usr/local/Cellar/go/1.14/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.14/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/hp/q7nph21s1q76nw1hv1hfxv2m0000gn/T/go-build988616937=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre></details>

### What did you do?

I wrote a CSV with a single column and missing data.

I expected to load the data back, intact.

### What did you see instead?

I lost the missing values, `encoding/csv` skipped them as it skips blank lines. In this case, a blank line actually represents data.

---

I'm not sure I understand the rationale behind skipping blank lines. Neither in terms of common practice (why would I have blank lines in my CSVs?) nor in terms of standards (the closest we have is RFC 4180 and I couldn't find anything about blank lines - so I'm not sure if Go [follows it](https://github.com/golang/go/commit/fa3f484800415662cc741bbb8968ebb72896e20a)).

Here's a reproduction of the problem. I wrote a dataset into a file and was unable to read it back.

<details>
<pre>
package main

import (
	"encoding/csv"
	"errors"
	"log"
	"os"
	"reflect"
)

func writeData(filename string, data [][]string) error {
	f, err := os.Create(filename)
	if err != nil {
		return err
	}
	defer f.Close()
	cw := csv.NewWriter(f)
	defer cw.Flush()
	if err := cw.WriteAll(data); err != nil {
		return err
	}
	return nil
}

func readData(filename string) ([][]string, error) {
	f, err := os.Open(filename)
	if err != nil {
		return nil, err
	}
	defer f.Close()
	cr := csv.NewReader(f)
	rows, err := cr.ReadAll()
	if err != nil {
		return nil, err
	}
	return rows, nil
}

func run() error {
	fn := "data/roundtrip.csv"
	data := [][]string{{"john"}, {"jane"}, {""}, {"jack"}}

	if err := writeData(fn, data); err != nil {
		return err
	}

	returned, err := readData(fn)
	if err != nil {
		return err
	}
	if !reflect.DeepEqual(returned, data) {
		log.Println("expected", data, "got", returned)
		return errors.New("not equal")
	}

	return nil
}

func main() {
	if err := run(); err != nil {
		log.Fatal(err)
	}
}

</pre>
</details>

kokes:
Empty lines are ignored by default, this change introduces a switch that
changes this behaviour - empty line is then interpreted as a single empty
field. This is especially important when dealing with single-column CSVs,
where an empty line is actually data (an empty field).

Fixes #39119.

### DIFF
--- a/src/encoding/csv/reader.go
+++ b/src/encoding/csv/reader.go
@@ -16,8 +16,8 @@
 //
 // Carriage returns before newline characters are silently removed.
 //
-// Blank lines are ignored. A line with only whitespace characters (excluding
-// the ending newline character) is not considered a blank line.
+// Blank lines are ignored by default. A line with only whitespace characters
+// (excluding the ending newline character) is not considered a blank line.
 //
 // Fields which start and stop with the quote character " are called
 // quoted-fields. The beginning and ending quote are not part of the
@@ -141,6 +141,10 @@ type Reader struct {
 	// the backing array of the previous call's returned slice for performance.
 	// By default, each call to Read returns newly allocated memory owned by the caller.
 	ReuseRecord bool
+
+	// If InterpretBlankLines is true, blank lines are interpreted as a single empty field,
+	// the default is to skip these lines.
+	InterpretBlankLines bool
 
 	TrailingComma bool // Deprecated: No longer used.
 
@@ -268,7 +272,7 @@ func (r *Reader) readRecord(dst []string) ([]string, error) {
 			line = nil
 			continue // Skip comment lines
 		}
-		if errRead == nil && len(line) == lengthNL(line) {
+		if !r.InterpretBlankLines && errRead == nil && len(line) == lengthNL(line) {
 			line = nil
 			continue // Skip empty lines
 		}
--- a/src/encoding/csv/reader_test.go
+++ b/src/encoding/csv/reader_test.go
@@ -20,13 +20,14 @@ func TestRead(t *testing.T) {
 		Error  error
 
 		// These fields are copied into the Reader
-		Comma              rune
-		Comment            rune
-		UseFieldsPerRecord bool // false (default) means FieldsPerRecord is -1
-		FieldsPerRecord    int
-		LazyQuotes         bool
-		TrimLeadingSpace   bool
-		ReuseRecord        bool
+		Comma               rune
+		Comment             rune
+		UseFieldsPerRecord  bool // false (default) means FieldsPerRecord is -1
+		FieldsPerRecord     int
+		LazyQuotes          bool
+		TrimLeadingSpace    bool
+		ReuseRecord         bool
+		InterpretBlankLines bool
 	}{{
 		Name:   "Simple",
 		Input:  "a,b,c\n",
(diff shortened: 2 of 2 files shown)


kokes:
@googlebot I signed it!

src/encoding/csv/reader_test.go
```diff
@@ -78,6 +79,25 @@ field"`,
 			{"a", "b", "c"},
 			{"d", "e", "f"},
 		},
+	}, {
+		Name:  "BlankLineInterpreted",
+		Input: "a,b,c\n\nd,e,f\n\n",
```
sandan, line 84: Doesn't this test violate FieldsPerRecord?

>  // If FieldsPerRecord is 0, Read sets it to
    // the number of fields in the first record, so that future records must
    // have the same field count.

sandan (commented):
I agree that interpreting empty lines is important for single field cases. The current implementation requires a fixed number of fields in some cases though. So in those cases, an error should be thrown when an empty line is seen but the "header row/first row" has more than one field or FieldsPerRecord is greater than 1.

src/encoding/csv/reader.go
```diff
@@ -268,7 +272,7 @@ func (r *Reader) readRecord(dst []string) ([]string, error) {
 			line = nil
 			continue // Skip comment lines
 		}
-		if errRead == nil && len(line) == lengthNL(line) {
+		if !r.InterpretBlankLines && errRead == nil && len(line) == lengthNL(line) {
```
sandan, line 275: I think the comment should be updated above where it says:
```
	// Read line (automatically skipping past empty lines and any comments).
```
Should just be:
```
	// Read line
```

kdumontnu:
Is this PR still active? I'm running into this issue currently in a project where I need to map the csv line number to the actual line number in a file. This PR looks like it would resolve my issue as is. 

I have multiple columns, but simply need the csv decode to retain an empty field for the empty row.

kdumontnu:
See: https://github.com/golang/go/pull/45228

@kokes If you are still available to work on this PR and would like to take it over the finish line please feel free to message me and I will close the new PR.

kokes:
@kdumontnu it's not really an issue of my availability - I submitted a PR that solves the issue as I see it and retains the current behaviour as default, but it was rejected by Ian from the Go team, so there's not much I can do about that - see the issue I opened originally.

> We've pretty much decided that we are not going to add additional knobs to encoding/csv. There are just too many variations in CSV files out there. Trying to handle all of them would lead to a massive proliferation of knobs and an unusable package API.

I haven't reviewed your PR fully, but from the look of it, it will not pass either.

I solved it for myself by forking `encoding/csv`, not much else I can do about this.

kdumontnu:
Ah, I see - thanks for the reply. That's frustrating... You are correct - my PR is just a continuation of yours with an updated comment.

PR #39771, opened 2020-06-23
time: add widely used time.Duration Day

leaxoy:
This PR add new time constant `time.Day`, which is `time.Duration`. Duration `Day` is widely used like pre-defined `Hour`/`Minute`/`Second`/`Millisecond` etc, but is not pre-defined in `time` package. So I add it.


leaxoy:
@googlebot I signed it!

leaxoy:
Closed

Issue #38737, opened 2020-04-29
Marshal ASN.1 OBJECT IDENTIFIER then unmarshal causes unmarshal error


sebastien-rosset:
### What version of Go are you using (`go version`)?

go version go1.14.2 linux/amd64

### Does this issue reproduce with the latest release?

Yes

### What operating system and processor architecture are you using (`go env`)?

```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/serosset/.cache/go-build"
GOENV="/home/serosset/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/serosset/goroot"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build216516397=/tmp/go-build -gno-record-gcc-switches"
```
### What did you do?

Link to program:
https://play.golang.org/p/tI2WDW2A8Wm

### What did you expect to see?

1. Encode Object Identifier "1.2.36.67006527840.666.66.1.1" in DER format using the asn1.Marshal function.
1. Verify manually the byte encoding is correct
1. Take the ASN.1 encoded object identifier and try to unmarshal. Unmarshal should succeed and the same original object identifier should be retrieved.

### What did you see instead?

Step (4) is failing. asn1.Unmarshal fails with error **asn1: structure error: base 128 integer too large**

odeke-em:
Thank you for this report @sebastien-rosset!

Kindly pinging the crypto squad @katiehockman @FiloSottile

rolandshoemaker:
[...]

PR #39795, opened 2020-06-24
encoding/asn1: support ObjectIdentifier sub-oid values greater than 2^31-1

sebastien-rosset:
This PR adds support for ASN.1 object identifiers where at least one sub-oid has a value greater than 2^31-1. As per ASN.1 specification, each sub-oid is an ASN.1 INTEGER, which can be of arbitrary size. For example, the value of a sub-oid could be `2 << 80`. In practice, scenarios with large sub-oids are rare but they do exist.

One problem is that historically, the `ObjectIdentifier` type has been defined as `[]int`; this means the max value of sub-oids is math.MaxInt32. The problem leaks in multiple go packages, such as `crypto.x509`. This is a problem because some real-world X.509 certificates have sub-oids values greater than 2^31.

```go
type Certificate struct {
	...
	// Policy identifiers. This can only contains sub-oids whose value is less than math.MaxInt32
	PolicyIdentifiers []asn1.ObjectIdentifier
}
```

This is an initial proposal using option 3 below. I should add more unit tests, run some benchmarks and optimize code, but before that it would great to know if the approach is going in the right direction.

Besides fixing the ASN.1 compliance issue for the ASN.1 unmarshaler, another issue is how to parse X.509 certificates when policy identifiers have sub-oids larger than math.MaxInt32. The fix in the x.509 package is dependent on the asn1 fix. I am including the x509 fix, let me know if it makes more sense to have two separate PRs. 

### Option 1
Redefine `asn1.ObjectIdentifier` as a `[]int64` or even `[]*big.Int`. This is probably a non-starter because a lot of software has been built with this assumption. It's not possible to change the type to  `[]int64` or  `[]*big.Int` without breaking existing software. 
```go
type ObjectIdentifier []int64
```
or
```go
type ObjectIdentifier []*big.Int
```

### Option 2
Add two new types:
```golang
type ObjectIdentifierInt64  []int64
type ObjectIdentifierBigInt []*big.Int
```
If all sub-oids are less than math.MaxInt32, then the return type could still be `ObjectIdentifier` to retain backward compatibility. If at least one sub-oid is more than math.MaxInt32, then the return value is ObjectIdentifierInt64. If at least one sub-oid is more than math.MaxInt64, then the return value is ObjectIdentifierBigInt.

One problem with this approach is it's cumbersome for consumers (such as `crypto.x509`) to deal with three different types.

### Option 3
Add a new type that can take int, int64 or *big.Int as sub-oid values:
```go
type ObjectIdentifierExt []interface{}
```

In particular, this could be used in the `x509` package to support certificates that have large sub-oid values. The existing `PolicyIdentifiers` field cannot be removed without breaking the backward compatibility. It's doable but somewhat cumbersome to have two possible PolicyIdentifiers fields. 

```go
type Certificate struct {
	...
	// Policy identifiers. This can only contains sub-oids whose value is less than math.MaxInt32
	PolicyIdentifiers []asn1.ObjectIdentifier
        // Policy identifiers that contain sub-oid values greater than math.MaxInt32
	PolicyIdentifiersExt []asn1.ObjectIdentifierExt
}
```

### Option 4
Use go generics:

```go
type SignedInteger interface {
	type int, int32, int64, *big.Int
} 
type ObjectIdentifier(type T SignedInteger) []T
```

Developers would be able to instantiate object identifiers with different sub-types:
```go
var oid1 ObjectIdentifier(int) = ObjectIdentifier(int){1, 2, 3}
var oid1 ObjectIdentifier(int64) = ObjectIdentifier(int64){1, 2, 1 << 60}
```

`x509.Certificate` could also be a generic type. Developers would have the choice of supporting X.509 certificates with policy identifiers that have large sub-oids:

```go
type Certificate(type T SignedInteger) struct {
	...
	// Policy identifiers. This can only contains sub-oids whose value is less than math.MaxInt32
	PolicyIdentifiers []asn1.ObjectIdentifier(T)
}
```

However, this generic approach does not work for existing code because that would result in compilation failure. For example, the code below causes "cannot use generic type ObjectIdentifier(type T SignedInteger) without instantiation"
```go
var oid1 ObjectIdentifier = ObjectIdentifier{1, 2, 3}
```

To get around this problem, I was wondering if a default instantiation type has been considered when in the go generic specification, something like below. This would make it possible to retrofit existing code while providing extended use cases.
```go
type ObjectIdentifier(type T SignedInteger=int)  // similar to C++ default type notation
```

Or if somehow it were possible to have one generic version and one non-generic version:
```go
type ObjectIdentifier(type T SignedInteger) []T
type ObjectIdentifier []int
```

One little annoyance is that even with generics, it's not possible to use common arithmetic operators (+, *, /) that work across int, int64 and *big.Int.

But even if there is a way to use generics without breaking code, it would seem odd to be required to expose the sub-oid generic type to all types (e.g. `x509.Certificate`) that use ObjectIdentifier.

### Current use of asn1.ObjectIdentifier type

Below are the publicly exposed asn1.ObjectIdentifier fields in the golang/go repo:

1. encoding/asn1
1. crypto/x509/pkix package has 4 exported fields using asn1.ObjectIdentifier
1. crypto/x509 package:
    1. ECDSA keys
    1. Used in the x509.Certificate struct:

```go
// A Certificate represents an X.509 certificate.
type Certificate struct {
  ...
  UnhandledCriticalExtensions []asn1.ObjectIdentifier
  UnknownExtKeyUsage []asn1.ObjectIdentifier
  PolicyIdentifiers []asn1.ObjectIdentifier
}
```

For #38737


sebastien-rosset:
@googlebot I signed it!

zzo38:
I think it would be better (although I don't know if it can easily be changed; it is not a programming language that I use) to store a OID as the string of bytes that it is encoded as, and use that for comparison. Decoding it into the sequence of numbers is only necessary for display or conversion, not for comparison (if you need to compare a dotted decimal OID with a binary OID, it is better to encode the dotted decimal OID to binary instead of the other way around).

Another alternative (which might avoid needing to change the types) might be to use negative numbers to represent OIDs that cannot be represented with the existing implementation; one possible way to do this would be, if a part of the OID cannot be decoded, use the bitwise complement of each individual byte for the rest of the OID. This has the disavantage that OIDs that can be decoded in the usual way have an alternate representation which encodes into the same thing, and this might be undesirable.

PR #39825, opened 2020-06-24
runtime: remove tracebackinit and unused skipPC

liu-xuewen:
The related commit:  https://github.com/golang/go/commit/69c2c56453cdea1ad32d50fc82193f06b1b33f93


liu-xuewen:
@googlebot I signed it!

PR #39844, opened 2020-06-25
cmd/internal/objfile: cache goobj.Arch for PCToLine

qingyunha:
Implement an unresolved TODO to cache goobj.Arch.


qingyunha:
https://github.com/golang/go/issues/40874

PR #39870, opened 2020-06-26
crypto/rsa: add rand initialization for rsa.SignPSS

shibumi:
If nil as random source is being passed to rsa.SignPSS
this is going to lead to a nil pointer dereference and invalid memory
access. This commit intents to this fix via initializing
a secure random source with crypto/rand.Reader

### DIFF
--- a/src/crypto/rsa/pss.go
+++ b/src/crypto/rsa/pss.go
@@ -260,8 +260,14 @@ func (opts *PSSOptions) saltLength() int {
 //
 // digest must be the result of hashing the input message using the given hash
 // function. The opts argument may be nil, in which case sensible defaults are
-// used. If opts.Hash is set, it overrides hash.
+// used. If opts.Hash is set, it overrides hash. The rand argument may be nil
+// if nil rand will get initialized via crypto/rand.Reader
 func SignPSS(rand io.Reader, priv *PrivateKey, hash crypto.Hash, digest []byte, opts *PSSOptions) ([]byte, error) {
+	// if no random source has been passed
+	// initialize with secure random from crypto/rand.Reader
+	if rand == nil {
+		rand = rand.Reader
+	}
 	if opts != nil && opts.Hash != 0 {
 		hash = opts.Hash
 	}


shibumi:
Hi,
This is my first PR for Go, so please look twice over this PR.

shibumi:
The nil pointer dereference happens here: https://github.com/golang/go/blob/c2406259692427ade0be638e02e8d42a926d41d9/src/crypto/rsa/pss.go#L283

A possible stacktrace for it looks like this:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x49e5e8]

goroutine 19 [running]:
testing.tRunner.func1.1(0x5f4660, 0x7be6c0)
	/usr/lib/go/src/testing/testing.go:940 +0x2f5
testing.tRunner.func1(0xc00014e240)
	/usr/lib/go/src/testing/testing.go:943 +0x3f9
panic(0x5f4660, 0x7be6c0)
	/usr/lib/go/src/runtime/panic.go:969 +0x166
io.ReadAtLeast(0x0, 0x0, 0xc00014c640, 0x20, 0x20, 0x20, 0x736c616972657461, 0x616e222c7d7b3a22, 0x6f6f66223a22656d)
	/usr/lib/go/src/io/io.go:310 +0x58
io.ReadFull(...)
	/usr/lib/go/src/io/io.go:329
crypto/rsa.SignPSS(0x0, 0x0, 0xc000132fc0, 0x5, 0xc00014c620, 0x20, 0x20, 0xc00017dde0, 0x44fd58, 0xc000182228, ...)
	/usr/lib/go/src/crypto/rsa/pss.go:263 +0x105
```

shibumi:
@heschi can you provide some information why this has been closed? I still think that throwing panics is bad practice and initializing the random as fallback is a good way.

heschi:
I closed old PRs to reduce load on the Gerrit importer (#50197), sorry for the trouble. I'll reopen the CL and PR.

shibumi:
@heschi no worries :) @FiloSottile I don't know if discussions around this should happen in gerrit or github, but can I get an answer on my latest comment or a final decision? I am not sad, if you reject this commit. I would just like to know if I should invest more time into this.

PR #39952, opened 2020-06-30
runtime: use local version of strings.Index and strings.IndexByte.

veezhang:


randall77:
Why?

veezhang:
@randall77 Hi
When looking at the source code, I found that `s [i] == t [0]` has been compared in the runtime.index function. So my first version like tihs:

```go
func index(s, t string) int {
	if len(t) == 0 {
		return 0
	}
	for i := 0; i < len(s); i++ {
		// compare the next character
		if s[i] == t[0] && hasPrefix(s[i+1:], t[1:]) {
			return i
		}
	}
	return -1
}
```

Then, I realized that there can add a case for byte-sized values. I remember, there is such an implementation in strings pacake.

So this came.

randall77:
Why is `indexbyte` just an indirection of `bytealg.IndexByteString`? That doesn't seem to help anything.

`index` is only used by plan9 signal handler, experiment parsing, debug variable parsing, and traceback printing (via `contains`). All of these things execute approximately never, and certainly never in a performance-sensitive path. In such cases, I'd rather have the small, easy to understand implementation.

So I don't think we need this extra complication. Show us a profile where these functions show up, and we can reopen.

PR #39989, opened 2020-07-02
io: add ErrBadWriteCount

### LINKED ISSUE #39978: io: Copy reports ErrShortWrite for long writes

<pre>
$ go version
go version devel +c875503cf7 Sun Jun 28 03:14:10 2020 +0000 darwin/amd64
</pre>

Yes

$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOOS="darwin"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
</pre></details>

### What did you do?

I implemented a buggy io.Writer which sometimes writes too many bytes (i.e., more that the length of the []byte parameter to the Write method), and used io.Copy to write to it.

https://play.golang.org/p/UeO1GCEvmNP

A reasonable error message (indicating that the wrong number of bytes were written, or too many bytes were written)

### What did you see instead?

ErrShortWrite.

Based on the documented meaning of this error variable, I believe io.Copy is using it incorrectly in this case.
> ErrShortWrite means that a write accepted fewer bytes than requested but failed to return an explicit error.

For what it's worth, this came up while debugging real code: the Writer in question buffers writes and was buggy.

qingyunha:
Fixes #39978

### DIFF
--- a/src/io/io.go
+++ b/src/io/io.go
@@ -30,6 +30,9 @@ var ErrShortWrite = errors.New("short write")
 // ErrShortBuffer means that a read required a longer buffer than was provided.
 var ErrShortBuffer = errors.New("short buffer")
 
+// ErrBadWriteCount means that a Write returned impossible count
+var ErrBadWriteCount = errors.New("Write returned impossible count")
+
 // EOF is the error returned by Read when no more input is available.
 // Functions should return EOF only to signal a graceful end of input.
 // If the EOF occurs unexpectedly in a structured data stream,
@@ -414,6 +417,10 @@ func copyBuffer(dst Writer, src Reader, buf []byte) (written int64, err error) {
 			}
 			if ew != nil {
 				err = ew
+				break
+			}
+			if nw < 0 || nw > nr {
+				err = ErrBadWriteCount
 				break
 			}
 			if nr != nw {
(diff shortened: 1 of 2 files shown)


src/io/io.go
```diff
@@ -416,6 +419,10 @@ func copyBuffer(dst Writer, src Reader, buf []byte) (written int64, err error) {
 				err = ew
 				break
 			}
+			if nw < 0 || nw > nr {
+				err = ErrBadWriteCount
+				break
+			}
```
funcnet, line 425: This change is not necessary:
1. `nw < 0` is covered by `if ew != nil ` above
2. `nw > nr` is covered by `nr != nw` below. 
3. this new code causes the existing `if nr != nw` logic not reachable _sometimes_
4. why the code do have contain `nw < nr`?

PR #40054, opened 2020-07-05
[dev.go2go] fix url of draft design

levonet:
Now with the link "design project draft" page with error `NOT_FOUND: Requested entity was not found.` opens.


levonet:
@googlebot I signed it!

PR #40073, opened 2020-07-06
misc/wasm: fix the empty global fs check

alanshaw:
`Object.keys` returns an array so can never equal zero. I believe this check is
looking for a stubbed `fs` object with no keys. I'm using vuepress and a stub
object like this is being added to the `window` object.


alanshaw:
@googlebot I signed it!

PR #40112, opened 2020-07-07
plugin: Adds Symbols function

arrase:


arrase:
@googlebot I signed it!

Issue #30037, opened 2019-01-31
x/build/cmd/gerritbot: not possible to make cherry-pick CLs via PRs


dmitshur:
I investigated whether it's possible to create cherry-pick CLs by making a Pull Request, a question that was brought up on the golang-dev mailing list [here](https://groups.google.com/d/msg/golang-dev/Tc8OlLRsju0/AysH6zdeAwAJ).

In order to learn whether it's supported, I made an attempt and learned from the experience:

- Backport issue - https://github.com/golang/go/issues/29922#issuecomment-457381644
- Cherry-pick PR I created - PR #29926
- Cherry-pick CL that gerritbot imported - [CL 159478](https://golang.org/cl/159478)

It was possible to create the cherry-pick CL against the correct release branch by having the appropriate base branch in the PR.

However, gerritbot does not recognize it when the PR description contains Change-Id and other lines from the commit being backported. In the PR, the description ended with:

```
<body>

Change-Id: I09a2212eb74c63db575223277aec363c55421ed8
Reviewed-on: https://go-review.googlesource.com/c/159157
Run-TryBot: Brad Fitzpatrick <bradfitz@golang.org>
TryBot-Result: Gobot Gobot <gobot@golang.org>
Reviewed-by: Filippo Valsorda <filippo@golang.org>
```

In the imported CL, the commit message ended up having:

```
<body>

Change-Id: I09a2212eb74c63db575223277aec363c55421ed8
Reviewed-on: https://go-review.googlesource.com/c/159157
Run-TryBot: Brad Fitzpatrick <bradfitz@golang.org>
TryBot-Result: Gobot Gobot <gobot@golang.org>
Reviewed-by: Filippo Valsorda <filippo@golang.org>

Change-Id: I09a2212eb74c63db575223277aec363c55421ed8
GitHub-Last-Rev: 396cc6b3a61292ccbe3ccad4e55c3a9842145cab
GitHub-Pull-Request: golang/go#29926
```

In order for it to be possible to make cherry-pick CLs via gerritbot, it would need to gain support for recognizing when the PR description contains `Change-Id` lines from the original commit, and reuse it. So the final commit message imported by gerritbot would need to look something like:

```
<body>

Change-Id: I09a2212eb74c63db575223277aec363c55421ed8
Reviewed-on: https://go-review.googlesource.com/c/159157
Run-TryBot: Brad Fitzpatrick <bradfitz@golang.org>
TryBot-Result: Gobot Gobot <gobot@golang.org>
[...]

PR #40190, opened 2020-07-13
all: modify VERSION file

dmitshur:
DO NOT REVIEW
DO NOT SUBMIT

This is a test.


dmitshur:
This was a test for #30037. It's no longer needed.

PR #40219, opened 2020-07-14
bufio: SplitAt added

xaoirse:
bufio: SplitAt added

A function that take a spliter and return each 
spliter-separated part of text inserted.


xaoirse:
Using strings.split() is a simple way for this

PR #40271, opened 2020-07-17
net/http: fixes off-by-one error in http2getDataBufferChunk

keyneston:
Previously it was doing `i < len(...)-1`, because it is a 'less than' instead of a 'less than or equal to' the -1 ends up being redundant. This in turn causes the largest buffer size to not be reused.


keyneston:
Went to write a go playground to demonstrate the issue, and realised there are some some non-obvious reasons that this does work as intended. Apologies for the false PR.

PR #40293, opened 2020-07-19
time: support fractional timezone minutes in MarshalBinary

### LINKED ISSUE #39616: time: MarshalBinary fails on valid Time

<pre>
$ go version
go version go1.14.2 darwin/amd64
</pre>

Yes.

$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/path/Library/Caches/go-build"
GOENV="/Users/path/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOOS="darwin"
GOPATH="/Users/path/code/go"
GOROOT="/usr/local/opt/go/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/opt/go/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/bf/vxlr89hd4jqfx43ncc0y10b00000gn/T/go-build493699043=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre></details>

### What did you do?

Call `In` on a zero value `time.Time` creates in valid time when the result would be before the zero time.

https://play.golang.org/p/BJQVBcW9Rcj

The zero value time would have its time zone set to the specified timezone yet still be the zero value date.

For Example: 
Setting `0001-01-01 00:00 UTC` to `EDT` should return `0001-01-01 00:00 EDT` or return an error if the value is outside the range (which in this case it is)

Using `zeroTime.After(newTime)` should return `true` or some other mechanism to check the time is valid should exist.

### What did you see instead?
A time value before the range was returned with an invalid time zone `0000-12-31 19:03:58 -0456 LMT`.

Attempting to compare the result against the `zero` time using `After` does not function as `zeroTime.After(newTime)` and `newTime.After(zeroTime)` both return false so you can have a false positive that the time is valid.

Also the resulting time value prevents `MarshalBinary` from working

howjmay:
If the time is in 'LMT' and has fractional minute, then
`MarshalBinary()` and `UnmarshalBinary()` will encode/decode the time
in `timeBinaryVersionV2` in which the fractional minute is at
bit 15 and 16, and presented in seconds.

Fixes #39616

### DIFF
--- a/src/time/time.go
+++ b/src/time/time.go
@@ -1163,19 +1163,26 @@ func (t Time) UnixNano() int64 {
 	return (t.unixSec())*1e9 + int64(t.nsec())
 }
 
-const timeBinaryVersion byte = 1
+const (
+	timeBinaryVersionV1 byte = iota + 1 // For general situation
+	timeBinaryVersionV2                 // For LMT only
+)
 
 // MarshalBinary implements the encoding.BinaryMarshaler interface.
 func (t Time) MarshalBinary() ([]byte, error) {
 	var offsetMin int16 // minutes east of UTC. -1 is UTC.
+	var offsetSec int8
+	version := timeBinaryVersionV1
 
 	if t.Location() == UTC {
 		offsetMin = -1
 	} else {
 		_, offset := t.Zone()
 		if offset%60 != 0 {
-			return nil, errors.New("Time.MarshalBinary: zone offset has fractional minute")
+			version = timeBinaryVersionV2
+			offsetSec = int8(offset % 60)
 		}
+
 		offset /= 60
 		if offset < -32768 || offset == -1 || offset > 32767 {
 			return nil, errors.New("Time.MarshalBinary: unexpected zone offset")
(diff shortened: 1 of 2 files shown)


HowJMay:
@rsc is there anything that I should add or fix in this pr?

PR #40375, opened 2020-07-23
Add TryLock method to Mutex

madeken:
TryLock is a really useful primitive.  A very common pattern for example is something like:

```
func cleanup() {
   mutex.Lock()
   defer mutex.Unlock();


    // thread unsafe cleanup code
}
```

But it suffers from a pretty typical issue where a bunch of them could queue up. A try lock would allow someone to write the code to abort immediately if the code isn't in progress

### DIFF
--- a/src/sync/mutex.go
+++ b/src/sync/mutex.go
@@ -170,6 +170,18 @@ func (m *Mutex) lockSlow() {
 	}
 }
 
+// Tries to lock m without blocking.
+// Returns if it locked the mutex or not
+func (m *Mutex) TryLock() bool {
+	if atomic.CompareAndSwapInt32(&m.state, 0, mutexLocked) {
+		if race.Enabled {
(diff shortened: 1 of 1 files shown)


madeken:
@googlebot I signed it!

danneu:
We use this heavily, except we named it `LockIfUnlocked()` which makes its behavior obvious.

madeken:
Yeah, I know Rob Pike didn't like the name `TryLock` either and suggested `CanLock` or `GetLock`, but I think `TryLock` has the benefit of being pretty standard in other languages.  I actually do like the name `LockIfUnlocked` although it's a bit verbose, if there's consensus that it's a better name

neilalexander:
It appears something like this was merged already in 645d07819b2224ba4d759829443f7c6442162c69.

PR #40384, opened 2020-07-24
database/sql: add on-connect hook support

lukaszgryglicki:
Hi, I know this PR will *not* be merged, but I wanted to propose some kind of solution giving users an ability to execute some queries every time a new connection is used.

Here is why I need this: I want to set a session-level variable, DB doesn't matter (I've tried on MySQL/MariaDB). I've googled a lot and wasn't able to find any way to do that correctly. `database/sql` maintains its own pool of connections and it is not enough to set this once on DB connect, I also don't want to prepend this to every SQL call, I just want to have a way to declare some kind of callback that happens every time when `database/sql` is creating a new connection (and not reusing the existing one).

This is quite close to what I want (just need to define a handler for `state=2`) but I can't use `DB` object there because DB != connection, and I cannot make `driverConn` available to the outside, so I can only pass a list of queries and their args to be executed in an event of new connection create.

I'm new to golang code, so this PR will be rejected 100%, but can you point me in the right direction, please?

### DIFF
--- a/src/database/sql/sql.go
+++ b/src/database/sql/sql.go
@@ -387,6 +387,13 @@ type Out struct {
 // defers this error until a Scan.
 var ErrNoRows = errors.New("sql: no rows in result set")
 
+// Can be use to specify on-connect handler params
+type OnConnParams struct {
+	Enabled [3]bool
+	SQLs    []string
+	Args    [][]interface{}
+}
+
 // DB is a database handle representing a pool of zero or more
 // underlying connections. It's safe for concurrent use by multiple
 // goroutines.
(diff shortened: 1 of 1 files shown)


lukaszgryglicki:
To actually use it, do something like this:
```
  origin := "my-app"
  d.SetOnConn(
    &sql.OnConnParams{
      Enabled: [3]bool{false, false, true},
      SQLs:    []string{"set @origin = ?"},
      Args:    [][]interface{}{{origin}},
    },
  )
```

lukaszgryglicki:
@googlebot I signed it!

PR #40392, opened 2020-07-24
runtime: reduce syscall when call runtime.clone

cuiweixie:


cuiweixie:
@googlebot I signed it!

PR #40428, opened 2020-07-27
strings: improve performance of ToUpper and ToLower

suzaku:
ASCII-only strings are treated separately in these two functions,
instead of using a pre-allocated strings.Builder we might just as well
use a byte slice.

In the benchmarks I run on my environment, using []byte directly is
generally faster than using a strings.Builder.


suzaku:
@googlebot I signed it!

GiFarina (approved):

PR #40662, opened 2020-08-10
math: add Tau in constants

lypwig:
Add Tau in mathematical constants.

In many cases, formulas can be more elegant using the Tau constant instead
of Pi. You will find a good number of arguments and explanations
[in the Tau manifesto](https://tauday.com/tau-manifesto), which describes Tau as:

> the most important numbers in mathematics, perhaps the most important:
> the circle constant relating the circumference of a circle to its linear
> dimension. For millennia, the circle has been considered the most
> perfect of shapes, and the circle constant captures the geometry of the
> circle in a single number.

The constant is based on [the corresponding page in the online
encyclopedia of integer sequences](https://oeis.org/A019692), and is also
equal to `math.Pi * 2`.

Usage:

```go
perimeter := math.Tau * radius
```


Goutte:
Thanks so much for this !

I've been missing it since the beginning, but I was too lazy to do it myself.

As I stated before, TAU is the circle constant, and PI the disk constant.

> When civilizations make mistakes, the inertia is heavy.

ALTree:
The proposal was rejected (#40663) so I'm closing this PR.

PR #40678, opened 2020-08-10
doc/mem: reword happens-rules in channel communication docs

hanyucui:
The concept "happen" is inadequate to describe send and receive in the doc.
"Happen" usually describes zero-duration events while the duration of send
and receive is meaningful both in the doc and in practice. I replace
"happen" with "start", which clarifies which event we are talking about and
matches the use of "complete".

### DIFF
--- a/doc/go_mem.html
+++ b/doc/go_mem.html
@@ -213,7 +213,7 @@ usually in a different goroutine.
 </p>
 
 <p class="rule">
-A send on a channel happens before the corresponding
+A send on a channel starts before the corresponding
 receive from that channel completes.
 </p>
 
@@ -256,7 +256,7 @@ yields a program with the same guaranteed behavior.
 </p>
 
 <p class="rule">
-A receive from an unbuffered channel happens before
+A receive from an unbuffered channel starts before
 the send on that channel completes.
 </p>
 
@@ -296,7 +296,7 @@ crash, or do something else.)
 </p>
 
 <p class="rule">
-The <i>k</i>th receive on a channel with capacity <i>C</i> happens before the <i>k</i>+<i>C</i>th send from that channel completes.
+The <i>k</i>th receive on a channel with capacity <i>C</i> starts before the <i>k</i>+<i>C</i>th send from that channel completes.
 </p>
 
 <p>
--- a/doc/go_mem.html
+++ b/doc/go_mem.html
@@ -213,7 +213,7 @@ usually in a different goroutine.
 </p>
 
 <p class="rule">
-A send on a channel starts before the corresponding
+A send on a buffered channel happens before the corresponding
 receive from that channel completes.
 </p>
 
@@ -256,7 +256,7 @@ yields a program with the same guaranteed behavior.
 </p>
 
 <p class="rule">
-A receive from an unbuffered channel starts before
+A receive from an unbuffered channel happens before
 the send on that channel completes.
 </p>
 
@@ -296,7 +296,7 @@ crash, or do something else.)
 </p>
 
 <p class="rule">
-The <i>k</i>th receive on a channel with capacity <i>C</i> starts before the <i>k</i>+<i>C</i>th send from that channel completes.
+The <i>k</i>th receive on a channel with capacity <i>C</i> happens before the <i>k</i>+<i>C</i>th send from that channel completes.
 </p>
 
 <p>


davecheney:
Please abandon this change. _happens before_ is a well understood concept when discussing memory synchronising events. It is defined in the document https://golang.org/ref/mem#tmp_2

hanyucui:
> Please abandon this change. _happens before_ is a well understood concept when discussing memory synchronising events. It is defined in the document https://golang.org/ref/mem#tmp_2

"Happens before" is well understood, but the way the doc uses it is inaccurate. Like I said in the description, send and receive have duration. When the doc says a send happens, it is not entirely clear whether it is the start of the send that happens or the completion of the send. Although I would usually assume it is talking about the start of the send, it doesn't hurt to further clarify it and leave no room for ambiguity.

davecheney:
I disagree and again ask you to abandon this change.

randall77:
"happens before" is a term of art. Please don't change that wording.

This edit might be what you're trying to say:

> A send on a channel happens before the corresponding receive from that channel completes.

> The start of a send on a channel happens before the corresponding ￼receive from that channel completes.

I'm not sure that really provides any additional useful information, though. How is the distinction between the start of a send and some arbitrarily chosen point during the send relevant? It generates the same happens-before relationships.

hanyucui:
@googlebot I signed it!

hanyucui:
> "happens before" is a term of art. Please don't change that wording.
> 
> This edit might be what you're trying to say:
> 
> > A send on a channel happens before the corresponding receive from that channel completes.
> 
> > The start of a send on a channel happens before the corresponding ￼receive from that channel completes.
> 
> I'm not sure that really provides any additional useful information, though. How is the distinction between the start of a send and some arbitrarily chosen point during the send relevant? It generates the same happens-before relationships.

@randall77 Thanks for spending the time. I understand Go's documentation generally tries to be simple and practical, but in this particular case, the wording leaves a lot of room for different interpretations. My PR is just one possible interpretation. I don't claim it's entirely correct but just wanted to start a conversation. I can certainly withdraw the PR and start a separate discussion instead. To name a few things here:
- Send and receive are treated differently in a single sentence. If we omit "start", we shouldn't mention "complete" either, which is a clear message we are not looking into the details of send and receive, e.g., "A send on a channel happens before the corresponding receive from that channel".
- A single operation, be it send or receive, is treated inconsistently at different places. The sentence you quote here, it does not care whether a send is starting or completing, but in the unbuffered channel case, it explicitly says "before the send on that channel **completes**".
- Another possible interpretation is, the sentence you quote only applies to a "buffered" channel and we should add "buffered" to the statement, otherwise, the two statements contradict each other: an unbuffered channel is also a channel, and how can a send on a channel happen before a receive but a send on an unbuffered channel (still a channel) happen after a receive?
- The definition of happens-before explicitly talks about only reads and writes. Sends and receives are higher-level operations and both could contain multiple reads and writes. Hence happens-before does not directly apply to send and receive if we don't look into the reads and writes they perform. 

Overall, I think the documentation is ambiguous. Of course, experienced Go programmers probably have more context and implicit knowledge here. And I am sure the code would have no ambiguity. However, I am guessing the maintainers want the spec to transcend any concrete implantation, but it's not quite there yet.

randall77:
> Send and receive are treated differently in a single sentence. If we omit "start", we shouldn't mention "complete" either, which is a clear message we are not looking into the details of send and receive, e.g., "A send on a channel happens before the corresponding receive from that channel".

That sounds fine to me (dropping "completes").

There's a tricky case, though, which is unbuffered channels. The problem there is that we want both

- The start of a send on an unbuffered channel happens-before the end of the receive on that channel.
- The start of a receive on an unbuffered channel happens-before the end of the send on that channel.

If we just drop start/end, then the happens-before relationship now has a cycle.
We need some way to distinguish start vs. end, for this case. (I think this is the only place where it matters.)

> The definition of happens-before explicitly talks about only reads and writes.

Happens-before relates all the operations, as is clear from saying things like "The go statement that starts a new goroutine happens before the goroutine's execution begins." I guess what that really means is that the last memory operation before the `go` statement executes happens-before the first memory operation of the new goroutine. But you kind of get that from the transitive property of happens-before. Imagine introducing new "nodes" like "go statement execution" and "goroutine beginning" which aren't really memory operations, but pseudo-operations.

The definition of happens-before says "memory operations". It should probably be "memory and synchronization operations", perhaps? Those would include channel sends and receives, goroutine starts, locks, ....
But we're not developing a formal proof system for happens-before, so I don't want to start defining classes of nodes and edges in a partially ordered DAG or anything. We're trying to describe it in a way that's readable and understandable, if somewhat imprecise.

>  Sends and receives are higher-level operations and both could contain multiple reads and writes.

Sends and receives don't contain any reads or writes. At least, at the spec level they don't. (The implementation does, but that's not relevant in the spec.)

hanyucui:
> 
> > The definition of happens-before explicitly talks about only reads and writes.
> 
> Happens-before relates all the operations, as is clear from saying things like "The go statement that starts a new goroutine happens before the goroutine's execution begins." I guess what that really means is that the last memory operation before the `go` statement executes happens-before the first memory operation of the new goroutine. But you kind of get that from the transitive property of happens-before. Imagine introducing new "nodes" like "go statement execution" and "goroutine beginning" which aren't really memory operations, but pseudo-operations.

Yeah, when we say X happens before Y, we usually mean the last memory operation of X happens before Y. In the case of "a send on a channel happens before the corresponding receive from that channel completes", it would mean the last memory operation of the send happens before the corresponding receive from the same channel completes, which is only true for a buffered channel with available capacity, not just any channel. I think you are more knowledgeable on Go, so I just pushed a different interpretation that you seem to be suggesting.

@randall77 Again, I really appreciate your spending the time, and I agree with most of what you said. I just wish the documentation could be more precise, but it could be just me. We can close the PR if the change doesn't make sense. Thanks to you, I've had the clarification I need.

PR #40778, opened 2020-08-14
Correct words

Howie59:
I correct three wrong words


Howie66:
> @googlebot I signed it!

@googlebot I signed it!

PR #40801, opened 2020-08-14
plugin: add OpenFile to allow loading of Plugins from file descriptors

fionera:
This allows loading of plugins that are stored in a memfd. Thanks to @terorie for the initial testing.


fionera:
@googlebot I signed it!

Issue #40874, opened 2020-08-18
x/build/cmd/gerritbot: PRs that are closed and re-opened no longer continue to get imported


randall77:
We saw this in CL 239710, where a contributor accidentally closed a PR, reopened it, and was unable to get further updates into gerrit.
@dmitshur @FiloSottile @qingyunha

dmitshur:
We have logic in GerritBot so that a Gerrit CL is automatically abandoned when the PR author closes the PR.

We also have logic so that a GitHub PR is closed when a Gerrit CL is abandoned by a reviewer (including copying the abandon reason to the GitHub PR).

It seems that PRs that are closed and re-opened are not supported. We can use this as the tracking feature to add support for that edge case. Until it's supported, a workaround is to avoid closing PRs or to send a new PR if it happens unintentionally.

/cc @toothrot @andybons @cagedmantis per [owners](https://dev.golang.org/owners).

PR #40877, opened 2020-08-19
cmd/internal/objfile: cache computation of goobj.Arch

qingyunha:


qingyunha:
Reopen the RP due to a bug of GerritBot  https://github.com/golang/go/issues/40874

PR #40903, opened 2020-08-19
net/http: refactor ResponseWriter.ReadFrom to permit splice on Linux

### LINKED ISSUE #40888: net/http: permit splice from socket to ResponseWriter

<pre>
$ go version
go version go1.15 darwin/amd64
</pre>

(my complaint actually helps linux and is a non-issue on darwin, but I target linux hosts too)

yes

$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/pforgey/Library/Caches/go-build"
GOENV="/Users/pforgey/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/pforgey/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/pforgey/go"
GOPRIVATE=""
GOROOT="/Users/pforgey/goroot"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/Users/pforgey/goroot/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="0"
GOMOD="/Users/pforgey/goroot/src/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/ck/hxb66hs93wd58w5ylk4v0vfxk_3ncr/T/go-build576313242=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre></details>

### What did you do?

Assumed a simple io.Copy() between a socket and an http.ResponseWriter would go through splice()

To be treated like io.Copy() between two spliceable FDs

### What did you see instead?
A surprising, but not unreasonable workaround to issue #5660 caused by the fact we start writing headers immediately before the source reader is readable.

Instead of going through the trouble of determining if the source is a file before proceeding along the presumed sendfile (but maybe splice!) route, I have successfully tested a local patch (attached) which does a small read instead. Regardless of needsSniff(), this satisfies the sniff step, but also waits for the source to be actually readable.

A small body of < 512 (sniffLen) bytes would have little benefit from sendfile/splice, and we were already doing a syscall round trip. Would the extra and and write of 512 bytes from kernel->user->kernel have a significant impact in the overall case of the larger bodies? I'm not too sure it would, but it is a legitimate point to object with, but it wouldn't be the first example of zero copy network code slopping a little bit of the initial data portion like that. For the smaller bodies, it may actually have slightly less setup time.

(removed gist link to patch and submitting a PR instead)

PaulForgey:
Rather than probe and guess if sendfile will work inside ResponseWriter.ReadFrom(src),
this change fixes the underlying issue of starting to respond before src is readable
We'll no longer send a status OK if a header has not yet been written and reading
from src is destined to fail. This small change implicitly takes care of the need for
the server to sniff the response body to determine the Content-Type.

This allows splice to work on Linux when src is a socket or any non-regular file that's spliceable.

The extra read of 512 bytes may raise an objection, and that's fair, but
we're already swapping some syscall prep work for another and a read of
512 probably will not impact the overall performance. For shorter
bodies, there's likely less setup time. A little initial slop is not too
unusual in zero copy network code, and sometimes actually helps.

Fixes #40888


PaulForgey:
@googlebot I signed it!

PR #41000, opened 2020-08-24
sync: document keys of sync.Map will never be deleted

### LINKED ISSUE #40999: sync: sync.Map keys will never be garbage collected

<pre>
$ go version
go1.15
</pre>

yes

Can reproduce on any os and arch.

### What did you do?

We are using some big struct as keys for `sync.Map`, struct{} as value, and we observed a memory leak after upgrading to go 1.15. I confirmed that this is because `sync.Map` will never delete keys. Instead, it only set the value to nil.

It worked well before 1.15 because we happened to have no read operation during the lifetime, and we were only reading it during shutdown so we didn't discover the memory leak, as some code like this: https://play.golang.org/p/YdY4gOcXVMO. So we happened to only have no key prompted to `sync.Map.read`, and thus Delete could delete the key in `sync.Map.dirty`.

In go1.15, this behaviour was changed by https://go-review.googlesource.com/c/go/+/205899, and causes a memory leak in our code.

This isn't the same as the behaviour of the native map. I admit I have misused `sync.Map`, but I think this behaviour should  either be documented clearly or be changed.

PureWhiteWu:
Keys of sync.Map will never be deleted, this behaviour isn't the same as the native map.
We should document this clearly.

Fixes #40999 .

### DIFF
--- a/src/sync/map.go
+++ b/src/sync/map.go
@@ -24,6 +24,7 @@ import (
 // contention compared to a Go map paired with a separate Mutex or RWMutex.
 //
 // The zero Map is empty and ready for use. A Map must not be copied after first use.
+// Keys in the Map will never be deleted, thus they will never be garbage collected.
 type Map struct {
 	mu Mutex
 


PureWhiteWu:
closed as discussed.

lilien1010:
I think it should not change the document, we shoud change the code ,it shoud be  a BUG

PR #41041, opened 2020-08-26
time: avoid racing condition in NewTricker example

secret104278:
Avoid racing condition due to golang non-priority selection in NewTricker example.

Signed-off-by: Yu Chih <secret104278@gmail.com>

### DIFF
--- a/src/time/example_test.go
+++ b/src/time/example_test.go
@@ -195,10 +195,17 @@ func ExampleNewTicker() {
 		done <- true
 	}()
 	for {
+		// b/c there is no priority selection in golang.
+		// It is possible to race when both channel are
+		// non-empty, and trigger ticker.C after 10 seconds.
+		// In order to mitigate we should check done seperatly.
 		select {
 		case <-done:
 			fmt.Println("Done!")
 			return
+		default:
+		}
+		select {
 		case t := <-ticker.C:
 			fmt.Println("Current time: ", t)
 		}


docmerlin:
This is turns a lock into a spin-lock. I'm not sure this is exactly what you want.

openset:
@docmerlin This example is may be better for `select`, not good for `NewTricker`.

secret104278:
Agree, maybe this is not a good place.

BTW, I'm curious abound that theoretically, there will be 50% chance that the example should print "m=+11". But no matter how hard I try, it always succeeded at "m=+10". Is it cause by the underlying go timing mechanism which will lead to this result?

openset:
@secret104278 `for` is too faster.
```go
func main() {
	now := time.Now()
	t1 := time.NewTicker(time.Millisecond)
	defer t1.Stop()
	done := make(chan bool)
	go func() {
		time.Sleep(2 * time.Second)
		done <- true
	}()
	for {
		select {
		case t := <-t1.C:
			fmt.Println("t1:", t)
			time.Sleep(time.Second)
		case <-done:
			fmt.Println("Done:", time.Now().Sub(now))
			return
		}
	}
}
// Output:
Done: 2.00168602s   || 
Done: 4.00392365s   || 
Done: 6.002414447s
```

secret104278:
@openset Not sure is there any relationship with speed of for.

Image when you are at 9.xxxx s, the execution is actually blocking at **select** statement.
When the 10th second comes, there are two possible situation:
**CASE A**
```
time.Sleep exit (in goroutine)
done <- true (in goroutine)
Ticker send Tick to t1
```
will print `Done` at 10th second.
**CASE B**
```
Ticker send Tick to t1
time.Sleep exit (in goroutine)
done <- true (in goroutine)
```
will print `Done` at 11th second.

Theoretically, there are 50/50% change for CASE A and CASE B, but actually there is only CASE A.

PR #41045, opened 2020-08-26
net/http: add a package to parse and serialize Structured Field Values

### LINKED ISSUE #41046: Add support for the upcoming "Structured Field Values for HTTP" RFC
[Structured Field Values for HTTP](https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html) is an upcoming RFC from the HTTP Wording Group defining a set of well-defined data types to use in HTTP headers and trailers. 

This new format will improve the interoperability and the safety of HTTP by allowing to create generic parsers and serializers suitable for all HTTP headers. It is already used in the wild, for instance for the new security headers supported by Google Chrome (`Sec-Fetch-Dest` etc), the [`Signature` proposal](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#signature-header) in [Prefer-Push](Prefer-Push) or in [Vulcain](https://vulcain.rocks).

After the RFC publication, it would be nice to be able to parse and generate such headers directly using the standard library.

I proposed a patch to support this RFC in #41045. The code is also available as a standalone library: https://github.com/dunglas/httpsfv

dunglas:
Structured Field Values fields value for HTTP is an upcoming RFC
defining data types for headers and trailers.
This specification is already used by some popular software
(including Google Chrome) and will most likely be used for most
new standard (and maybe custom) HTTP headers.
This PR is proposal to introduce a new package implementing the specification.

Fixes golang/go#41046.

New methods are also added to the Header type to manipulate structured
values.

This implementation is also available as a standalone package:
https://github.com/dunglas/httpsfv

I also published a blog post about this:
https://dunglas.fr/2020/08/a-structured-http-fields-parser-and-serializer-for-the-go-programming-language/

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/net/http/sfv/structured-field-tests"]
+	path = src/net/http/sfv/structured-field-tests
+	url = https://github.com/httpwg/structured-field-tests
--- a/src/go/build/deps_test.go
+++ b/src/go/build/deps_test.go
@@ -414,6 +414,16 @@ var depsRules = `
 	< golang.org/x/net/idna
 	< golang.org/x/net/http/httpguts, golang.org/x/net/http/httpproxy;
 
+	encoding/base64,
+	errors,
+	fmt,
+	io,
+	math,
+	reflect,
+	strconv,
+	strings
+	< net/http/sfv;
+
 	NET, crypto/tls
 	< net/http/httptrace;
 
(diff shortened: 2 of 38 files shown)


dunglas:
Related proposal: golang/go#41046

dunglas:
For the record, this code is available as a standalone package: https://github.com/dunglas/httpsfv

heschi:
I closed old PRs to reduce load on the Gerrit importer (#50197), sorry for the trouble. I'll reopen the CL and PR.

dunglas:
No worries. I think this one can stay closed at least while the proposal to include such feature in core hasn't be accepted.

PR #41314, opened 2020-09-10
cmd/go: allow output in non-existent directory

### LINKED ISSUE #41313: cmd/go: inconsistent result when go build output is a non-existent directory

go1.15 darwin/amd64

Yes

### What did you do?

<details>
<summary>Project Hierarchy</summary>

```
.
├── go.mod
└── main.go
```

</details>

The first time to build, `go` complain about an existent directory that is actually created by `go` itself and not exists before the build.

```bash
$ go build -o bin/ main.go
go build command-line-arguments: build output "bin/" already exists and is a directory
```

<details>
<summary>Project Hierarchy</summary>

```
.
├── bin
├── go.mod
└── main.go
```

</details>

The second time to build. Since the directory already exists, `go` works fine at this time.

```bash
$ go build -o bin/ main.go
```

<details>
<summary>Project Hierarchy</summary>

```
.
├── bin
│   └── main
├── go.mod
└── main.go
```

</details>

An idempotent result from `go build`.

1. Keep complaining about the directory

```
$ go build -o bin/ main.go
go build command-line-arguments: build output "bin/" already exists and is a directory
$ go build -o bin/ main.go
go build command-line-arguments: build output "bin/" already exists and is a directory
```

2. Allow `go build` output is an inexistent directory

-----

I prefer the second solution since it's more convenient for users.

Pros:
- No need to create a directory by users.
- Easy to write a common build command in scripts. (Putting binary in a directory seems like a common convention to build)

Before: `go build -o bin/foo .` `go build -o bin/bar .` (We need to change the output name)
After: `go build -o bin/ .`

- Consistency. `go build -o bin/foo .` also creates the directory for users.

XSAM:
When 'go build' is given an output path with -o, if the output path
ends with a path separator, always treat it as a directory.

Fixes #41313

### DIFF
--- a/src/cmd/go/alldocs.go
+++ b/src/cmd/go/alldocs.go
@@ -93,8 +93,9 @@
 //
 // The -o flag forces build to write the resulting executable or object
 // to the named output file or directory, instead of the default behavior described
-// in the last two paragraphs. If the named output is a directory that exists,
-// then any resulting executables will be written to that directory.
+// in the last two paragraphs. If the named output is an existing directory or
+// ends with a slash or backslash, then any resulting executables
+// will be written to that directory.
 //
 // The -i flag installs the packages that are dependencies of the target.
 //
--- a/src/cmd/go/internal/work/build.go
+++ b/src/cmd/go/internal/work/build.go
@@ -53,8 +53,9 @@ serving only as a check that the packages can be built.
 
 The -o flag forces build to write the resulting executable or object
 to the named output file or directory, instead of the default behavior described
-in the last two paragraphs. If the named output is a directory that exists,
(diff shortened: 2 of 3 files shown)


tpaschalis:
Hello, and thanks for attempting to solve this issue!

The error that you're referencing _is not_ coming from `runBuild` but from `copyFile` which is present [here](https://github.com/golang/go/blob/3f568625980f6e23a603cfd1cb4fcd2bf5c895d7/src/cmd/go/internal/work/exec.go#L1671). 

Let's continue the investigation on the parent issue first, and  understand what's really happening here. This will allow to propose a solid solution, along with the relevant tests to avoid future regressions.

PR #41359, opened 2020-09-12
math/big: correct a typo in the SetBit method documentation

szerfas1:
I believe there's a mistake in the documentation.

`f b is 0 SetBit sets z = x &^ (1 << i). If b is not 0 or 1,` should be `f b is 0 SetBit sets z = x & ~(1 << i). If b is not 0 or 1,`

(That is, switch the `^` operator, which is binary, to the `~` operator, which unary.)


szerfas1:
@googlebot I signed it!

PR #41371, opened 2020-09-14
Magic for Ian

JackDanger:
Correcting English pluralization.


davecheney:
no thank you, this is the plural of the abbreviation, `GOOS`

PR #41433, opened 2020-09-16
syscalls: Separate illumos from solaris.

cneira:
Although illumos and solaris share syscalls, Illumos provide ones that
solaris does not have, at this point is easier for illumos to have it's own syscall table code so changes won't affect solaris.


cneira:
@googlebot I signed it!

PR #41560, opened 2020-09-22
mime/multipart: wrap error returned from Part.NextPart()

mattnibs:
Part.NextPart() returns io.EOF on the first NextPart call that reaches the end of
a form. On subsequent calls it returns an unwrapped io.EOF. Change the
returned error to a wrapped error so at least `errors.Is` can be used to introspect
the true nature of the returned error.


mattnibs:
@googlebot I signed it!

PR #41564, opened 2020-09-22
codehost: forward the error from the underlying git call

legitparty:
The loadRefs() method is called in several places in this file. The 3 line pattern is used in all other calls to loadRefs(), but for some reason this one has been missed, and not having that underlying error has made finding the true error much more difficult, because I have encountered several different reasons already.


legitparty:
@googlebot I signed it!

legitparty:
@googlebot I fixed it.

PR #41624, opened 2020-09-25
syscall: add more ulimit constants

bozaro:
I whitelisted more ulimit constants in `zerrors_linux_*.go` generator.
For example:
```diff
--- a/src/syscall/zerrors_linux_amd64.go
+++ b/src/syscall/zerrors_linux_amd64.go
        RLIMIT_FSIZE                     = 0x1
+       RLIMIT_MEMLOCK                   = 0x8
        RLIMIT_NOFILE                    = 0x7
+       RLIMIT_NPROC                     = 0x6
+       RLIMIT_RSS                       = 0x5
        RLIMIT_STACK                     = 0x3
```


bozaro:
I closed it as unnecessary:

> The syscall package is frozen as per its godoc comment. New functionality should be added to the golang.org/x/sys/unix package instead. It seems the constants you want to add are already available there.

PR #41751, opened 2020-10-02
net/http: remove duplicate declaration of error

geektutu:
there is no need to declare a error variable here.


geektutu:
@googlebot I signed it!

PR #41957, opened 2020-10-13
syscall: expose bInheritHandles of CreateProcess

JohanKnutzen:
Certain use cases require this parameter to be false. This includes
spawning a child process in a different windows session than session 0.

Docs regarding the behavior of this parameter to CreateProcess:
https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa

Fixes #42098


yohan1234:
@googlebot I signed it!

PR #42042, opened 2020-10-17
io: refactor teeReader's Read func

0x2d3c:
Rewriting the implementation of nested if will make teeReader processing Short and easy to understand


0x2d3c:
@googlebot I signed it!

Issue #13400, opened 2015-11-25
encoding/xml: fix name spaces


rsc:
There are many pending issues related to the handling of xml name spaces. We need to rethink this. One more try and then we're going to give up.

- 7535	encoding/xml: Encoder duplicates namespace tags
- 11496	encoding/xml: Serializing XML with namespace prefix
- 9775	encoding/xml: Unmarshal does not properly handle NCName in XML namespaces
- 12624	encoding/xml: brittle support for matching a namespace by identifier or url
- 8167	encoding/xml: disallow attributes named xmlns:*
- 11735	encoding/xml: empty namespace conventions are badly documented
- 8068	encoding/xml: empty namespace prefix definitions should be illegal
- 8535	encoding/xml: failure to handle conflicting tags in different namespaces
- 11431	encoding/xml: loss of xmlns= in encoding since Go 1.4
- 7113	encoding/xml: missing nested namespace not handled
- 11724	encoding/xml: namespaced and non-namespaced attributes conflict
- 7233	encoding/xml: omit parent tags if value is empty #2
- 12406	encoding/xml: support QName values / expose namespace bindings
- 9519	encoding/xml: support for XML namespace prefixes

aeden:
> then we're going to give up

Please don't give up :smile:

justinlindh-wf:
Seconding the comment above. Please don't give up on this. There is still value in handling namespaces properly in golang XML.

rsc:
I did say we'd try one more time before giving up.

casualjim:
+1 for not giving up

simplicitylab:
+1 I was researching Go as a possible replacement for Python in processing ODT files and I was quickly hitting snags when dealing with multiple namespaces.

keltia:
I feel your pain, I'm also stuck with SOAP-based web services and the namespace issue is a thorn here.  Going to generate XML w/ templates for now.  Please do not give up :)

eastany:
+1

rsc:
In general we don't pay attention to comments that only say "+1". In fact,
I think we have a bot that deletes them.

What would be very useful instead would be pointers to programs (ideally
self-contained ones on play.golang.org) illustrating what you are trying to
do with encoding/xml, what output you wish you were getting, and how the
current output differs/breaks your use case. That information would help
guide any eventual fixes we make here.

Thanks.

keltia:
[...]

PR #42043, opened 2020-10-17
encoding/xml: increase test coverage

charlesjlee:
Add tests to increase coverage from 86.2% to 93.3%

For #6094


charlesjlee:
I didn't address the original request
>name spaces, custom marshalers, custom unmarshalers, and the interaction between all those

because name space support is currently inconsistent, see https://github.com/golang/go/issues/13400

PR #42052, opened 2020-10-18
cmd/buildid: move and reuse duplicated HashToString code to cmd/internal/buildid/buildid

proggga:


proggga:
Please add 'hacktoberfest-approved'

PR #42059, opened 2020-10-19
cmd/compile/internal/gc: use plural nouns in comments

hkhere:


dsmello (approved):
That's doesn't break anything. 
😄

hkhere:
I think I made a bad choice. The purpose was to make it consistent and now I think I should have chosen singular forms as suggested by the code review comment.

PR #42083, opened 2020-10-20
net/http: cst.close() should be called before t.Fatal()

caixiangyue:


caixiangyue:
@googlebot I signed it!

PR #42120, opened 2020-10-21
cmd/compile/internal/gc: fix comments

hkhere:


hkhere:
Sorry, `xN.Innermost` should be `xN.Defn`, not `x1.Innermost`. @rsc

Issue #41960, opened 2020-10-14
log/syslog: writes hostname when using customized syslog socket because “unixgram” or “unix” are not recognized as local transports 


larryleguo:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.15 linux/amd64
</pre>

### Does this issue reproduce with the latest release?

Yes.

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/usr/local/google/home/leguo/.cache/go-build"
GOENV="/usr/local/google/home/leguo/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/usr/local/google/home/leguo/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/usr/local/google/home/leguo/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/lib/google-golang"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/google-golang/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build869140076=/tmp/go-build -gno-record-gcc-switches"
</pre></details>

### What did you do?

syslog.New() uses default syslog socket "/dev/log". To specify socket path, syslog.Dial("unixgram", "/dev/log", ...) can be used. The behaviors should be the same but but they are not. Running the attached example code:

[syslog_debug.go.txt](https://github.com/golang/go/files/5374835/syslog_debug.go.txt)

$ go run syslog_debug.go; tail /var/log/debug -n 100 | grep 'Syslog debug: logging with'

[...]

PR #42122, opened 2020-10-21
Fix issue #41960

imxyb:
Fix issue #41960


imxyb:
@googlebot I signed it!

Issue #36612, opened 2020-01-16
cmd/compile: set prologue_end on every arch


derekparker:
As of Go 1.11 and commit https://github.com/golang/go/commit/7bac2a95f65418ec18d1fb5f30ae47e91980f30c support was added to put information about the end of the function prologue into the produced DWARF information, however it currently only works for x86.

Ideally the DWARF information produced is consistent across arches. In Delve if this information is not present we have to disassemble the function in order to pattern match known function prologues in order to skip them. This is undesirable to say the least and I would like to delete that code.

Opening this issue to track adding prologue_end to the DWARF information produced on every arch.

derekparker:
cc @heschik @dr2chase @aarzilli

heschik:
https://golang.org/cl/223297 did part of this. Any more planned for 1.15?

heschik:
...though that CL isn't merged yet.

dr2chase:
Any time you want to +2 it....
It would make sense to maybe get ahead on some other architectures, might help gdb, also depends on where delve is going next.

odeke-em:
The CL got merged on May 13th aka 16 days ago. @dr2chase I shall move this to Go1.16 then.

dr2chase:
I'm okay with go1.16.  Part of the problem is knowing exactly what's right; because of better-user-experience issues, Delve wants it to be someplace that apparently doesn't match what gdb would expect.  (My personal user experience with gdb is so terrible that I would never recommend it to a Go user.)

aclements:
@dr2chase , what's the current status of this? Are you planning to do any more on this for 1.16 or should we bump it to 1.17?

dr2chase:
Derek Parker (i.e., Delve developers) recently added ppc64.
I think this stays live till all the architectures are done, which might take a while.

One reason to wait on demand from Delve is that ensures we do it the way they want, instead of accidentally picking wrong and then someone else depends on that choice.

ianlancetaylor:
Moving to Backlog.

PR #42261, opened 2020-10-28
cmd/internal/obj: add prologue_end DWARF stmt for ppc64

derekparker:
This patch adds a prologue_end statement to the DWARF information for
the ppc64 arch.

Prologue end is used by the Delve debugger in order to determine where
to set a breakpoint to avoid the stacksplit prologue.

Updates #36612


derekparker:
cc @dr2chase @aarzilli @heschik

PR #42262, opened 2020-10-28
syscall: call tcsetpgrp from parent thread when spawning a foreground child process

### LINKED ISSUE #37217: os/exec: (*SysProcAttr).Foreground causes the spawned process to hang

<pre>
$ go version
go version go1.13.6 linux/amd64
</pre>

Yes — tried with `go1.14rc1`

$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/markse/.cache/go-build"
GOENV="/home/markse/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/markse"
GOPRIVATE=""
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build088413182=/tmp/go-build -gno-record-gcc-switches"
</pre></details>

### What did you do?

```go
package main

import (
    "os"
    "os/exec"
    // "os/signal"
    "syscall"
)

func main() {
    // This does make it work, but shouldn't be necessary and causes side-effects
    // signal.Ignore(syscall.SIGTTIN, syscall.SIGTTOU)

    cmd := exec.Command("echo", "hello")
    cmd.SysProcAttr = &syscall.SysProcAttr{
        Foreground: true,
    }
    cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
    cmd.Run()
}

```

The process should execute `echo` and display `hello`.

### What did you see instead?

The process hangs.

`ps` shows the child process as suspended and doesn't appear to have called `execve` yet:

```console
$ ps j
 PPID   PID  PGID   SID TTY      TPGID STAT   UID   TIME COMMAND
…
 6072 25553 25553  6072 pts/0    25553 Dl+   1000   0:00 ./foreground
25553 25558 25558  6072 pts/0    25553 T     1000   0:00 ./foreground
…
```

# Explanation

From what we can determine, we believe this is what is happening:

1. The process is forked ([ref](https://github.com/golang/go/blob/go1.14rc1/src/syscall/exec_linux.go#L205))
2. The child calls `setpgid`/`SYS_SETPGID`, creating a new background process group ([ref](https://github.com/golang/go/blob/go1.14rc1/src/syscall/exec_linux.go#L265))
3. The child calls `tcsetpgrp`/`TIOCSPGRP`, causing the process to be suspended ([ref](https://github.com/golang/go/blob/go1.14rc1/src/syscall/exec_linux.go#L280))

While a process in the foreground process group is allowed to call `tcsetpgrp`, a process in a background process group has an additional hurdle. From the man page for `tcsetpgrp`:

> If `tcsetpgrp()` is called by a member of a background process group in its session, and the calling process is not blocking or ignoring `SIGTTOU`, a `SIGTTOU` signal is sent to all members of this background process group.

Which explains why the process is getting suspended, as the default action for `SIGTTOU` is to stop the process. This also explains when ignoring `SIGTTOU` before spawning the process fixes the problem. However, this has an undesired side-effect: the subsequently exec'd process then ignores `SIGTTOU` as well.

# Conclusion

Unfortunately, because all of this work is done in the child fork, there is no way for Go user code to reset the ignored `SIGTTOU` before exec'ing the process.

So there is no way to have a user ignore the `SIGTTOU` signal without an adverse effect on the child process.

Proposal: if the `SIGTTOU` signal could be ignored for the duration of the `tcsetpgrp`/`TIOCSPGRP` call and restored to its prior state immediately following, we believe this would elicit the expected behavior without adverse side-effects.

pupper68k:
Currently we call tcsetpgrp from child thread causing the foreground process to hang.
The child pgrp is either found in the SysProcAttr or equal to the child pid.
Since we have child pgrp we can call tcsetpgrp from the parent thread.
Fixes #37217

### DIFF
--- a/src/syscall/exec_linux.go
+++ b/src/syscall/exec_linux.go
@@ -90,7 +90,19 @@ func forkAndExecInChild(argv0 *byte, argv, envv []*byte, chroot, dir *byte, attr
 
 	// parent; return PID
 	pid = int(r1)
+	if sys.Foreground {
+		pgrp0 := int32(sys.Pgid)
+		if pgrp0 == 0 {
+			pgrp0 = int32(pid)
+		}
 
+		// Place process group in foreground.
+		_, _, err3 := RawSyscall(SYS_IOCTL, uintptr(sys.Ctty), uintptr(TIOCSPGRP), uintptr(unsafe.Pointer(&pgrp0)))
+		if err3 != 0 {
+			return 0, err3
+		}
+	}
+	
 	if sys.UidMappings != nil || sys.GidMappings != nil {
 		Close(p[0])
 		var err2 Errno
@@ -278,21 +290,6 @@ func forkAndExecInChild1(argv0 *byte, argv, envv []*byte, chroot, dir *byte, att
 	if sys.Setpgid || sys.Foreground {
 		// Place child in process group.
 		_, _, err1 = RawSyscall(SYS_SETPGID, 0, uintptr(sys.Pgid), 0)
-		if err1 != 0 {
-			goto childerror
-		}
-	}
-
-	if sys.Foreground {
-		pgrp := int32(sys.Pgid)
-		if pgrp == 0 {
-			r1, _ = rawSyscallNoError(SYS_GETPID, 0, 0, 0)
-
-			pgrp = int32(r1)
-		}
-
-		// Place process group in foreground.
-		_, _, err1 = RawSyscall(SYS_IOCTL, uintptr(sys.Ctty), uintptr(TIOCSPGRP), uintptr(unsafe.Pointer(&pgrp)))
 		if err1 != 0 {
 			goto childerror
 		}


pupper68k:
@googlebot I signed it!

PR #42272, opened 2020-10-29
encoding/csv: improve docs for ReadAll when no input is left

3noch:
`encoding/csv`'s `ReadAll` returns `nil` instead of `[][]string{}` when there is no input left. This is not documented but should be.


3noch:
@googlebot I signed it!

Issue #41509, opened 2020-09-20
encoding/asn1: error instead of panic on invalid value to Unmarshal


KimMachineGun:
Most packages for encoding return an error when decoding(or unamrshaling) fails due to an invalid argument.
But sadly, the `encoding/asn1` panics instead of returning an error, and I think it's pretty dangerous because it is unpredictable and fragile.

So, I propose that `asn1.Unmarshal` should return an error on invalid argument.

Since this proposal can make breaking changes, I'm not sure if this poposal comply to the [go 1 compatibility promise](https://golang.org/doc/go1compat). However, even if this proposal is not accepted, it may be worth mentioning that 'a panic can occur during unmarshaling' in the `encoding/asn1` document.

### PR
PR for this proposal: #41485 
In this PR, the `asn1.Unmarshal` verifies that argument is non-nil pointer. And if it isn't, `asn1.Unmarshal` will return a new error `InvalidUnmarshalError` like `encoding/json`.

### Examples
https://play.golang.org/p/7ChlnCY_ioI
```go
package main

import (
	"encoding/asn1"
)

func main() {
	// panic: reflect: call of reflect.Value.Elem on zero Value
	asn1.Unmarshal([]byte{0x05, 0x00}, nil)

	// panic: reflect: call of reflect.Value.Elem on struct Value
	asn1.Unmarshal([]byte{0x05, 0x00}, asn1.RawValue{})

	// panic: reflect: call of reflect.Value.Type on zero Value
	var p *asn1.RawValue
	asn1.Unmarshal([]byte{0x05, 0x00}, p)
}

```

rolandshoemaker:
Seems like a reasonable change. This would only break code that relied on a panic happening to prevent brokenness, which seems less than ideal in the first place. Is there a precedent for holding up changes that fix this kind of behavior (there is a [similar issue in crypto/x509](https://github.com/golang/go/issues/41169) as well).

slrz:
> Most packages for encoding return an error when decoding(or unamrshaling) fails due to an invalid argument.
> But sadly, the `encoding/asn1` panics instead of returning an error, and I think it's pretty dangerous because it is unpredictable and fragile.

It seems to me that a panic can only be triggered by passing a bogus *second* argument to Unmarshal. That's a programmer error and thus **should** panic. It cannot happen with correct code, so I don't see what's unpredictable about it.

[...]

PR #42315, opened 2020-11-01
doc/go1.16: document new behavior of asn1.Unmarshal on invalid argument

KimMachineGun:
For #41509


KimMachineGun:
@FiloSottile 
Could you please review it?
I think #41485 needs to be documented.

PR #42378, opened 2020-11-04
LICENSE: Changed "Google Inc." to "Google LLC"

JackWinterburn:
I noticed that the LICENSE file still has Google written down as an incorporated company, so i changed it to suit Google's current status.


JackWinterburn:
@googlebot I signed it!

PR #42383, opened 2020-11-04
net/http: This change modifies Go to add support for charset in reverseproxy for event-stream

mr-agrwal:


mr-agrwal:
@googlebot I signed it!

Issue #26088, opened 2018-06-27
net/http/httputil: make ReverseProxy forward 1xx responses


bradfitz:
[Continuation of #17739]

The ReverseProxy does not forward 1xx responses currently. 

https://tools.ietf.org/html/rfc7231#section-6.2

> A proxy MUST forward 1xx responses unless the proxy itself requested the generation of the 1xx response.

bradfitz:
This requires #26089 (Server/Handler/ResponseWriter doesn't allowing sending arbitrary 1xx responses).

PR #42597, opened 2020-11-13
net/http: allow sending 1xx responses

### LINKED ISSUE #26089: net/http: Server/Handler/ResponseWriter doesn't allowing sending arbitrary 1xx responses
As part of #17739 (properly handling 1xx responses per the HTTP specs), we need to be able to forward 1xx responses in ReverseProxy (#26088)

To forward such responses, we need a way for for an http.Handler in a Server to use the ResponseWriter (or something) to send a 1xx response.

### LINKED ISSUE #36734: net/http: support status code 102 (Processing) in ResponseWriter
HTTP status code 102 (Processing) is defined by [RFC 2518, section 10.1](https://tools.ietf.org/html/rfc2518#section-10.1):

> This status code SHOULD only be sent when the server has a reasonable expectation that the request will take significant time to complete. [...] The server MUST send a final response after the request has been completed.

As of now (Go 1.13.6), there is no way to actually send a 102 response because ResponseWriter.WriteHeader() can only be called once. Any successive calls are rejected ([here](https://github.com/golang/go/blob/8d7be1e3c9a98191f8c900087025c5e78b73d962/src/net/http/server.go#L1121-L1125)).

Since the ResponseWriter interface cannot be extended without breaking backwards compatibility, a new interface should be added (and implemented by the existing ResponseWriter implementations) that exposes a function for sending an interim response with status 102.

dunglas:
Currently, it's not possible to send informational responses such as
103 Early Hints or 102 Processing.

This patch allows calling WriteHeader() multiple times in order
to send informational responses before the final one.

If the status code is in the 1xx range, the current content of the header map
is also sent. Its content is not removed after the call to WriteHeader()
because the headers must also be included in the final response.

The Chrome and Fastly teams are starting a large-scale experiment to measure
the real-life impact of the 103 status code.
Using Early Hints is proposed as a (partial) alternative to Server Push,
which are going to be removed from Chrome:
https://groups.google.com/a/chromium.org/g/blink-dev/c/K3rYLvmQUBY/m/21anpFhxAQAJ

Being able to send this status code from servers implemented using Go would
help to see if implementing it in browsers is worth it.

Fixes #26089
Fixes #36734
Updates #26088

### DIFF
--- a/src/net/http/clientserver_test.go
+++ b/src/net/http/clientserver_test.go
@@ -1625,7 +1625,7 @@ func TestEarlyHintsRequest_h2(t *testing.T) { testEarlyHintsRequest(t, h2Mode) }
 func testEarlyHintsRequest(t *testing.T, h2 bool) {
 	defer afterTest(t)
 	if h2 {
-		t.Skip("Waiting for H2 support to be merged: https://go-review.googlesource.com/c/net/+/291029")
+		t.Skip("Waiting for H2 support to be merged: https://go-review.googlesource.com/c/net/+/406494")
 	}
 
 	var wg sync.WaitGroup
--- a/src/net/http/clientserver_test.go
+++ b/src/net/http/clientserver_test.go
@@ -1632,6 +1632,8 @@ func testEarlyHintsRequest(t *testing.T, h2 bool) {
 	wg.Add(1)
 	cst := newClientServerTest(t, h2, HandlerFunc(func(w ResponseWriter, r *Request) {
 		h := w.Header()
+
+		h.Add("Content-Length", "123") // must be ignored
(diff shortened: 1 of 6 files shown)


amitbasuri (approved):

docteurklein (approved):

dunglas:
Hi @bradfitz,

Is the tree open again? I improved the PR to flush headers as soon as possible and improved the documentation.
Also, a similar PR adding support for the 103 status code to the HTTP/3 implementation has been merged: https://github.com/lucas-clemente/quic-go/pull/3047

Cheers,

dunglas:
Hi @bradfitz,

I replied to all your comments and rebased. Is there anything else I can do to get this patch merged?

Thanks!

PR #42609, opened 2020-11-14
runtime: add force_small_arena build tag

calder:
Some embedded Linux platforms disable overcommit to reduce dynamic behavior
in the system. Go already uses 4MB arenas on 32-bit platforms, but on
64-bit platforms the 64MB arena adds a steep cost to any Go binary.

This PR adds a `force_small_arena` build tag to allow users to opt into 4MB
arenas at build time.


calder:
@googlebot I fixed it.

PR #42768, opened 2020-11-22
net: pass MSG_CMSG_CLOEXEC flag in ReadMsgUnix 

### LINKED ISSUE #42765: net: ReadMsgUnix should pass MSG_CMSG_CLOEXEC on Linux
The implementation of `ReadMsgUnix` ultimately does not pass any flags to the `recvmsg` syscall. This means that any file descriptors sent in via a Unix rights SCM message will not be marked with the close-on-exec flag. For consistent behavior with how Go handles all other file descriptors on Linux, the `MSG_CMSG_CLOEXEC` flag should be passed. The need for this is the same as all other uses of the various `CLOEXEC` flags - to prevent a race condition where a child process is forked (and exec'd) after the file descriptor is created and before it can be marked close-on-exec, and is thus leaked.

Unfortunately, I don't think there's any equivalent of this flag on Mac, so there's no easy way to address the inherent race condition there.

howjmay:
As mentioned in #42765, calling "recvmsg" syscall on Linux should come
with "MSG_CMSG_CLOEXEC" flag.

For other systems which not supports "MSG_CMSG_CLOEXEC". ReadMsgUnix()
would check the header. If the header type is "syscall.SCM_RIGHTS",
then ReadMsgUnix() would parse the SocketControlMessage and call each
fd with "syscall.CloseOnExec"

Fixes #42765

### DIFF
--- a/src/net/unixsock_readmsg_linux.go
+++ b/src/net/unixsock_readmsg_linux.go
@@ -1,0 +1,17 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build linux
+// +build linux
+
+package net
+
+import (
+	"syscall"
+)
+
+const readMsgFlags = syscall.MSG_CMSG_CLOEXEC
+
+func setReadMsgCloseOnExec(oob []byte) {
+}
--- a/src/net/unixsock_readmsg_other.go
+++ b/src/net/unixsock_readmsg_other.go
@@ -1,0 +1,13 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build (js && wasm) || windows
+// +build js,wasm windows
+
+package net
+
+const readMsgFlags = 0
+
+func setReadMsgCloseOnExec(oob []byte) {
+}
--- a/src/net/unixsock_readmsg_posix.go
+++ b/src/net/unixsock_readmsg_posix.go
@@ -1,0 +1,33 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build aix || darwin || dragonfly || freebsd || netbsd || openbsd || solaris
+// +build aix darwin dragonfly freebsd netbsd openbsd solaris
+
+package net
+
+import (
+	"syscall"
+)
+
+const readMsgFlags = 0
+
+func setReadMsgCloseOnExec(oob []byte) {
+	scms, err := syscall.ParseSocketControlMessage(oob)
+	if err != nil {
+		return
+	}
+
+	for _, scm := range scms {
+		if scm.Header.Level == syscall.SOL_SOCKET && scm.Header.Type == syscall.SCM_RIGHTS {
+			fds, err := syscall.ParseUnixRights(&scm)
+			if err != nil {
+				continue
+			}
+			for _, fd := range fds {
+				syscall.CloseOnExec(fd)
+			}
+		}
+	}
+}
--- a/src/net/unixsock_readmsg_test.go
+++ b/src/net/unixsock_readmsg_test.go
@@ -1,0 +1,105 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build aix || darwin || dragonfly || freebsd || linux || netbsd || openbsd || solaris
+// +build aix darwin dragonfly freebsd linux netbsd openbsd solaris
+
+package net
+
+import (
+	"os"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestUnixConnReadMsgUnixSCMRightsCloseOnExec(t *testing.T) {
+	if !testableNetwork("unix") {
+		t.Skip("not unix system")
+	}
+
+	scmFile, err := os.Open(os.DevNull)
+	if err != nil {
+		t.Fatalf("file open: %v", err)
+	}
+	defer scmFile.Close()
+
+	rights := syscall.UnixRights(int(scmFile.Fd()))
+	fds, err := syscall.Socketpair(syscall.AF_LOCAL, syscall.SOCK_STREAM, 0)
+	if err != nil {
+		t.Fatalf("Socketpair: %v", err)
+	}
+
+	writeFile := os.NewFile(uintptr(fds[0]), "write-socket")
+	defer writeFile.Close()
+	readFile := os.NewFile(uintptr(fds[1]), "read-socket")
+	defer readFile.Close()
+
+	cw, err := FileConn(writeFile)
+	if err != nil {
+		t.Fatalf("FileConn: %v", err)
+	}
+	defer cw.Close()
+	cr, err := FileConn(readFile)
+	if err != nil {
+		t.Fatalf("FileConn: %v", err)
+	}
+	defer cr.Close()
+
+	ucw, ok := cw.(*UnixConn)
+	if !ok {
+		t.Fatalf("got %T; want UnixConn", cw)
+	}
+	ucr, ok := cr.(*UnixConn)
+	if !ok {
+		t.Fatalf("got %T; want UnixConn", cr)
+	}
+
+	oob := make([]byte, syscall.CmsgSpace(4))
+	err = ucw.SetWriteDeadline(time.Now().Add(5 * time.Second))
+	if err != nil {
+		t.Fatalf("Can't set unix connection timeout: %v", err)
+	}
+	_, _, err = ucw.WriteMsgUnix(nil, rights, nil)
+	if err != nil {
+		t.Fatalf("UnixConn readMsg: %v", err)
+	}
+	err = ucr.SetReadDeadline(time.Now().Add(5 * time.Second))
+	if err != nil {
+		t.Fatalf("Can't set unix connection timeout: %v", err)
+	}
+	_, oobn, _, _, err := ucr.ReadMsgUnix(nil, oob)
+	if err != nil {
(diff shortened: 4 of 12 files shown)


src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestUnixConnReadMsgUnixSCMRightsCloseOnExec(t *testing.T) {
+	if !testableNetwork("unix") {
+		t.Skip("not unix system")
+	}
+
+	fds, err := syscall.Socketpair(syscall.AF_LOCAL, syscall.SOCK_STREAM, 0)
+	if err != nil {
+		t.Fatalf("Socketpair: %v", err)
+	}
+	writeFile := os.NewFile(uintptr(fds[0]), "parent-reads")
```
rittneje, line 25: You forgot to defer closing `writeFile`.
Also, you accidentally named `writeFile` as "parent-reads". That being said, the notion of "parent" here is a holdover from the original syscall unit test that involved a parent process and a child process, so perhaps these should just be named "write-socket" and "read-socket"?

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Fatalf("unexpected FileConn type; expected UnixConn, got %T", c)
+	}
+
+	buf := make([]byte, 32) // expect 1 byte
+	oob := make([]byte, 32) // expect 24 bytes
+	err = uc.SetReadDeadline(5 * time.Second)
+	if err != nil {
+		t.Fatalf("Can't set unix connection timeout: %v", err)
+	}
+	_, oobn, _, _, err := uc.ReadMsgUnix(buf, oob)
```
rittneje, line 46: You still need to write something to the connection first (see point 4 [here](https://github.com/golang/go/issues/42765#issuecomment-744038431)).

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Fatalf("FileConn: %v", err)
+	}
+	defer c.Close()
+
+	uc, ok := c.(*UnixConn)
+	if !ok {
+		t.Fatalf("unexpected FileConn type; expected UnixConn, got %T", c)
+	}
+
+	buf := make([]byte, 32) // expect 1 byte
```
rittneje, line 40: This variable is not needed. You can just pass `nil` below.

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,75 @@
...
+	oob := make([]byte, 32) // expect 24 bytes
```
rittneje, line 41: The size should be `syscall.CmsgSpace(4)` (because one file descriptor is 4 bytes). You should also remove/update the comment.

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	scm := scms[0]
+	gotFds, err := syscall.ParseUnixRights(&scm)
+	if err != nil {
+		t.Fatalf("syscall.ParseUnixRights: %v", err)
+	}
+	if len(gotFds) != 1 {
+		t.Fatalf("wanted 1 fd; got %#v", gotFds)
+	}
+
```
rittneje, line 66: For completeness, you should defer closing `gotFds[0]`.

src/net/unixsock_readmsg_test.go
HowJMay: Got it. I have updated a fix for this, but I am still thinking how should I write the file of `os.Open(os.DevNull)` to `writeFile`

HowJMay: done

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"testing"
+	"time"
+)
+
+func TestUnixConnReadMsgUnixSCMRightsCloseOnExec(t *testing.T) {
+	if !testableNetwork("unix") {
+		t.Skip("not unix system")
+	}
+
+	scmFile, err := os.Open(os.DevNull)
```
rittneje, line 22: You should defer closing this file as well.

HowJMay: done!!

rittneje:
Looks good, thanks for your help on this! Also be sure to run the unit tests if you haven't already. https://golang.org/doc/contribute.html#testing

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,106 @@
...
+	if !ok {
+		t.Fatalf("unexpected FileConn type; expected UnixConn, got %T", cw)
+	}
+	ucr, ok := cr.(*UnixConn)
+	if !ok {
+		t.Fatalf("unexpected FileConn type; expected UnixConn, got %T", cr)
+	}
+
+	oob := make([]byte, syscall.CmsgSpace(4))
+	err = ucw.SetWriteDeadline(5 * time.Second)
```
rittneje, line 60: This needs to be `time.Now().Add(5 * time.Second)`. Same for `SetReadDeadline` below.

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		t.Fatalf("Can't set unix connection timeout: %v", err)
+	}
+	_, _, _, _, err = ucw.WriteMsgUnix(nil, rights, nil)
```
rittneje, line 64: Should be `_, _, err = ...`

rittneje:
@HowJMay Currently the test `TestSCMCredentials` in syscall/creds_test.go is failing on line 107 because it expects the flags returned by `ReadMsgUnix` to be `0`. However, due to what I believe to be a Linux bug, it appears that the kernel does not clear the flags that it receives in the call to `recvmsg`, and the same bitmap is used for input flags and output flags, so it effectively is _returning_ `MSG_CMSG_CLOEXEC`. Thankfully, syscall/creds_test.go is only compiled for Linux anyway, so you can just change it to fail only if `flags != syscall.MSG_CMSG_CLOEXEC`.

HowJMay:
> MSG_CMSG_CLOEXEC

Thank you for helping!!! Everything you did help me so much!

rittneje:
@HowJMay Just wanted to follow up with you on those last pending items.

HowJMay:
> @HowJMay Just wanted to follow up with you on those last pending items.

@rittneje I think I have finished most of them

rittneje:
@HowJMay Sorry for the confusion - I forgot to click the "Submit review" button.

HowJMay:
> @HowJMay Sorry for the confusion - I forgot to click the "Submit review" button.

haha that's why XD
I have fixed the corresponding reviews. Plz take a look

src/net/unixsock_readmsg_linux.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// go:build linux
```
rittneje, line 5: This is unnecessary, since the filename already ends in _linux.

src/net/unixsock_readmsg_other.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2020 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// go:build js,wasm windows
```
rittneje, line 5: It's hard to find any references around this (it's not even mentioned in the in-progress 1.17 release notes), but assuming this works like the other directives, there should not be a space between `//` and `go`. As currently written I think this directive is being entirely ignored by the compiler.

src/net/unixsock_readmsg_posix.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// go:build aix darwin dragonfly freebsd netbsd openbsd solaris
```
rittneje, line 5:
```suggestion
//go:build aix darwin dragonfly freebsd netbsd openbsd solaris
```

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,105 @@
+// Copyright 2020 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// go:build aix darwin dragonfly freebsd linux netbsd openbsd solaris
```
rittneje, line 5:
```suggestion
//go:build aix darwin dragonfly freebsd linux netbsd openbsd solaris
```

src/net/unixsock_readmsg_test.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		t.Fatalf("file open: %v", err)
+	}
+	defer scmFile.Close()
+
+	rights := syscall.UnixRights(int(scmFile.Fd()))
+	fds, err := syscall.Socketpair(syscall.AF_LOCAL, syscall.SOCK_STREAM, 0)
+	if err != nil {
+		t.Fatalf("Socketpair: %v", err)
+	}
```
rittneje, line 31: @tklauser I don't have to ability to leave comments in Gerrit, but you left the following comment:
> We need to defer close the sockets here, otherwise we might leak file descriptors:
  defer syscall.Close(fds[0])
  defer syscall.Close(fds[1])

This is actually NOT true, and it would be a major bug to add it. This is because `os.NewFile` just wraps the existing fd instead of duplicating it so you cannot do both `syscall.Close(fds[0])` and `writeFile.Close()`, as it would try to close the same fd twice. Likewise for `fds[1]` and `readFile`.

In short, @HowJMay please do NOT apply the aforementioned suggestion.

tklauser: We're using Gerrit for code reviews, so please leave your future comments regarding the CL there. See https://golang.org/doc/contribute for more information on how to create and account.

I've replied to the comment in question.

PR #42808, opened 2020-11-24
encoding/xml: prevent multiple XMLNS attributes

### LINKED ISSUE #42807: encoding/xml: encoding a start token with a namespace set in Name and Attr results in invalid XML

<pre>
$ go version
go version devel +6965b01ea248 20201124070706 linux/amd64
</pre>

Yes

### What did you do?

https://play.golang.org/p/ZK8yhFoVwW9

```
	e := xml.NewEncoder(os.Stdout)
	e.EncodeToken(xml.StartElement{
		Name: xml.Name{Space: "space", Local: "local"},
		Attr: []xml.Attr{{Name: xml.Name{Local: "xmlns"}, Value: "spaceAttr"}},
	})
	e.Flush()
	// <local xmlns="space" xmlns="spaceAttr">
```

One "xmlns" attribute.

### What did you see instead?

Two "xmlns" attributes.

While the provided example is obviously very contrived this can happen in the real world (without a user doing something silly like manually adding a second xmlns attribute) in the situation where the user decodes some XML, makes modifications to the tokens, then encodes those tokens somewhere else (probably a fairly common use case). The decoder will add the "xmlns" attribute and fill out Name, then the encoder will duplicate it. The user would have to manually remove xmlns before encoding, or use `RawToken` (but this is not an option if the decoder is an arbitrary `xml.TokenReader` so this isn't always possible). Skipping xmlns attributes on marshal if `start.Name` is set is a simple patch that would fix this issue.

SamWhited:
Having multiple attributes of the same name is a violation of the XML
well-formedness rules. This patch does not fix every possible way this
could happen, but it does prevent marshaling from introducing such an
issue when a start token is encoded that contains both a Name attribute
with a namespace set and an xmlns attribute.

Fixes #42807

### DIFF
--- a/src/encoding/xml/marshal_test.go
+++ b/src/encoding/xml/marshal_test.go
@@ -326,6 +326,20 @@ func (m *MyMarshalerTest) MarshalXML(e *Encoder, start StartElement) error {
 	return nil
 }
 
+type MarshalMultipleXMLNS struct{}
+
+func (MarshalMultipleXMLNS) MarshalXML(e *Encoder, _ StartElement) error {
+	start := StartElement{
+		Name: Name{Space: "space", Local: "MarshalMultipleXMLNS"},
+		Attr: []Attr{{Name: Name{Local: "xmlns"}, Value: "attrSpace"}},
+	}
+	err := e.EncodeToken(start)
+	if err != nil {
+		return err
+	}
+	return e.EncodeToken(start.End())
+}
+
 type MyMarshalerAttrTest struct {
 }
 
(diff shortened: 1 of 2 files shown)


src/encoding/xml/marshal_test.go
```diff
@@ -2144,7 +2162,7 @@ var encodeTokenTests = []struct {
 			{Name{"space", "a"}, "value"},
 		}},
 	},
-	want: `<foo xmlns="space" xmlns="space"><foo xmlns="space" xmlns:_xmlns="xmlns" _xmlns:y="space" xmlns:space="space" space:a="value">`,
```
SamWhited, line 2147: I was a little concerned that this was intentional behavior when I saw that existing tests had this, but couldn't see why this would be the case so I went ahead and "fixed" them hoping this was just an oversight while copy/pasting some output for these individual tests.

SamWhited:
Hi, could I get an explanation for why this was closed without review? Thank you.

heschi:
I closed old PRs to reduce load on the Gerrit importer (#50197), sorry for the trouble. I'll reopen the CL.

PR #42840, opened 2020-11-26
net/http/fcgi: eliminate race, keep request id until end of stdin 

hillu:
There was a race condition that could lead to child.serveRequest
removing the request ID before child.handleRequest had read the empty
FCGI_STDIN message that indicates end-of-stream which in turn could
lead to child.serveRequest blocking while trying to consume the
request body.

Now, we remove the request ID from within child.handleRequest after
the end of stdin has been detected, eliminating the race condition.

Since there are no more concurrent modifications/accesses
to child.requests, we remove the accompanying sync.Mutex.

### DIFF
--- a/src/net/http/fcgi/child.go
+++ b/src/net/http/fcgi/child.go
@@ -16,7 +16,6 @@ import (
 	"net/http/cgi"
 	"os"
 	"strings"
-	"sync"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


hillu:
A more detailed description (including reproducer) can be found in <https://github.com/hillu/go-fcgi-breakage>.

PR #42852, opened 2020-11-27
Adjust the code position and add judgment comments

kingljl:
Code: can I put it in ( t.altMu.RLock() )forward judgment
```
if s := req.URL.Scheme; s != "http" && s != "https" {
		req.closeBody()
		return nil, &badStringError{"unsupported protocol scheme", s}
	}
	if req.Method != "" && !validMethod(req.Method) {
		return nil, fmt.Errorf("net/http: invalid method %q", req.Method)
	}
	if req.URL.Host == "" {
		req.closeBody()
		return nil, errors.New("http: no Host in request URL")
	}
```


kingljl:
The code is too old

PR #43290, opened 2020-12-20
crypto/elliptic: extend Curve to support general elliptic curve y² = x³ + ax + b

dustinxie:
Curve implements the short-form Weierstrass curve y² = x³ + ax + b, but
only with a = -3, which are the case for NIST-recommended curves P224,
P256,P384, and P521. For a general curve with a != -3, one would have to
rely on external packages, which is quite an inconvenience. For example,
such a popular curve is secp256k1 with equation y² = x³ + 7.

A careful examination reveals that 'a' is only needed in computing the point
scalar multipilication kP, but not needed for the point addition P + Q. To be
more specific, algorithm addJacobian() keeps the same regardless of the
value of a, only doubleJacobian() needs to be adapted to handle the case
a != -3, and the required change is relatively small.
    
The first commit provides a pure-Go implementation for the general elliptic
curve that removes the restriction a = -3 by:
1. adding an A *big.Int to CurveParams
2. adapting doubleJacobian() to be able to handle a general 'a' value

The second commit implements the curve y² = x³ + 7 known as secp256k1,
as defined in (https://www.secg.org/sec2-v2.pdf, section 2.4.1)


dustinxie:
@googlebot I signed it!

PR #43366, opened 2020-12-24
Add 'any' type as shorthand to interface{} for better code readability

Thor-x86:
It would be better if we allow programmers to use "any" instead of "interface{}" because it's quicker to type and has better common sense. For your information, this changes inspired from TypeScript which has "any" type, similar with golang's "interface{}".

### DIFF
--- a/src/builtin/builtin.go
+++ b/src/builtin/builtin.go
@@ -10,6 +10,9 @@
 */
 package builtin
 
+// any is an universal type, shorthand of interface{}
+type any interface{}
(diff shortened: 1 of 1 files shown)


Thor-x86:
@googlebot I signed it!

mdlayher:
The built-in package only exists for documentation and this is not the way to start a discussion about such a thing. The proposal process should be followed, and I suspect this has come up before. I'm going to close this.

PR #43372, opened 2020-12-25
net/smtp: Add 'From' header to smtp/example_test.go

MilesLin:
I was following the `ExampleSendMail` example to send mails, but it showed the error of `send mail error: 550 MIME message is missing 'From' header`. And I see the document of `smtp.SendMail`, it says 
> ... The msg headers should usually include fields such as "From", "To", "Subject", ...

And it works for me after I added the 'From' header in the msg.

Note: The SMTP server I used is SendGrid.


MilesLin:
@googlebot I signed it!

PR #43412, opened 2020-12-29
Package hash 汉化完成

guodawn:


agnivade:
This PR just changes comments to Chinese. It is unlikely we will accept PRs like these. However, if you have a proposal to discuss, feel free to create an issue first.

PR #43448, opened 2020-12-31
net/http: fix 302 empty location header error

st3rv04ka:
We need to add a Location check for other redirection codes, otherwise in some cases it causes a fatal error.

Example:
```
package main

import (
	"fmt"
	"io/ioutil"
	"net/http"
)

func main() {
	/**
	index.php
		<?php
		http_response_code(302);
		?>
		<!DOCTYPE html>
		<html>
			<body>
				body
			</body>
		</html>
	**/
	resp, _ := http.Get("http://example.com")
	defer resp.Body.Close()
	body, _ := ioutil.ReadAll(resp.Body)
	fmt.Println(string(body))
}
```

And error:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x564a6681987e]

goroutine 1 [running]:
main.main
        /home/test2.go:23
exit status 2
```

### DIFF
--- a/src/net/http/client.go
+++ b/src/net/http/client.go
@@ -500,7 +500,13 @@ func redirectBehavior(reqMethod string, resp *Response, ireq *Request) (redirect
 		redirectMethod = reqMethod
 		shouldRedirect = true
 		includeBody = false
-
+		if resp.Header.Get("Location") == "" {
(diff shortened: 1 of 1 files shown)


bimkos:
@googlebot I signed it!

PR #43491, opened 2021-01-04
strings: add StripSlashes function for strings pkg

LockGit:
Add a StripSlashes function , Un-quotes a quoted string.

Description：
Returns a string with backslashes stripped off. (\' becomes ' and so on.)
Double backslashes (\\) are made into a single backslash (\).


LockGit:
error

PR #43533, opened 2021-01-06
syscall/plan9: remove spooky fd action at a distance

### LINKED ISSUE #43524: os/exec: program hangs if it reads from pipe on Plan 9 (9front)

<pre>
$ go version
go version devel +9eef49cfa6eb016e3b20df189e540c6c5a71f365 2021-01-04 17:59:30 +0000 +0000 plan9/amd64
</pre>

Yes

amd64/plan9 (only 9front)

### What did you do?

Run this program:
```Go
package main

import (
	"bytes"
	"os"
	"os/exec"
)

func main() {
	var buf bytes.Buffer
	buf.WriteString("hello world\n")

	cmd := exec.Command("/bin/cat")
	cmd.Stdin = &buf
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
	if err := cmd.Run(); err != nil {
		panic(err)
	}
}
```

No hang

### What did you see instead?

Program hangs trying to read from stdin, and also the parent process hangs trying to read error string from the child. Speaking to 9front devs, this seems to be due to a [recent change](https://code.9front.org/hg/plan9front/rev/e498040721f9) in how OCEXEC flag works. The OCEXEC flag semantics have been changed to refer to the file descriptor instead of the channel.

Note: you can only reproduce this when stdin is a pipe. It doesn't hang if the stdin is a disk file. So, there are two pipes involved here. One is [created](https://github.com/golang/go/blob/9eef49cfa6eb016e3b20df189e540c6c5a71f365/src/os/exec/exec.go#L258) by `os/exec` to write to program stdin, and the other pipe is used by the `syscall` package to [read error string](https://github.com/golang/go/blob/9eef49cfa6eb016e3b20df189e540c6c5a71f365/src/syscall/exec_plan9.go#L435) from the child. The write end of the pipe created by syscall is [marked closed-on-exec](https://github.com/golang/go/blob/9eef49cfa6eb016e3b20df189e540c6c5a71f365/src/syscall/exec_plan9.go#L323). Since the close-on-exec is no longer set correctly on 9front, we hang trying to read the error string from the child. Write to stdin pipe also likely hangs due to this.

It's possible to add a workaround which works on both 9front and 9legacy.

@gopherbot Add labels OS-Plan9, NeedsFix

oridb:
Change Plan 9 fork/exec to use the O_CLOEXEC file
descriptor, instead of relying on spooky at a
distance.

Historically, Plan 9 has set the O_CLOEXEC flag on
the underlying channels in the kernel, rather
than the file descriptors -- if two fds pointed
at a single channel, as with dup, changing the
flags on one of them would be observable on the
other.

The per-Chan semantics are ok, if unexpected,
when a chan is only handled within a single
process, but this isn't always the case.

Forked processes share Chans, but even more of
a problem is the interaction between /srv and
OCEXEC, which can lead to unexectedly closed
file descriptors in completely unrelated
proceses. For example:

	func exists() bool {
		// If some other thread execs here,
		// we don't want to leak the fd, so
		// open it O_CLOEXEC
		fd := Open("/srv/foo", O_CLOEXEC)
		if fd != -1 {
			Close(fd)
			return true
		}
		return false
	}

would close the connection to any file descriptor
(maybe even for the root fs) in ALL other processes
that have it open if an exec were to happen(!),
which is quite undesriable.

As a result, 9front will be changing this behavior
for the next release.

Go is the only code observed so far that relies on
this behavior on purpose, and  It's easy to make the
code work with both semantics: simply using the file
descriptor that was opened with O_CEXEC instead of
throwing it away.

So we do that here.

Fixes #43524

### DIFF
--- a/src/syscall/exec_plan9.go
+++ b/src/syscall/exec_plan9.go
@@ -320,14 +320,15 @@ func cexecPipe(p []int) error {
 		return e
 	}
 
-	fd, e := Open("#d/"+itoa(p[1]), O_CLOEXEC)
+	fd, e := Open("#d/"+itoa(p[1]), O_RDWR|O_CLOEXEC)
 	if e != nil {
 		Close(p[0])
 		Close(p[1])
 		return e
 	}
 
-	Close(fd)
+	Close(p[1])
+	p[1] = fd
 	return nil
 }
 


oridb:
@googlebot I signed it!

PR #43550, opened 2021-01-06
math/bits: Left and right bit shifter for byte slices

pschou:
While working with hashing tools, I found there is a lack of good shifting tools to operate on arbitrary byte slices.  This allows for any slice to be shifted 0 to N number of bits without copying the whole array.

For #43552


mvdan:
https://github.com/golang/go/issues/43552 was declined; closing.

PR #43561, opened 2021-01-06
net/http: fix typo in comment

toasa:
Enclose the example words (`gzip` and `deflate`) in double quotes to represent a Go string type.

### DIFF
--- a/src/net/http/request.go
+++ b/src/net/http/request.go
@@ -145,7 +145,7 @@ type Request struct {
 	// then
 	//
 	//	Header = map[string][]string{
-	//		"Accept-Encoding": {"gzip, deflate"},
+	//		"Accept-Encoding": {"gzip", "deflate"},
 	//		"Accept-Language": {"en-us"},
 	//		"Foo": {"Bar", "two"},
 	//	}


andrius4669:
Is that actually a typo? Old will produce one header with comma, new will produce two headers.

Toasa:
> Is that actually a typo? 

I'm wrong and you are correct.

To confirm my misunderstand, I wrote following test code which parses a test HTTP header ([playground](https://play.golang.org/p/SqOoNd01hI-)):

```go
func main() {
	headerStr := "accept-encoding: gzip, deflate\r\n"

	reader := bufio.NewReader(strings.NewReader(headerStr + "\r\n"))
	tp := textproto.NewReader(reader)

	mimeHeader, err := tp.ReadMIMEHeader()
	if err != nil {
		panic(err)
	}

	httpHeader := http.Header(mimeHeader)
	
	for key, val := range httpHeader {
		fmt.Println("key:", key)                   // => key: Accept-Encoding
		fmt.Println("val:", val, "len:", len(val)) // => val: [gzip, deflate] len: 1
	}                                                                             ~~~~~~
}
```

PR #43618, opened 2021-01-10
cmd/compile/internal/gc: prealloc map space

cuiweixie:


cuiweixie:
why it can be imported to gerrit?

PR #43713, opened 2021-01-15
src/strings: strings.go -Enhance ContainsIn,EqualsIn to support optional strlist

ralic:
Enhance strings library to support contains of optional string list 

 ex. 
Original syntax :
strings.ContainsIn("abcdef","aa") ||
return strings.ContainsIn("abcdef","bb") ||
strings.ContainsIn("abcdef","cc") ||
strings.ContainsIn("abcdef","d")  => true

New Syntax :
strings.ContainsIn("abcdef", "aa","bb","cc","d")  => true
strings.ContainsIn("abcdef", "aa","bb","cc","dd")  => false

### DIFF
--- a/src/strings/strings.go
+++ b/src/strings/strings.go
@@ -65,6 +65,16 @@ func ContainsIn(str string, strlist ...string) bool {
 	}
 	return false
 }
+// Similary to ContainsIn, exactly equalto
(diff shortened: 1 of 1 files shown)


mvdan:
The CL was abandoned.

PR #43742, opened 2021-01-16
crypto/ed25519: This change modifies Go to clarify docs for PrivateKey.Sign()

billallen256:
Updated the docs for PrivateKey.Sign() to clarify that the rand parameter is unused. No functional change.


zamicol:
Why not just "The rand parameter is ignored."?

billallen256:
I thought it would be helpful to provide a quick, reasonable default.  I also didn't want the reader to have to dig into the underlying type of `crypto.SignerOpts` to see if `nil` would be acceptable.

PR #43799, opened 2021-01-20
io/ioutil: flag package and functions with the Deprecated marker

dlsniper:
All implementations are now part of io or os packages, per #42026.
Flag all implementations in ioutil, and the package itself, with
the Deprecated marker so that tooling can pick it up and interpret
it accordingly.

Updates #42026


jeffwidman:
@dlsniper do you want to rebase and fix merge conflicts per discussion over in https://github.com/golang/go/issues/51927#issuecomment-1078190876?

Issue #43756, opened 2021-01-18
image/color: Palette.Index can be optimized


zj360202:
### What did you do?
image/color  

```go
func (p Palette) Index(c Color) int {
	// A batch version of this computation is in image/draw/draw.go.

	cr, cg, cb, ca := c.RGBA()
	ret, bestSum := 0, uint32(1<<32-1)
	for i, v := range p {
		vr, vg, vb, va := v.RGBA()
		sum := sqDiff(cr, vr) + sqDiff(cg, vg) + sqDiff(cb, vb) + sqDiff(ca, va)
		if sum < bestSum {
			if sum == 0 {
				return i
			}
			ret, bestSum = i, sum
		}
	}
	return ret
}
```

is very slowly
image size 1366x768  --> cost > 2s

### What did you expect to see?

```go
func (p Palette) Index(c color.Color) int {
	// A batch version of this computation is in image/draw/draw.go.

	var mod uint8 = 51
	var splitNum = 255 / mod + 1
	cr, cg, cb, ca := c.RGBA()
	cr8, cg8, cb8, ca8 := uint8(cr), uint8(cg), uint8(cb), uint8(ca)
	ri, gi, bi, ai := cr8/mod, cg8/mod, cb8/mod, ca8/mod
	rm, gm, bm, am := cr8%mod, cg8%mod, cb8%mod, ca8%mod
	if rm > mod/2{
		ri += 1
	}
	if gm > mod/2 {
		gi += 1
	}
	if bm > mod/2 {
		bi += 1
	}
	if am > mod/2 {
		ai += 1
	}
	//ret := int(ri * 36 + gi * 6 + bi)
	ret := int(ri * splitNum * splitNum + gi * 6 + bi)
	return ret
}
```

will be faster
image size 1366x768  --> cost < 100ms

seankhliao:
@zj360202 please feel free to send in a PR / CL with the proposed improvements

zj360202:
@seankhliao OK, I will send a PR

nigeltao:
As I mentioned on https://golang.org/cl/287892, we can't assume that `p` (the receiver for the Palette.Index method) is the Netscale 6x6x6 color cube.

PR #43837, opened 2021-01-22
image/color: Palette.Index can be optimized

zj360202:
close [#43756](https://github.com/golang/go/issues/43756)


zj360202:
非常抱歉，第一次创建，有错误的地方，取消掉
I'm very sorry to create it for the first time. If there are errors, cancel it

PR #43900, opened 2021-01-25
[dev.boringcrypto] crypto/boring: Introduce OpenSSL port

derekparker:
This series of patches introduces OpenSSL support for the dev.boringcrypto branch.

The first CL refactors the existing codebase in anticipation for the port but without adding anything net new. The second CL introduces the OpenSSL port, and the final CL cleans up a few things with regard to the "notboring" mode of operation.

The major differences between the BoringCrypto port and the OpenSSL port are that the latter is not statically linked and instead uses `dlopen` to get OpenSSL, and also adds a "strict" mode which disabled certain APIs when adhering to a stricter FIPS standard.

---

 crypto/boring: Refactor boringcrypto codebase

This patch refactors the boringcrypto implementation in an attempt to
make it more general and abstract away the actual external crypto
backend being used.

This is in effort to enable other external crypto libraries such as
OpenSSL to be used instead.

---

 crypto/boring: Implement initial OpenSSL port

This patch introduces an initial OpenSSL port using the same interface
as the BoringCrypto code that already existed. A new tag is introduced:
'openssl' which explicitly enables OpenSSL over BoringCrypto.

Unlike the BoringCrypto implementation we do not statically link against
OpenSSL. There was a lot done in order to statically link against
BoringCrypto and some of that work severely limited arch support.
Instead we attempt to `dlopen` the OpenSSL library so that binaries can
still remain a bit more portable (e.g. moving an openshift client binary
to another machine which doesn't have OpenSSL will not brick the
binary).

The OpenSSL port has a few additions to the C layer in between Go and
<crypto library>. Some are backports from later OpenSSL versions for
consistency and some are so that we can take advantage of the EVP
interface and minimize Go <-> CGO context switching. There's still much
to be desired there, but it's a good start.


derekparker:
The OpenSSL port passes all tests with the exception of two:

```
--- FAIL: TestOverlongMessagePKCS1v15 (0.00s)
    pkcs1v15_test.go:232: RSA decrypted a message that was too long.
FAIL
--- FAIL: TestAlertFlushing (0.00s)                                                                                                                                                                                                                                                 handshake_client_test.go:2117: expected to find "remote error: tls: internal error" in error but error was "tls: invalid signature by the server certificate: crypto/rsa: verification error"                                                                               FAIL
```

The `TestAlertFlushing` test seems to fail because OpenSSL seems to not be doing the "AndCheck" portion of "decryptAndCheck" as noted in `crypto/rsa/pss.go:223`.

The `TestOverlongMEssagePKCS1v15` also seems to fail due to differences in OpenSSL and BoringCrypto with regards to how padding is checked.

I'm happy to fix these but would like a little guidance if possible on how to proceed there.

derekparker:
Also, the CLA issue seems to stem from my submitting these patches under my `@redhat.com` email. I've signed the CLA under my personal email, should I sign another or can this be resolved some other way?

derekparker:
cc @FiloSottile

skitalecs (approved):

derekparker:
Hey, I know there's a lot of work going on currently for the 1.16 release. Was just hoping to get an ack on this since my last comment from 3 weeks ago is the last movement on this CL. Even if a proper review won't happen until the 1.17 window opens, I'd just appreciate some acknowledgement so I know this isn't getting lost.

@FiloSottile @agl @rsc

derekparker:
Hey @FiloSottile I wanted to continue our discussion over mail and further some ideas to get this patch reviewed. I know it is a large patch overall, but I made efforts to contain each major change within it's own commit. Most of the changes in the first commit are just moving files around and organization.

As this is a patch porting an existing system, I'm not sure how to make this patch smaller. The C files are necessary because we must use EVP implementations of OpenSSL routines to ensure hardware acceleration and FIPS compliance. Not supporting Openssl 1.0 won't make too much of a difference in code size either.

One potential thought I had was leaving this PR open to capture the whole vision, and then possibly submitting the first commit as it's own CL which refactors to make further changes possible. Following that I could try submitting smaller, more targeting PRs for different algos such as AES, ECDSA, etc... This would mean the tests won't pass until all the functions are implemented (unless we have a fallback to Go functions, but that seems risky in case anything gets missed).

I'm open to thoughts on this, I'd really like to work together to get this merged upstream.

derekparker:
Ping @FiloSottile -- just wanted to see if I could get an update based on my comment above and additional thoughts on how to make this patch smaller / easier to review.

FiloSottile:
@derekparker breaking the change up in multiple CLs would be definitely an improvement (and for something of this size you'll probably want to skip gerritbot and use Gerrit directly) but if the final result is still thousands of lines, I still can't give you a good estimate of when I'll have the resources to do the review, sorry.

I'll try to have a look at the change soon to see if I can see ways to make it easier to land.

derekparker:
> @derekparker breaking the change up in multiple CLs would be definitely an improvement (and for something of this size you'll probably want to skip gerritbot and use Gerrit directly) but if the final result is still thousands of lines, I still can't give you a good estimate of when I'll have the resources to do the review, sorry.
> 
> I'll try to have a look at the change soon to see if I can see ways to make it easier to land.

Thanks for the response! I'll work on trying to make this PR smaller, even if it's just submitted multiple, smaller CLs but keeping this one up as a reference for the "why" of like, "why are we refactoring this?" I can then point to this commit which builds upon that. I'll also see if I can make the overall patch smaller by seeing if there's anything unused or superfluous I can take out.

derekparker:
Closing this PR out. I will be splitting this large patch set into a series of smaller, easier to review patch sets. Having this submitted in it's entirety, however, I think will be good for reference as to how the smaller patches I will be submitting fit into the bigger picture.

derekparker:
(Sorry for the delay in getting back to this, I've just returned from paternity leave following the birth of my 2nd child).

Issue #31544, opened 2019-04-18
cmd/go: spurious error message when external linking a pure Go program


cherrymui:
### What version of Go are you using (`go version`)?

tip (3235f7c0720338a160debe6e9c632b8af968b4dd)

### Does this issue reproduce with the latest release?

Yes, happens with Go 1.11, 1.12, and tip, but not with Go 1.10.

### What operating system and processor architecture are you using (`go env`)?

linux/amd64

### What did you do?

```
$ mkdir -p /tmp/src/hello
$ cat >hello.go
package main
func main() { println("hello") }
$ GO111MODULE=off GOPATH=/tmp go build -ldflags=-linkmode=external
# hello
loadinternal: cannot find runtime/cgo
$ echo $?
0
$ ./hello
hello
```

When using external linking for a pure Go program, the linker prints `loadinternal: cannot find runtime/cgo` message. The linking actually succeeded, and the generated binary works fine.

### What did you expect to see?
No error message, like with Go 1.10.
```
$ GOPATH=/tmp go1.10 build -ldflags=-linkmode=external 
$ ./hello 
hello
```

### What did you see instead?
Spurious `loadinternal: cannot find runtime/cgo` error message.

cherrymui:
This also happens in module mode:
```
$ go mod init hello
go: creating new go.mod: module hello
$ GO111MODULE=on go build -ldflags=-linkmode=external 
# hello
loadinternal: cannot find runtime/cgo
```

But it does not happen if the file name (hello.go) is given explicitly.
```
$ go build -ldflags=-linkmode=external hello.go 
$
```

bcmills:
I can reproduce the failure. I'm bisecting it now to try to get some idea of how it broke.

bcmills:
`git bisect` points to [CL 129059](https://golang.org/cl/129059).

bcmills:
This is a result of the `-ldflags=-linkmode=external` flag now (correctly) being passed when compiling the `hello` package, which reveals a typestate bug in `load.PackagesAndErrors`.

The decision in the `LinkerDeps` function about whether to add the `runtime/cgo` dependency is based on whether the `CmdlinePkg` flag is set:
https://github.com/golang/go/blob/19966e9b533a85039458eabf08228e0b85f3bf4e/src/cmd/go/internal/load/flag.go#L43

That field is set *after* the call to `loadImport`:
[...]

PR #43909, opened 2021-01-25
cmd/link: disable TestPIESize if CGO isn't enabled

nehaljwani:
With CGO disabled, the test throws the following error:
```
elf_test.go:291: # command-line-arguments
    loadinternal: cannot find runtime/cgo
```

### DIFF
--- a/src/cmd/link/elf_test.go
+++ b/src/cmd/link/elf_test.go
@@ -226,6 +226,8 @@ func main() {
 
 func TestPIESize(t *testing.T) {
 	testenv.MustHaveGoBuild(t)
+	testenv.MustHaveCGO(t)
+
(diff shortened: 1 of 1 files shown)


nehaljwani:
Uh.. the bot parsed the PR message instead of the commit and included backticks :( Closing.

cherrymui:
> Uh.. the bot parsed the PR message instead of the commit and included backticks :( Closing.

Yeah, you need to write the CL description in the first message. Thanks.

> loadinternal: cannot find runtime/cgo

FWIW, this is not actually an error. The build should complete and the output binary is probably working. (See #31544)

nehaljwani:
@cherrymui I closed this PR because it ended up adding backticks (```) in the commit message posted to gerrit at https://golang.org/cl/286553 , which I think is against the rules since markdown isn't allowed.

PR #43924, opened 2021-01-26
doc/go1.16: fix closing brace in .Export format

qt-luigi:
A parenthesis of go list "-f" flag format is double curly braces.


qt-luigi:
Thanks for merging me! q@w@p

PR #44017, opened 2021-01-30
NaN hex value fix

andybbruno:
Closes #44016


andybbruno:
@googlebot I signed it!

PR #44352, opened 2021-02-17
cmd/cgo: define exported Go func as function

### LINKED ISSUE #43830: cmd/cgo: Conflicting types for generated symbols prevent LTO-enabled compilation

<pre>
$ go version
go version go1.15.7 linux/amd64
</pre>

Yes.

$ go env
GOARCH="amd64"
GOHOSTOS="linux"
</pre></details>

### What did you do?
Enable link time optimization CFLAGS when building a Go program containing an `//export` directive.

```
package main

import "fmt"
import "C"

//export add
func add(x int32, y int32) int32 {
  return x + y
}

func main() {
  fmt.Println(add(1,1))
}
```

The above program builds normally using `go build` without additional CFLAGS.  However, building with `CGO_CFLAGS='-flto -ffat-lto-objects' go build` produces the following error.
```
_cgo_export.c:21:13: error: variable ‘_cgoexp_c166a99bca60_add’ redeclared as function
cgo-generated-wrappers:1:5: note: previously declared here
lto1: fatal error: errors during merging of translation units
compilation terminated.
lto-wrapper: fatal error: /usr/bin/gcc returned 1 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
```
This is due to the same symbol name being used in separate contexts as both a [function](https://github.com/golang/go/blob/ecf4ebf10054f70e51a0ce759b2ae91aa4febd1a/src/cmd/cgo/out.go#L980) and an [integer](https://github.com/golang/go/blob/ecf4ebf10054f70e51a0ce759b2ae91aa4febd1a/src/cmd/cgo/out.go#L1031).

Program builds when enabling link time optimization via CFLAGs.

### What did you see instead?
Generated intermediate C code for `//export` is incompatible with link time optimization.

### Notes
Additional conflicting symbols may be created if the program binds the value of an exported function's C handle to a Go variable.  The following program demonstrates the behavior.

```
package main

import "fmt"

// extern int add(int, int);
import "C"

//export add
func add(a int32, b int32) int32 {
  return a + b
}

func main() {
  fn := C.add
  fmt.Printf("%v\n", fn)
}

```

When built with `CGO_CFLAGS='-flto -ffat-lto-objects' go build`, the linker yields the following error.

```
cgo-generated-wrappers:1:13: error: function ‘add’ redeclared as variable
_cgo_export.c:24:9: note: previously declared here
_cgo_export.c:21:13: error: variable ‘_cgoexp_2526879f8d2a_add’ redeclared as function
cgo-generated-wrappers:4:5: note: previously declared here
lto1: fatal error: errors during merging of translation units
compilation terminated.
lto-wrapper: fatal error: /usr/bin/gcc returned 1 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
```

The interesting aspect to note is the redeclaration of the `add` symbol in addition to the previously encountered `_cgoexp*_add` symbol.  This comes from introducing the assignment statement of `C.add` to the variable `fn`, and causes cgo to emit an additional pair of matching symbols with conflicting types via [another code path](https://github.com/golang/go/blob/ecf4ebf10054f70e51a0ce759b2ae91aa4febd1a/src/cmd/cgo/out.go#L171).

derekparker:
In order to resolve symbols correctly during C compilation with CGO
the compiler emits a _cgo_main.c file which holds definitions for
symbols. The problem is that the definitions in the _cgo_main.c do not
match up with what's declared in _cgo_export.c which causes compilation
to fail if using CGO_CFLAGS="-flto -ffat-lto-objects" due to an
incompatibility with link time optimization in GCC.

This patch makes it so that the symbol emitted in _cgo_main.c lines up
with the type declared in _cgo_export.c. This should be safe to do as
the actual exported C symbol ends up as an argument to crosscall2 which
defines the function signature for these exported symbols. The function
sig will be consistent in that way since CGO generates wrappers and
structs to handle argument / return value passing.

Fixes #43830

### DIFF
--- a/src/cmd/cgo/gcc.go
+++ b/src/cmd/cgo/gcc.go
@@ -1614,6 +1614,8 @@ func (p *Package) gccCmd() []string {
 		c = append(c, "-maix64")
 		c = append(c, "-mcmodel=large")
 	}
+	// disable LTO so we get a real object
+	c = append(c, "-fno-lto")
 	c = append(c, "-") //read input from standard input
 	return c
 }
--- a/src/cmd/cgo/out.go
+++ b/src/cmd/cgo/out.go
@@ -168,6 +168,11 @@ func (p *Package) writeDefs() {
 			if *gccgo {
 				fmt.Fprintf(fc, "extern byte *%s;\n", n.C)
 			} else {
+				// Only emit the following symbols if they are not for
+				// a function pointer variable. The function symbol is already defined
+				// and redefining it here as such will break compilation when
+				// used in conjunction with link time optimization in external
+				// linkers.
 				if n.Kind != "fpvar" {
 					fmt.Fprintf(fm, "extern char %s[];\n", n.C)
 					fmt.Fprintf(fm, "void *_cgohack_%s = %s;\n\n", n.C, n.C)
--- a/src/cmd/cgo/out.go
+++ b/src/cmd/cgo/out.go
@@ -172,6 +172,8 @@ func (p *Package) writeDefs() {
 				// reference to all symbols Go code is referring
 				// to -- because otherwise the system linker might decline
 				// to add DT_NEEDED entries for the library that has the symbol definition.
+				// We treat function pointer types differently to avoid link-time failures
+				// when using LTO with an external linker.
 				if n.Kind == "fpvar" {
 					fmt.Fprintf(fm, "extern void %s();\n", n.C)
 				} else {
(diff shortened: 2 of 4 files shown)


derekparker:
I've added a test case for the first issue and included a second, related fix with a test case as well. Not seeing them in gerrit right now but I'm not sure how quick the sync works.

PR #44375, opened 2021-02-18
fmt: use “truncateString” not “truncate” in method doc

Code-Fight:


Code-Fight:
@googlebot I signed it!

PR #44533, opened 2021-02-23
io: add error check to WriteString Example test

Julian-Chu:


Julian-Chu:
> @googlebot I signed it!

@googlebot  I signed it!

PR #44565, opened 2021-02-24
net/http: support mulit same transfer-encoding header

Soontao:
Hello Golang STD Maintainers, 

Here is a proposal about the `net/http/transfer`, we can discuss about this. 

In my scenario,  the java spring framework will allow application response result with duplicated `transfer-encoding` header, as a result, when the `go service` call the `java app`, the `go service` will raise an error about transfer-encoding. 

so, as a result, I want to create a PR to make `go/net/http` could process response with duplicated `chunked` values in `tranfer-encoding` header.

Is that possible ?

Thanks


thinkwelltwd:
This PR would fix https://github.com/golang/go/issues/44591. It'd be very nice to have this, or some other fix merged!

PR #44604, opened 2021-02-25
all: remove trailing whitespace

sashamelentyev:


alexander-melentyev:
@googlebot I signed it!

PR #44786, opened 2021-03-04
runtime:fix wrong comment

cs-chicken:
original comment may be wrong
the dataSize is the size of new object (except allocating a defer block
and the arguments size is the actually rounded to the next malloc size class

### DIFF
--- a/src/runtime/mbitmap.go
+++ b/src/runtime/mbitmap.go
@@ -821,9 +821,11 @@ func heapBitsSetType(x, size, dataSize uintptr, typ *_type) {
 		mask3 = bitPointer | bitScan | mask2<<heapBitsShift // 01110111
 	)
 
-	// dataSize is always size rounded up to the next malloc size class,
-	// except in the case of allocating a defer block, in which case
-	// size is sizeof(_defer{}) (at least 6 words) and dataSize may be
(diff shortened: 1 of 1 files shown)


src/runtime/mbitmap.go
```diff
@@ -821,9 +821,11 @@ func heapBitsSetType(x, size, dataSize uintptr, typ *_type) {
...
 	)
 
-	// dataSize is always size rounded up to the next malloc size class,
-	// except in the case of allocating a defer block, in which case
-	// size is sizeof(_defer{}) (at least 6 words) and dataSize may be
+	// argument 'size' is always size rounded up to the next malloc size class,
+	// argument dataSize is always equal typ.size or typ.size * N (when typ is
+	// array and N is the size of array) except in the case of
+	// allocating a defer block, in which case
+	// dataSize is sizeof(_defer{}) (at least 6 words) and 'size' may be
```
Mifrill, line 828:
```suggestion
	// The argument 'size' is always the size rounded up to the next malloc size class.
	// The argument 'dataSize' is always equal to typ.size or typ.size * N
	// (when typ is an array and N is the size of the array),
	// except in the case of allocating a defer block.
	// In this case, 'dataSize' is sizeof(_defer{}) (at least 6 words), and 'size' may be...
```

PR #44857, opened 2021-03-08
cmd/compile: mention that -m can be increased or given multiple times

michael-schaller:
-m can be increased or it can be given up to 4 times to increase the verbosity of the printed optimization decisions: https://github.com/golang/go/search?q=LowerM


michael-schaller:
Should the flag description also be updated or should it be kept terse?
https://github.com/golang/go/blob/9a555fc24c318bf1b07bdc07d5c02e372681e401/src/cmd/compile/internal/base/flag.go#L72

PR #44935, opened 2021-03-11
misc/ios: change to explicitly use python3 in Xcode

tkmru:
In `misc/ios/go_ios_exec.go`, it calls python to use the llvm api.
However, since it is simply calling python, if other python is used, such as in pyenv, it will give an error like `Fatal Python error: _PyInterpreterState_Get(): no current thread state`. Since lldb library depends on the python in Xcode, we must use the python in Xcode.

So I changed the xcrun command to explicitly use the python in Xcode.


tkmru:
@googlebot I signed it!

PR #45146, opened 2021-03-21
cmd/go2go: add build tag support to the go2go command

maguro:
You can specify build tags, so that imports properly work during translation,
by using the `-tags` option, e.g. `go tool go2go run -tags=appengine x.go2`.
The `-tags` option is available for all the `go2go` sub-commands.


maguro:
Replaced w/ https://github.com/golang/go/pull/45147

PR #45155, opened 2021-03-22
net: only perform IPv4 map check for AF_INET6 sockets

jrick:
This change avoids executing syscalls testing if IPv4 address mapping
is possible unless the socket being opened belongs to the AF_INET6
family.

In a pledged OpenBSD process, this test is only allowed when the
"inet" pledge is granted; however this check was also being performed
for AF_UNIX sockets (separately permitted under the "unix" pledge),
and would cause the process to be killed by the kernel.  By avoiding
the IPv4 address mapping check until the socket is checked to be
AF_INET6, a pledged OpenBSD process using AF_UNIX sockets without the
"inet" pledge won't be killed for this misbehavior.

The OpenBSD kernel is not currently ready to support using UNIX domain
sockets with only the "unix" pledge (and without "inet"), but this is
one change necessary to support this.


jrick:
For the OpenBSD kernel patch which is also needed to try this, see https://marc.info/?l=openbsd-tech&m=161221087502321&w=2

PR #45243, opened 2021-03-25
net: never probe IPv4 map support on DragonFly BSD, OpenBSD

jrick:
DragonFly BSD and OpenBSD do not implement mapping IPv4 addresses to
the IPv6 address space, and a runtime check can be avoided.

As the IP stack capabilities probe was only being called from
supportsIPv4map to check for this support, the OS-specific handling
can be added to this function rather than continuing to run the probe.


jrick:
@sthen fyi

PR #45271, opened 2021-03-28
mime: support reading shared mime-info database on unix systems

Jacalz:
This adds support for reading the FreeDesktop Shared MIME-info Database on Unix systems, if it exists.
It should make lookups work on systems where the mime.types files are not present and
should lead to better mimetype lookup in general. If the shared mimetype database does not exist,
we will fall back to reading mime.types files in common locations.

Related to a bug on Solus bugtracker: https://dev.getsol.us/T9394
This change makes the mime package work on Solus.


Jacalz:
@googlebot I signed it!

PR #45289, opened 2021-03-30
cmd/compile/internal/noder: limit the number of goroutine

hqpko:


hqpko:
@googlebot I signed it!

PR #45421, opened 2021-04-07
runtime/map: update comment for gc/reflect

MakDon:
update comment cause gc/reflect.go has been moved to reflectdata/reflect.go

In the commit (attach below), gc/reflect.go is moved to reflectdata/reflect.go
So the  comment referring gc/reflect.go should be updated to reflectdata/reflect.go

There maybe other places that refers gc/reflect.go that should be updated.
I would work around it soon.

commit:
https://github.com/golang/go/commit/de65151e507e7b3c8e46d74f223d7c562177bedc
https://github.com/golang/go/commit/e4895ab4c0eb44de6ddc5dc8d860a827b20d2781


MakDon:
@googlebot I signed it!

PR #45500, opened 2021-04-11
net: reference the correct RFCs and sections for IP.IsPrivate

6543:
Properly cite RFC 1918 Section 3 for ipv4,
and RFC 4193 Section 8 for ipv6 comments.

Updates #29146


6543:
ref: https://tools.ietf.org/html/rfc4193#section-8

@odeke-em If you think there is a better commit msg / pull title just tell me :)

PR #45808, opened 2021-04-27
testing: document that TestMain can be used with benchmarks

### LINKED ISSUE #45764: testing: document that TestMain always runs before any benchmarks
The docs for `TestMain` say:

```
It is sometimes necessary for a test program to do extra setup
or teardown before or after testing.
```

https://golang.org/pkg/testing/#hdr-Main

But what if one needs to do extra setup before benchmarking? It appears that `TestMain` would also always run before benchmarks, but neither `TestMain` doc nor `M.Run` doc say anything about benchmarks.

JulienTant:
fixes #45764

### DIFF
--- a/src/testing/testing.go
+++ b/src/testing/testing.go
@@ -208,14 +208,14 @@
 //
 // Main
 //
-// It is sometimes necessary for a test program to do extra setup or teardown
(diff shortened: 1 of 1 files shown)


JulienTant:
@googlebot I signed it!

Issue #45940, opened 2021-05-04
cmd/cgo: cannot link (cgo) on linux/arm with `-no-pie`


jochen42:
### What version of Go are you using (`go version`)?

<pre>
$ go version

go version go1.16.3 linux/arm
</pre>

### Does this issue reproduce with the latest release?
yes

### What operating system and processor architecture are you using (`go env`)?
linux/arm

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env

GO111MODULE=""
GOARCH="arm"
GOBIN=""
GOCACHE="/home/pi/.cache/go-build"
GOENV="/home/pi/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="arm"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/pi/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/pi/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_arm"
GOVCS=""
GOVERSION="go1.16.3"
GCCGO="gccgo"
GOARM="6"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/pi/_src/somefancyapp/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -marm -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build1652971902=/tmp/go-build -gno-record-gcc-switches"
</pre></details>

### What did you do?

compiling a go-module with cgo and .cxx (swig generated) with the following command on linux/arm:

```bash
CGO_LDFLAGS="-no-pie" go build \
	-v -x \
	main.go
```

### What did you expect to see?

the cxx compiler (g++) command as a linker frontend is started **with** the `-no-pie` parameter and the **without** the `-pie` paramater

### What did you see instead?

[...]

Issue #26197, opened 2018-07-03
cmd/cgo: linux-{386,amd64}-clang and linux-arm builders failing


mundaym:
Appears to have been triggered by [CL 89655](https://golang.org/cl/89655).

linux-386-clang: https://build.golang.org/log/0bef8033f7380e36756d6ce63cdec841108010de

```
# _/workdir/go/misc/cgo/test.test
/workdir/go/pkg/tool/linux_386/link: running /usr/bin/clang failed: exit status 1
/usr/bin/ld: dynamic STT_GNU_IFUNC symbol `strcmp' with pointer equality in `/usr/bin/../lib/gcc/x86_64-linux-gnu/4.9/../../../../lib32/libc.a(strcmp.o)' can not be used when making an executable; recompile with -fPIE and relink with -pie
clang: error: linker command failed with exit code 1 (use -v to see invocation)

FAIL	_/workdir/go/misc/cgo/test [build failed]
```

linux-amd64-clang: https://build.golang.org/log/c233d2a2dd1486978f0030a3aec3a656340c5aa5

```
# _/workdir/go/misc/cgo/test.test
/workdir/go/pkg/tool/linux_amd64/link: running /usr/bin/clang failed: exit status 1
/usr/bin/ld: dynamic STT_GNU_IFUNC symbol `strcmp' with pointer equality in `/usr/bin/../lib/gcc/x86_64-linux-gnu/4.9/../../../x86_64-linux-gnu/libc.a(strcmp.o)' can not be used when making an executable; recompile with -fPIE and relink with -pie
clang: error: linker command failed with exit code 1 (use -v to see invocation)

FAIL	_/workdir/go/misc/cgo/test [build failed]
```

linux-arm: https://build.golang.org/log/d34739a782a08aed7a8ca08f130d35a7ec2767cb

```
# _/workdir/go/misc/cgo/test
/usr/bin/ld: /usr/lib/gcc/arm-linux-gnueabihf/5/crtbeginT.o: relocation R_ARM_THM_MOVW_ABS_NC against `a local symbol' can not be used when making a shared object; recompile with -fPIC
/usr/lib/gcc/arm-linux-gnueabihf/5/crtbeginT.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
FAIL	_/workdir/go/misc/cgo/test [build failed]
```

ianlancetaylor:
[...]

PR #45989, opened 2021-05-06
cmd/cgo: remove hardcoded '-pie' ldflag for linux/arm

jochen42:
a minimally invasive fix proposal for #45940. which keeps the fix for #26197.

an alternative for (#26197) could be to fail if we have both flags. adding/removing a flag without an message to the user is inconvenient.


jochen42:
@googlebot I fixed it

PR #46036, opened 2021-05-07
crypto/tls: use a cache to save x509 certificate in the tls connection

### LINKED ISSUE #46035: crypto/tls: same x509 certificate in different tls connections costs too much memory

<pre>
$ go version
go version go1.16.3 linux/amd64

</pre>

Yes

$ go env

</pre></details>

### What did you do?

In our production environment, we use tls to communicate between applications. A client application will access dozens to thousands of server applications, and server applications will have dozens to thousands of machines.
Each machine will maintain one or two long TCP connections for communication. A client may have tens of thousands of tls connections, and each connection will save a copy of the server's x509 certificate information, resulting in a large amount of memory usage.
For the same server application, the certificates held are the same. The actual number of different certificates is much lower than the number of connections, but the same size of memory is still allocated for each connection in the memory.
For the mtls(Mutual TLS) scenario, the server's tls connection may also save the client's certificate, which will cause the same memory consumption, and the number of client certificates is also much lower than the number of client connections.

The same x509 certificate is reused through caching, and memory is no longer calculated and allocated separately for each connection, which reduces the memory usage.

### What did you see instead?

Each connection will parse and save the x509 certificate information, which takes up a lot of memory.

nejisama:
Fixes #46035

### DIFF
--- a/src/crypto/tls/cache.go
+++ b/src/crypto/tls/cache.go
@@ -1,0 +1,44 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package tls
+
+import (
+	"crypto/x509"
+	"sync"
+)
+
+// certificateCache is a x509.Certificate cache indexed by ASN.1 DER data.
+type certificateCache struct {
+	rwmutex sync.RWMutex
+	cache   map[string]*x509.Certificate
+}
+
+var certCache = certificateCache{
+	cache: make(map[string]*x509.Certificate),
+}
+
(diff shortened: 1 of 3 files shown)


nejisama:
### Test Case 
#### test data 
+ Generate rootca and intermediate CA based on openssl, and then use go to write tool programs to generate 2000 sets of certificates/private keys in advance
#### test program and test 
+ Wrote a program that can be used to create a tls connection between the client and the server
+ The server will randomly select a certificate for the creation of tls connections, and the client will create a total of 20,000 tls connections
+ See the attachment for the complete program and certificate data
[test.tar.gz](https://github.com/golang/go/files/6440201/test.tar.gz)
### Test Details

PR #46149, opened 2021-05-13
net: add tcp WriteTo func to enable splice socket data to file

jim3ma:
When transfer data from tcp socket to file, splice will help us with low time cost in linux.

To do this in linux, there are some changes:
1. add LimitWriter to detect writing limit in internal package.
2. add spliceW to splice data from tcp socket to file.
3. add WriteTo for TCPConn.

### DIFF
--- a/src/go/build/deps_test.go
+++ b/src/go/build/deps_test.go
@@ -102,7 +102,9 @@ var depsRules = `
 	< container/heap;
 
 	RUNTIME
-	< io;
+	< io
+	< internal/iointernal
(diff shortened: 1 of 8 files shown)


jim3ma:
@ianlancetaylor Can you help review this PR ?

ianlancetaylor:
@jim3ma Thanks.  I posted a couple of comments.

Going forward, as it says above, please don't reply here, please instead reply at https://golang.org/cl/319593.  Thanks.

PR #46197, opened 2021-05-17
BigEndian add int16 to process negative value

April3xx:
I ran into a little bug that BigEndian doesn't support negative value, since it only has uint type parameter so I just add int16 to it.


April3xx:
@googlebot I signed it!

PR #46276, opened 2021-05-19
crypto/x509: add new FreeBSD 12.2+ trusted certificate folder

### LINKED ISSUE #46284: crypto/x509: add new FreeBSD root store path [freeze exception]
FreeBSD added a system root store that doesn't require installing a package. We should use that when it's available.

https://svnweb.freebsd.org/base?view=revision&revision=357082

There is a CL available already: https://golang.org/cl/321190

@golang/release, can I get a freeze exception for this? It's a very small change, platform specific, and lets us be more in sync with the trust policy of the platform.

lapo-luchini:
Up to FreeBSD 12.1 the package ca_root_nss was needed in order to have
certificates under /usr/local/share/certs as the base system didn't have
a system trusted certificate store.

This has been fixed in FreeBSD 12.2 using /etc/ssl/certs:
https://svnweb.freebsd.org/base?view=revision&revision=357082

Fixes #46284

### DIFF
--- a/src/crypto/x509/root_bsd.go
+++ b/src/crypto/x509/root_bsd.go
@@ -18,6 +18,7 @@ var certFiles = []string{
 // Possible directories with certificate files; stop after successfully
 // reading at least one file from a directory.
 var certDirectories = []string{
+	"/etc/ssl/certs",         // FreeBSD 12.2+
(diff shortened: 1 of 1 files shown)


lapo-luchini:
@googlebot I signed it!

PR #46368, opened 2021-05-25
runtime/proc: update forcePreemptNS magic number

helios741:
10\*1000\*1000 is still used in some places, without forcePreemptNS, using 10\*1000\*1000 will increase the difficulty of understanding.


helios741 (commented):
I think 10\*1000\*1000 here is forcePreemptNS. For example, **pd.syscallwhen + forcePreemptNS> now** means that if the syscall runs for more than 10ms (forcePreemptNS), it can be preempted.

PR #46421, opened 2021-05-27
net/rpc: reduce the execution of reflect and decode

lcr2000:


Mifrill (commented):
The safety of this change hinges on whether the subsequent code relies on the original `sname` value before the conditional assignment. In this specific instance, it does indeed depend on the previous `sname`. Consequently, I hold reservations about this modification and anticipate that it may result in failed tests on CI pipeline.

PR #46455, opened 2021-05-30
image: add ListFormats

MagicalTux:
This allows access to the list of currently registered format names.

This is useful for example when an image is available in multiple
formats to not have to open each format in order to find out which one
we are able to decode.


MagicalTux:
The len() check is there to prevent allocating an empty slice (`make([]string, 0)`) and instead return a nil slice, which makes more sense in case there is nothing to return.

PR #46548, opened 2021-06-03
syscall: update syscall numbers for linux/amd64

fntlnz:
Updated the syscall definitions for arm64 and amd64 to allow programs using Go to interact with Linux syscalls to have the proper constant for them and not having to figure out the number themselves.


fntlnz:
I followed the instructions and used the script to generated them using the latest mainline kernel. I didn't update the file manually.

PR #46604, opened 2021-06-07
cmd/go/internal/work: fix typo in comment

HeGaoYuan:


HeGaoYuan:
I got your reply, I will close the wrong pr, sorry for that

PR #46714, opened 2021-06-12
net/http: add MaxConnLifespan to Transport

### LINKED ISSUE #23427: http client and http.Transport IdleConnTimeout is not reliably doing DNS resolve

go version go1.9 linux/amd64

and 

go version go1.9.2 linux/amd64

Yes

% go env                                                             
GOARCH="amd64"
GOBIN="/home/sszuecs/go/bin"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/sszuecs/go"
GORACE=""
GOROOT="/usr/share/go"
GOTOOLDIR="/usr/share/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build505089582=/tmp/go-build -gno-record-gcc-switches"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"

### What did you do?

I am running `go run main.go` and change the /etc/hosts to change www.google.de to 127.0.0.1.
Depending on  the order it fails to switch DNS after IdleConnTimeout or it will never request DNS again. DNS will be tried to lookup in case you first target it to 127.0.0.1 and after that comment the entry in /etc/hosts. The problem is that if you want to change your target loadbalancers via DNS lookup, this will not be done. The workaround is commented in the code, which reliably will do the DNS failover. The problem seems to be that IdleConnTimeout is bigger then the time.Sleep duration in the code, which you can also change to see that this works. In case of being an edge proxy with high number of requests, the case IdleConnTimeout < process-next-request will never happen.

    package main
    
    import (
    	"log"
    	"net"
    	"net/http"
    	"time"
    )
    
    func main() {
    
    	tr := &http.Transport{
    		DialContext: (&net.Dialer{
    			Timeout:   5 * time.Second,
    			KeepAlive: 30 * time.Second,
    			DualStack: true,
    		}).DialContext,
    		TLSHandshakeTimeout: 5 * time.Second,
    		IdleConnTimeout:     5 * time.Second,
    	}
    
    	go func(rt http.RoundTripper) {
    		for {
    			time.Sleep(1 * time.Second)
    
    			req, err := http.NewRequest("GET", "https://www.google.de/", nil)
    			if err != nil {
    				log.Printf("Failed to do request: %v", err)
    				continue
    			}
    
    			resp, err := rt.RoundTrip(req)
    			if err != nil {
    				log.Printf("Failed to do roundtrip: %v", err)
    				continue
    			}
    			//io.Copy(ioutil.Discard, resp.Body)
    			resp.Body.Close()
    			log.Printf("resp status: %v", resp.Status)
    		}
    	}(tr)
    
    	// FIX:
    	// go func(transport *http.Transport) {
    	// 	for {
    	// 		time.Sleep(3 * time.Second)
    	// 		transport.CloseIdleConnections()
    	// 	}
    	// }(tr)
    
    	ch := make(chan struct{})
    	<-ch
    }

I want to to see that IdleConnTimeout will reliably close idle connections, such that DNS is queried for new connections again, similar to the goroutine case in the code. We need to slowly being able to fade traffic.

### What did you see instead?

In case you start the application with /etc/hosts entry is not set, and then change it, it will never fail the request, so the new DNS lookup is not being made.

josephcopenhaver:
Existing implementation appeared to have no way to set a connection
max lifetime. A max lifetime allows for refreshing connection lifecycle
concerns such as dns resolutions for those that need it.

When initialized to a non-zero value the connection will be closed after
the duration has passed. This change is backwards compatible.

Fixes #23427
---

The only other knob I would like to add to the default transport is an assurance that at most MaxConns # of connections will be opened but I think I can use the round-tripper interface to add this knob as long as the net.Conn interface is always guaranteed to have Close() called.

### DIFF
--- a/api/next.txt
+++ b/api/next.txt
@@ -1,0 +1,1 @@
+pkg net/http, type Transport struct, MaxConnLifespan time.Duration
--- a/src/net/http/h2_bundle.go
+++ b/src/net/http/h2_bundle.go
@@ -6999,6 +6999,7 @@ type http2ClientConn struct {
 	pendingRequests int                       // requests blocked and waiting to be sent because len(streams) == maxConcurrentStreams
 	pings           map[[8]byte]chan struct{} // in flight ping data to notification channel
 	br              *bufio.Reader
+	reuseDeadline   time.Time
 	lastActive      time.Time
 	lastIdle        time.Time // time last idle
 	// Settings from peer: (also guarded by wmu)
@@ -7388,9 +7389,20 @@ func (t *http2Transport) newClientConn(c net.Conn, singleUse bool) (*http2Client
 		pings:                 make(map[[8]byte]chan struct{}),
 		reqHeaderMu:           make(chan struct{}, 1),
 	}
-	if d := t.idleConnTimeout(); d != 0 {
-		cc.idleTimeout = d
-		cc.idleTimer = time.AfterFunc(d, cc.onIdleTimeout)
+
+	if t.t1 != nil && t.t1.MaxConnLifespan > 0 {
+		cc.reuseDeadline = time.Now().Add(t.t1.MaxConnLifespan)
+	}
+
+	cc.idleTimeout = t.idleConnTimeout()
+	ttl, hasTtl := cc.timeToLive()
+
+	if hasTtl || cc.idleTimeout > 0 {
+		timeout := cc.idleTimeout
+		if hasTtl && (timeout <= 0 || ttl < timeout) {
+			timeout = ttl
+		}
+		cc.idleTimer = time.AfterFunc(timeout, cc.onIdleTimeout)
 	}
 	if http2VerboseLogs {
 		t.vlogf("http2: Transport creating client conn %p to %v", cc, c.RemoteAddr())
@@ -7447,6 +7459,30 @@ func (t *http2Transport) newClientConn(c net.Conn, singleUse bool) (*http2Client
 
 	go cc.readLoop()
 	return cc, nil
+}
+
+// timeToLive checks if a http2ClientConn has been initialized
+// from a transport with MaxConnLifespan > 0 and returns the time
+// remaining for this connection to be reusable. The second response
+// would be true in this case.
+//
+// If the connection has a zero-value reuseDeadline set then
+// it returns (0, false)
+//
+// The returned duration will never be less than zero and the connection's
+// idle time is NOT taken into account.
+func (cc *http2ClientConn) timeToLive() (time.Duration, bool) {
+
+	if cc.reuseDeadline.IsZero() {
+		return 0, false
+	}
+
+	ttl := time.Until(cc.reuseDeadline)
+	if ttl < 0 {
+		return 0, true
+	}
+
+	return ttl, true
 }
 
 func (cc *http2ClientConn) healthCheck() {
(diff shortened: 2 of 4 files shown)


josephcopenhaver:
@googlebot I signed it!

leilei911:
Hi @josephcopenhaver, thanks for creating the PR. I was wondering when you think we could merge the PR. We have a DNS resolution issue with our service and could really use your fix.

josephcopenhaver:
@leilei911 I am a new contributor to go's standard sdk in general and unfortunately I do not have the influence to provide you with a timeline.

The latest comment from @neild at https://go-review.googlesource.com/c/go/+/327474 indicates that `Any new package API needs to go through the proposal process.` which shows as resolved and may be in an upcoming candidate patch set if I read this all correctly.

josephcopenhaver:
Cross posting from: https://github.com/golang/go/issues/23427#issuecomment-1116977666

I wonder how much of this problem is purely one of loadbalancing concepts people may be trying to implement.

If a host record resolves to 3 ip nodes and suddenly we add or remove one, then respectively we need to ensure the new node starts taking requests as soon as we know about it so load spreads ( no need to close connections until we approach the max connection limit or idle timeouts happen ) or remove connections tied to the deregistering ( potentially already deregistered ) ip address.

The latter case of removing a node is trivial and fairly self-correcting before any DNS TTL and refresh cycle would inform us on average.

In the same token, idle connection timeout can create "randomly hot" nodes over time by just the probability of FIFO rotating through connections cyclicly. A max lifespan is a fairly nuclear solution to this problem when perhaps envoy proxy should be the full-featured solution for balancing out-of-app.

For people wanting a solution in the same binary they really probably want LIFO connection queuing under a round-robin ( or some distance/efficiency-aware weighted function ) connection broker over the LIFO queues partitioned by IP. This would allow for any "logically extra" connections the service can live without to end up dying off just by virtue of the connection timeouts.

If DNS appears to remove a host we can assume the DNS source has determined that the host should get phased out and schedule the relevant connections to die as soon as their transactions finish or immediately inform the broker to kill that pool of trusted connections.

As a DNS entry is discovered for a hostname a LIFO queue would be created and added to the connection broker. Load would spread almost immediately across the upstream offering, and old connections no longer getting any use on the other nodes would start to time out and die off, expensive new connection flows continue to be avoided.

Connection max lifetime/reuse is likely not the feature people need or want. ( If you do need/want it cool, my PR offers it! ) It just looks like what we're reaching for in this conversation when we really want to observe+schedule DNS resolution and use that to inform a connection pooling strategy where we can change the implementation of popping and pushing to a queue we maintain externally that is LIFO or FIFO or anything else - but note we need to know the connection's target IP to achieve these goals. It's worth discussing in more detail by persons probably smarter than myself. Typically when users of one of my libs want agency I will happily give them the simplest interface they need to achieve their goals and default the behavior to off. Perhaps the maintainers are willing to do similar here?

PR #46808, opened 2021-06-17
cmd/compile/internal/noder: fix a deadlock when compiling large packages of incorrect files

NickyBoy89:
When compiling large packages of files that are syntactically correct, but otherwise incorrect (types, etc...), a deadlock happens because the line 62 `syntax.Parse` is trying to push multiple errors to the unbuffered error channel on the node, stopping the goroutines, and halting the main routine at line 68 waiting to range over an open channel.

Buffering the channel to allow for multiple errors to be pushed to the node fixes this problem.


NickyBoy89:
@googlebot I signed it!

PR #46848, opened 2021-06-21
cmd/compile: add Tilde (OCOM), deprecate XOR unary

Dentrax:
Using ^x in other languages causes an invalid compile and throws a syntax error. To be more align with the usage of the ~ unary op just like in other languages, it's had better to deprecate the usage of ^x.

This commit renames OBITNOT op to OCOM, since we already use COM opcode in the assembly calls, in order to do bitwise not operations.

This is an experimental PR for the learning of the compiler internals. Your feedbacks are valuable! Thanks.

Fixes #46847

Signed-off-by: Dentrax <furkan.turkal@trendyol.com>

tilde-optimized


YoutacVA (approved):

mvdan:
Please note that you should fill https://github.com/golang/proposal/blob/master/go2-language-changes.md when proposing a language change. You should do that in the issue tracker.

I'm going to close this for now, because a pull request implementing a change is not going to get attention when the language change hasn't been proposed yet.

PR #46851, opened 2021-06-21
runtime/internal/atomic: remove incorrect pointer indirection in comment

yigenshutiao:


yigenshutiao:
@googlebot I signed it!

PR #46862, opened 2021-06-21
strings: adding micro-optimization for TrimSpace

Illirgway:
replace for string's end trimming TrimFunc -> TrimRightFunc

strings.TrimSpace string's end trimming should use more specific TrimRightFunc instead of common TrimFunc (because start has already trimmed before)


Illirgway:
@googlebot I signed it!

stokito (approved):

PR #46923, opened 2021-06-25
net/http: guarantee that the Transport dial functions are respected in js/wasm

### LINKED ISSUE #27495: net/http: Expose the normal transport RoundTripper for WASM/js
On WASM, the standard `http.Transfer` RoundTripper is not exposed for external use. Buildflags for wasm replace the standard implementation of RoundTripper with a special version that uses the browser Fetch API. The special JS RoundTripper does not seem to use or honor any of the documented settings provided by `http.Transport`, such as `Dial` or `DialContext`. But the way the special JS RoundTripper is implemented as methods on `http.Transport`, it prevents using the standard RoundTripper.

There are use cases for using the normal RoundTripper, for instance for testing and simulation in the browser. The normal RoundTripper is compiled and included in the binary anyway, and works perfectly well, but is masked by the browser version.

go1.11

GOOS=js
GOARCH=wasm

### What did you do?

I tried to configure a `http.Client` to use a transport with a custom dial function:
https://play.golang.org/p/dnyTivQ643J

This works on all other platforms including the playground, but on WASM, the same, special Roundtripper is always used and there is no way to use the standard Roundtripper.

```
Get https://api.ipify.org/: expected error, all ok
```

### What did you see instead?

On WASM:
```
192.0.2.77
Done.
```

On other platforms, I get the expected result

neilalexander:
The net/http package has a documented contract that if DialContext, DialDLSContext, Dial or DialTLS are specified in an instance of Transport, that they will be used to set up the connection. If they are not specified, then a reasonable fallback is made (e.g. using the net package).

This is ordinarily true, except for when compiling for the js/wasm target, where the browser's Fetch API is preferred in all cases (except for when it is undefined/unavailable) and therefore the dial functions are all ignored. As a result, the http.Transport implementation under js/wasm doesn't meet that contract.

This PR updates the RoundTrip behaviour of http.Transport so that if DialContext, DialTLSContext, Dial or DialTLS are specified, they are used as expected. The Fetch API will be used as a fallback if they are not specified.

Fixes #27495

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1015,6 +1015,7 @@ Nathan Youngman <git@nathany.com>
 Nathaniel Cook <nvcook42@gmail.com>
 Naveen Kumar Sangi <naveenkumarsangi@protonmail.com>
 Neelesh Chandola <neelesh.c98@gmail.com>
+Neil Alexander <neilalexander@neilalexander.dev>
 Neil Lyons <nwjlyons@googlemail.com>
 Netflix, Inc.
 Neuman Vong <neuman.vong@gmail.com>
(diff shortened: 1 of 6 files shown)


neilalexander:
@googlebot I signed it!

PR #46953, opened 2021-06-28
cmd/go/internal/work: remove handler shared work state

emcfarlane:
Removes the handler shared work state by communicating over two channels.
Ready queue is managed by the caller.


emcfarlane:
@googlebot I signed it!

PR #46971, opened 2021-06-29
cpu: add HasSHA512 feature bit in ARM64

howjmay:
Add HasSHA512 feature bit in ARM64 which allows us to check
if the running machine support SHA512 SIMD features.


HowJMay:
This PR will help #46970

PR #47073, opened 2021-07-06
crypto/boring: Refactor boringcrypto codebase

derekparker:
This patch refactors the boringcrypto implementation in an attempt to
make it more general and abstract away the actual external crypto
backend being used.

This is in effort to enable other external crypto libraries such as
OpenSSL to be used instead.

This approach uses a series of small, targeted and unexported interfaces which represent an "external" crypto library. The initial "external" crypto library is, of course, boringcrypto. The change makes it so the "backend" crypto implementation of the `boring` package can be swapped out without requiring any new imports throughout the rest of the code base. Calls into the `boring` package will defer the actual implementation to this "external" crypto library.


derekparker:
This change was broken out of https://github.com/golang/go/pull/43900. I tried to make this CL as small as I could, however going any smaller would result in a CL that would not build, so this is the minimum set of changes that can be merged while still producing a correct build which passes all tests.

Since this is part of a larger effort, and assuming this gets approved eventually, I wonder if it makes sense to create another development branch which will eventually get merged back into `dev.boringcrypto` once all patches have been merged and the implementation of adding OpenSSL support is complete?

(cc @FiloSottile )

derekparker:
This is a pure refactor and does not include any changes to functionality or behavior, other than introducing the interface indirection.

PR #47218, opened 2021-07-15
cmd/link: fix zero offset in funcnametab

yota9:
The zero offset in funcnametab is considered to be invalid at
runtime (runtime.cfuncname checks for zero offset). Fix the funcnametab
generation for zero offsets to be invalid, the same way as pctab does it.


yota9:
Casting @jeremyfaller

PR #47250, opened 2021-07-16
text/scanner: use Go convention in Position doc comment

mehradsadeghi:


mehradsadeghi:
@googlebot I signed it!

PR #47293, opened 2021-07-19
go/token: match the implementation of index selection with sort.Search

shmsr:
name          old time/op    new time/op    delta
SearchInts-8    15.5ns ± 2%    13.7ns ± 4%  -11.87%  (p=0.008 n=5+5)

(see CL 36332 for the original change to sort.Search)


shmsr:
Now that's Go 1.17 is released, this could be merged.

PR #47371, opened 2021-07-24
cmd/go: go test flag -failfast should be cacheable

### LINKED ISSUE #47355: cmd/go: go test flag `-failfast` should be cacheable
Running tests with `-failfast` should use the test cache. It was unexpected that this was not the case. https://github.com/golang/go/issues/42544 uncovers that this is documented. My use case is running `go test all` in projects with a large number of dependencies and reporting failing tests to them.

korzhao:
Add failfast to cacheable list and update docs

Fixes #47355


ek233:
@googlebot I fixed it.

PR #47394, opened 2021-07-26
net: update IP.String doc to reflect RFC 5952 conformance

### LINKED ISSUE #44485: net: comments preceding IP.String should reflect RFC 5952 conformance status

<pre>
go version go1.15.7 darwin/amd64
</pre>

Y

The method in src/net/ip.go `func (ip IP) String()` is supposed to conform to RFC 5952 according to this comment:
https://github.com/golang/go/issues/30264#issuecomment-464540981

The comment is by @mikioh 

But, unfortunately, the source-code comments above src/net/ip.go `func (ip IP) String()` do
not in any way mention that the IPv6 output is supposed to conform to RFC 5952. 
RFC 5952 is "A Recommendation for IPv6 Address Text Representation"

**_If, indeed, -> src/net/ip.go `func (ip IP) String()` should conform to RFC 5952 for IPv6 addresses,
could the comments preceding the method be changed?_**

I would be happy to submit a PR if it is deemed:
1. True that src/net/ip.go `func (ip IP) String()` does (shall) need to conform to RFC 5952
2. The comment explaining this is useful to others.

The reason this is valuable, is some developers need to store, and then compare the ASCII representations of
IPv6 addresses.
A developer presented with the issue of needing to compare ASCII representations is then presented with the 
issue of how to normalize the ASCII representation, so that systems, such as databases can use a uniform
method of representing the IPv6 addresses. This is important so that addresses which are truly equal are also
recognized when comparing the ASCII representation.
"stuff needs to work correctly, as expected when storing & comparing IPv6 addresses"

hitzhangjie:
Fixes #44485

### DIFF
--- a/src/net/ip.go
+++ b/src/net/ip.go
@@ -310,6 +310,7 @@ func ubtoa(dst []byte, start int, v byte) int {
 //   - dotted decimal ("192.0.2.1"), if ip is an IPv4 or IP4-mapped IPv6 address
 //   - IPv6 ("2001:db8::1"), if ip is a valid IPv6 address
 //   - the hexadecimal form of ip, without punctuation, if no other cases apply
+// For IPv6, the returned forms conform to RFC 4291 and RFC 5952.
 func (ip IP) String() string {
 	p := ip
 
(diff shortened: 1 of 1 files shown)


hitzhangjie:
@gopherbot

PR #47417, opened 2021-07-27
list 'e.next' points to nil - Dev.boringcrypto.go1.16

liuyongchen:
When 'at.next' points to nil, it causes e.next.prev to refer to an empty address; which is judged to be e.next.prev, and when the element after e points to nil, e.next.prev is no longer assigned.


liuyongchen:
@googlebot I signed it!

seankhliao:
boringcrypto isn't a maintained branch anymore

Issue #43264, opened 2020-12-18
cmd/link: -buildmode=c-shared not supported on linux/mips64le


Mengkzhaoyun:
### What version of Go are you using (`go version`)?

<pre>
$ go version
1.15.6
</pre>

### Does this issue reproduce with the latest release?

yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="mips64le"
GOBIN=""
GOCACHE="/root/.cache/go-build"
GOENV="/root/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="mips64le"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/go"
GOPRIVATE=""
GOPROXY="https://goproxy.cn,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_mips64le"
GCCGO="gccgo"
GOMIPS64="hardfloat"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/go/src/github.com/cilium/cilium/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -mabi=64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build215838537=/tmp/go-build -gno-record-gcc-switches"
</pre></details>

### What did you do?

When I tried to migrate Cilium to the mips64le (loongnix) platform, I got an error, among others, when I compiled a class library, Proxylib, and found that Cilium was handling it as a .so file.
Here is the compile command

<pre>
[...]

PR #47497, opened 2021-08-02
cmd/link: support -buildmode=c-shared on linux/mips64le

XiaodongLoong:
For golang/go#43264

Signed-off-by: Xiaodong Liu <liuxiaodong@loongson.cn>


XiaodongLoong:
@gopherbot import this PR

PR #47589, opened 2021-08-07
math: add ARM64 implementation of frexp

howjmay:
The implementation of Frexp on  ARM64 has been implemented in this PR.

The following benchmark was run on Apple Silicon M1 chips.
BenchmarkArchFrexp-8    1.581 ns/op  0 B/op  0 allocs/op
BenchmarkFrexpGO-8   	2.079 ns/op  0 B/op  0 allocs/op

The time spent on each iteration drop 23.95%.

### DIFF
--- a/src/math/frexp_arm64.s
+++ b/src/math/frexp_arm64.s
@@ -1,0 +1,57 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#include "textflag.h"
+
+#define PosInf 0x7FF0000000000000
+#define Float64SmallestNormal 2.2250738585072014e-308
+
+// func archFrexp(x float64) (frac float64, exp int)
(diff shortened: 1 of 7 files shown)


src/math/frexp_arm64.s
```diff
@@ -0,0 +1,57 @@
...
+	FCMPD	F29, F3
+	BGE	pass
+
+	// if abs(x) < SmallestNormal ,then run the following special case
+	FMOVD	$(4503599627370496.0), F2 // 4503599627370496.0 is (1<<52)
+	FMULD	F2, F0, F0
+	SUB	$(52), R1, R1 // set R1 to -52
+	FMOVD	F0, R0
+	// tmp = int((x>>shift)&mask)
+    UBFX $52, R0, $11, R0
```
lovung, line 38: Is it a wrong indent?

HowJMay: Thank you for pointing out. This style twist was causing by space. I have replaced it by tab. Thank you.

PR #47597, opened 2021-08-08
time: update current time comment

astraw99:
In the time package, the ticker and timer both send
current time to channel C, so this PR update the comment
to understand them better.


astraw99:
@googlebot I signed it!

PR #47640, opened 2021-08-11
bufio: add Reader.DiscardRune(n int)

cafra:
the Gerrit change.

Fixes golang/go#47621


mvdan:
The proposal was retracted; closing.

PR #47685, opened 2021-08-13
encoding/json: fix unmarshal null array will not change the original value

a631807682:
The problem is `Unmarshal` `null`  will not reset array to zero value.
```
v := [2]int64{1, 2}
json.Unmarshal([]byte(`null`), &v) // [1, 2]
```


a631807682:
@googlebot I signed it!

a631807682:
primitives

PR #47772, opened 2021-08-18
net: listen on specified multicast address

### LINKED ISSUE #34728: net: ListenPacket can't be used on multicast address

go version go1.12.9

Yes, the code is unchanged on master.

$ go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/quentin/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/quentin/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/kn/0gfzt13x37j63313z_5szy3h0000gn/T/go-build593496325=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre></details>

### What did you do?

```
net.ListenPacket("udp4", "224.0.0.128:5076")
```

A socket bound to 224.0.0.128:5076 (but not joined to a multicast group)

### What did you see instead?

A socket bound to 0.0.0.0:5076 (and not joined to any multicast group)

`net.ListenPacket`'s documentation says:

> For UDP and IP networks, if the host in the address parameter is empty or a literal unspecified IP address, ListenPacket listens on all available IP addresses of the local system except multicast IP addresses. 

But in fact it also applies this logic if the address is a multicast address. I suspect this is because `net.ListenMulticastUDP` is documented as having this behavior:

> ListenMulticastUDP listens on all available IP addresses of the local system including the group, multicast IP address.

and it shares an implementation function with `ListenPacket`:

https://github.com/golang/go/blob/master/src/net/sock_posix.go#L210

And no, golang.org/x/net/ipv4 does not provide a workaround here, because all of its APIs require that you start with `net.Listen*` to get a `net.PacketConn`. Likewise, the `syscall` package doesn't provide an escape hatch because `net.FilePacketConn` is not implemented on Windows (though I will likely end up doing that anyway, with my resulting code only supporting POSIX environments).

danielkucera:
Fixes #34728

the Gerrit change.

  + No trailing period
+ No Markdown

### DIFF
--- a/src/net/sock_posix.go
+++ b/src/net/sock_posix.go
@@ -208,25 +208,10 @@ func (fd *netFD) listenStream(laddr sockaddr, backlog int, ctrlFn func(string, s
 func (fd *netFD) listenDatagram(laddr sockaddr, ctrlFn func(string, string, syscall.RawConn) error) error {
 	switch addr := laddr.(type) {
 	case *UDPAddr:
-		// We provide a socket that listens to a wildcard
-		// address with reusable UDP port when the given laddr
-		// is an appropriate UDP multicast address prefix.
-		// This makes it possible for a single UDP listener to
-		// join multiple different group addresses, for
-		// multiple UDP listeners that listen on the same UDP
-		// port to join the same group address.
 		if addr.IP != nil && addr.IP.IsMulticast() {
 			if err := setDefaultMulticastSockopts(fd.pfd.Sysfd); err != nil {
 				return err
 			}
-			addr := *addr
-			switch fd.family {
-			case syscall.AF_INET:
-				addr.IP = IPv4zero
-			case syscall.AF_INET6:
-				addr.IP = IPv6unspecified
-			}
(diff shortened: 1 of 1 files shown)


danielkucera:
@googlebot I signed it!

PR #47835, opened 2021-08-20
os: use readlock to optimize getwd

chyroc:


chyroc:
/comments Thanks for your review, sorry to disturb you, I will close this pr.

PR #47902, opened 2021-08-23
encoding/gob: use reflect.Value.MapRange for map iteration

KimMachineGun:
This change will reduce memory usage and improve performance by eliminating unnecessary map iteration for making key slice.


KimMachineGun:
Duplicate of #47476

PR #47940, opened 2021-08-24
math/big: Support precision of -1 in Rat.FloatString

### LINKED ISSUE #47911: math/big: Support precision of -1 in Rat.FloatString
The `strconv.FormatFloat` function accepts a special precision of `-1` to mean to use the smallest number of digits necessary such that `ParseFloat` will return float exactly. This is a feature request to accept the same for `math/big.Rat.FloatString`.

180909:
Fixes #47911

### DIFF
--- a/src/math/big/ratconv.go
+++ b/src/math/big/ratconv.go
@@ -323,6 +323,10 @@ func (x *Rat) RatString() string {
 // digits of precision after the radix point. The last digit is rounded to
 // nearest, with halves rounded away from zero.
 func (x *Rat) FloatString(prec int) string {
+	if prec == -1 {
+		return x.FloatString(16)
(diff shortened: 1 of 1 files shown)


src/math/big/ratconv.go
```diff
@@ -323,6 +323,10 @@ func (x *Rat) RatString() string {
 // digits of precision after the radix point. The last digit is rounded to
 // nearest, with halves rounded away from zero.
 func (x *Rat) FloatString(prec int) string {
+	if prec == -1 {
+		return x.FloatString(16)
```
Thomasdezeeuw, line 327: I don't how or why `16` would fix this, but couldn't you set `prec =16` if it does?

PR #48015, opened 2021-08-27
os/user: add users and groups iterators

Mantelijo:
Fixes #47907 

Go standard library has os/user package which allows to lookup user
or group records via either user/group name or id. This pull request
extends capabilities of os/user package by introducing iteration
functionality for users and groups.
Users and groups iteration functionality might be useful in cases where
a full or partial list of all available users/groups is required in an
application.

On unix, iteration functionality relies
on two internal implementations: one that does not use cgo and parses
contents of /etc/passwd and /etc/group files for users and groups
respectively, and another one that does use cgo and utilizes POSIX
compliant libc library routines getpwent and getgrent for users and
groups respectively. On windows,
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList
registry key is iterated for SID values, each SID value represents either
a user or a group.

Tested on Debian GNU/Linux 10 (buster), macOS 11.5.1, freebsd 13.0,
plan9 fourth edition, windows 10 version 21H1.

Comparison of go iterators vs getent (with time -v) with ~17000 entries of users and ~4000 entries of groups:

getent passwd:
- User time (seconds): 0.01
- System time (seconds): 0.00
- Percent of CPU this job got: 100%
- Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.01
- Maximum resident set size (kbytes): 2200
- Average resident set size (kbytes): 0

go users/groups iterators passwd:
- User time (seconds): 0.02
- System time (seconds): 0.00
- Percent of CPU this job got: 112%
- Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.02
- Maximum resident set size (kbytes): 6664
- Average resident set size (kbytes): 0

getent group:
- User time (seconds): 0.02
- System time (seconds): 0.00
- Percent of CPU this job got: 100%
- Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.02
- Maximum resident set size (kbytes): 3364
- Average resident set size (kbytes): 0

go users/groups iterators group:
- User time (seconds): 0.00
- System time (seconds): 0.00
- Percent of CPU this job got: 100%
- Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.00
- Maximum resident set size (kbytes): 4144
- Average resident set size (kbytes): 0

### DIFF
--- a/src/internal/syscall/windows/registry/key.go
+++ b/src/internal/syscall/windows/registry/key.go
@@ -87,9 +87,9 @@ func OpenKey(k Key, path string, access uint32) (Key, error) {
 	return Key(subkey), nil
 }
 
-// ReadSubKeyNames returns the names of subkeys of key k.
-func (k Key) ReadSubKeyNames() ([]string, error) {
-	names := make([]string, 0)
+// ReadSubKeyNames iterates over the names of subkeys of key k. Callback function fn receives each iterated subkey name.
+// If fn returns non-nil error, iteration is terminated and that error is returned.
+func (k Key) ReadSubKeyNames(fn func(string) error) error {
 	// Registry key size limit is 255 bytes and described there:
 	// https://msdn.microsoft.com/library/windows/desktop/ms724872.aspx
 	buf := make([]uint16, 256) //plus extra room for terminating zero byte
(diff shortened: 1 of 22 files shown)


mvdan:
The proposal was declined; closing.

Issue #48089, opened 2021-08-31
the builtin function `append` annotations looks like should be updated


Jeyrce:
### What version of Go are you using (`go version`)?

<pre>
1.16.7

</pre>

### Does this issue reproduce with the latest release?

Yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
amd64

</pre></details>

### What did you do?

I saw the annotation for builtin function `append`, found it seems make mistake

```
// The append built-in function appends elements to the end of a slice. If
// it has sufficient capacity, the destination is resliced to accommodate the
// new elements. If it does not, a new underlying array will be allocated.
// Append returns the updated slice. It is therefore necessary to store the
// result of append, often in the variable holding the slice itself:
//	slice = append(slice, elem1, elem2)
//	slice = append(slice, anotherSlice...)
// As a special case, it is legal to append a string to a byte slice, like this:
//	slice = append([]byte("hello "), "world"...)
func append(slice []Type, elems ...Type) []Type
```

### What did you expect to see?

I tried  append([]byte("s"), "x"), but can not run it

### What did you see instead?

Is the annotation should be updated?

robpike:
You said you can run it. Can you? I get

```
./prog.go:4:12: cannot use "x" (type untyped string) as type byte in append
```

which is the error I expected. Perhaps you think "x" is a byte, but it is not: it is a string containing one byte. Try 

```
append([]byte("s"), 'x')
```

instead.

Jeyrce:
Yeah, `rune` type is the same as `byte` so it works, but the annotation in builtin.go is:

```
// As a special case, it is legal to append a string to a byte slice, like this:
//	slice = append([]byte("hello "), "world"...)
func append(slice []Type, elems ...Type) []Type
```

It doesn't work, so do we need to update the annotation?

rsheasby:
I think @Jeyrce misunderstood the annotation a bit. The relevant part of the annotation is this:
```
// As a special case, it is legal to append a string to a byte slice, like this:
//	slice = append([]byte("hello "), "world"...)
```

Note that what @Jeyrce tried was this:
```
append([]byte("s"), "x")
```

This is incorrect, because you're missing the `...` after `"x"`.

The way that special case works is like so:
[...]

PR #48090, opened 2021-08-31
builtin: remove confusing mention of special case in append annotation

rsheasby:
The current documentation for append shows string exploding, but describes it as a special case. As far as I can tell, this isn't actually a special case. This PR will update that to properly explain why the example works.

Inspired by this issue: #48089


rsheasby:
@googlebot I signed it!

PR #48104, opened 2021-08-31
net/http: add MaxBytesHandler

earthboundkid:
Fixes #39567

### DIFF
--- a/src/net/http/server.go
+++ b/src/net/http/server.go
@@ -3567,3 +3567,12 @@ func tlsRecordHeaderLooksLikeHTTP(hdr [5]byte) bool {
 	}
 	return false
 }
+
+// MaxBytesHandler returns a Handler that runs h with its ResponseWriter and Request.Body wrapped by a MaxBytesReader.
+func MaxBytesHandler(h Handler, n int64) Handler {
+	return HandlerFunc(func(w ResponseWriter, r *Request) {
+		r2 := *r
+		r2.Body = MaxBytesReader(w, r.Body, n)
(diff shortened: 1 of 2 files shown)


src/net/http/server.go
```diff
@@ -3567,3 +3567,12 @@ func tlsRecordHeaderLooksLikeHTTP(hdr [5]byte) bool {
 	}
 	return false
 }
+
+// MaxBytesHandler returns a Handler that runs h with its ResponseWriter and Request.Body wrapped by a MaxBytesReader.
+func MaxBytesHandler(h Handler, n int64) Handler {
+	return HandlerFunc(func(w ResponseWriter, r *Request) {
+		r2 := *r
```
inliquid, line 3574: Sorry, may I ask you what's the purpose of copying request here?
I already use my own implementation in middleware like this, for a while already, and didn't get into any issues so far:
```go
func maxBodyMiddleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		r.Body = http.MaxBytesReader(w, r.Body, int64(cfg.MaxBodySize))

		next.ServeHTTP(w, r)
	})
}
```
so would like to know if there are some edge cases which I missed.

carlmjohnson: From the docs for `http.Handler`,

> Except for reading the body, handlers should not modify the provided Request.

You will often be fine modifying a request, but you aren’t supposed to do it because it might be a race condition if some other middleware happens to read the request at the same time.

PR #48138, opened 2021-09-02
syscall: change default flag for utimensat to AT_SYMLINK_NOFOLLOW on linux

zq-david-wang:
The default behavior for utimensat is following links to the real file, this makes it impossible to change the timestamp of a symbol link itself in golang. This code changes the behavior of syscall.UtimesNano on symbol link on linux, and do not follow links anymore. This way, the timestamp for the real file could be changed after following the link, and the link file cloud be changed directly.


zq-david-wang:
@googlebot I signed it!

PR #48169, opened 2021-09-03
 misc/wasm: make wasm_exec.js try catch when require module

serkodev:
Some of the environment (e.g. Cloudflare Worker) maybe not support for "fs" (or other modules) and returns 'Uncaught Error: No such module' error.
This changes can fix the issue.

### DIFF
--- a/misc/wasm/wasm_exec.js
+++ b/misc/wasm/wasm_exec.js
@@ -105,12 +105,14 @@
 	}
 
 	if (!global.crypto && global.require) {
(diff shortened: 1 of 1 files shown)


serkodev:
@googlebot I signed it!

serkodev:
@neelance Thanks for making this [solution](https://go-review.googlesource.com/c/go/+/347353) to provide polyfills for different environment. 🎉

PR #48267, opened 2021-09-08
text/template: add the 'parent' keyword

paulsmith:
Add the keyword `parent` to the text/template language. The keyword,
when used in place of a template name string in a template execution
node, causes execution of that template to refer to a previously-defined
template of the same name. In the context of template inheritance, this
allows child templates to access the contents of the parent template they
are overriding.

For example:

base template: `{{define "title"}Example.com{{end}}`

inheriting: `{{define "title"}}Welcome - {{template parent .}}{{end}}`

Would produce: `Welcome - Example.com`

Fixes #48266

### DIFF
--- a/src/text/template/exampleparent_test.go
+++ b/src/text/template/exampleparent_test.go
@@ -1,0 +1,78 @@
+package template_test
+
+import (
+	"log"
+	"os"
+	"text/template"
+)
+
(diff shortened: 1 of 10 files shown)


mvdan:
The proposal was declined; closing.

PR #48273, opened 2021-09-09
crypto/tls: populate peerCertificates before validation fails

sooslaca:
With this change, peer certificate details can be examined in case of provided client certificate is invalid (eg expired).


sooslaca:
@googlebot I signed it!

PR #48388, opened 2021-09-14
cmd/go/internal/test: clarify `-count` default

### LINKED ISSUE #39056: feature request: add flag `go test -no-cache`
#24573 is closed, but it has all the information. `-count=1` to most people does not mean `-no-cache`. It's just not intuitive.

This is a request to implement `-no-cache` flag on go test so that this kind of behavior isn't hidden from users.

Alternatively (backwards incompatible), since caching of test results is a bit surprising, instead allow uses to decide to cache or not explicitly with `-cache` flag, and set the default to not cache.

ScottSuarez:
Count default is only 1 when the test is not cached. 
If the test is cached it will not rerun the test. 

This documentation causes confusion because the only way to run tests without cache is to use `-count=1`. According to the documentation as is `-count=1` should have no effect. Changing this CLI documentation to clarify the behavior.

ref: https://github.com/golang/go/issues/39056?ts=4#issuecomment-631688012
resolves: https://github.com/golang/go/issues/39056

### DIFF
--- a/src/cmd/go/internal/test/test.go
+++ b/src/cmd/go/internal/test/test.go
@@ -208,7 +208,7 @@ control the execution of any test:
 	    (for example, -benchtime 100x).
 
 	-count n
-	    Run each test and benchmark n times (default 1).
+	    Run each test and benchmark n times (default 1 unless cached).
 	    If -cpu is set, run n times for each GOMAXPROCS value.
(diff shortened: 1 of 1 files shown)


ScottSuarez:
closing this after comment inline clarified with explanation for current description.

PR #48426, opened 2021-09-16
Unify "unreachable" panic messages

YenForYang:
Existing code occasionally uses "not reached" or "not reachable"
when panicking. Use "unreachable" as the panic message instead,
since it is already most commonly used.


YenForYang:
@googlebot I signed it!

PR #48450, opened 2021-09-17
encoding/xml: allow mismatched `XMLName` / field metadata tag names

### LINKED ISSUE #219: encoding/xml error when field metadata doesn't match XMLName
I'm not sure what to do about this situation, so just opening up an issue in case someone has a good idea.

I generated client code from [this WSDL](https://community.workday.com/sites/default/files/file-hosting/productionapi/Payroll/v37.0/Payroll.wsdl), using a patched gowsdl 9e1cc9a7689230f4da66504844af631a0c4767cc with #214 and #218 applied.  The generated code returns an error when unmarshaling:

    xml: name "Off-cycle_Type_Reference" in tag of payroll.Payroll_ResultType.Offcycle_Type_Reference conflicts with name "Payroll_Off-cycle_TypeObjectType" in *payroll.Payroll_Offcycle_TypeObjectType.XMLName

The crux of the issue seems to be this poorly named `complexType`:

```xml
      <xsd:complexType name="Payroll_Off-cycle_TypeObjectType">
        <xsd:annotation wd:Is_Reference_ID="1"/>
        <xsd:sequence>
          <xsd:element name="ID" type="wd:Payroll_Off-cycle_TypeObjectIDType" minOccurs="0" maxOccurs="unbounded"/>
        </xsd:sequence>
        <xsd:attribute name="Descriptor" type="xsd:string">
          <xsd:annotation>
            <xsd:documentation>Display information used to describe an instance of an object. This 'optional' information is for outbound descriptive purposes only and is not processed on inbound Workday Web Services requests.</xsd:documentation>
          </xsd:annotation>
        </xsd:attribute>
      </xsd:complexType>
```

Go symbols can't have dashes in their names, so the WSDL name gets mangled into `Payroll_Offcycle_TypeObjectType`: https://github.com/hooklift/gowsdl/blob/9e1cc9a7689230f4da66504844af631a0c4767cc/types_tmpl.go#L145

Because the struct name is now different than the type name in the WSDL, this conditional evaluates to `true`, and adds an `XMLName` field to the struct: https://github.com/hooklift/gowsdl/blob/9e1cc9a7689230f4da66504844af631a0c4767cc/types_tmpl.go#L151-L153

But, the places where this is actually used specify a different tag name.  ex the `Payroll_ResultType` specifies that it goes into an `Off-cycle_Type_Reference` tag:

```xml
          <xsd:element name="Off-cycle_Type_Reference" type="wd:Payroll_Off-cycle_TypeObjectType" minOccurs="0">
            <xsd:annotation>
              <xsd:documentation>Off-cycle Type Reference</xsd:documentation>
            </xsd:annotation>
          </xsd:element>
```

gowsdl generates correct output, with the field having the correct type and the metadata specifying the name from the `xsd:element`:

```go
Offcycle_Type_Reference *Payroll_Offcycle_TypeObjectType `xml:"Off-cycle_Type_Reference,omitempty" json:"Off-cycle_Type_Reference,omitempty"`
```

…but `encoding/xml` then blows up because the tag name in `Payroll_ResultType.Offcycle_Type_Reference`'s field metadata is different than the one in `Payroll_Offcycle_TypeObjectType.XMLName`.

I think the most correct fix here is for `encoding/xml` to prefer the tag name from the containing struct, if set.  In the absence of that, I'm not sure what to do other than postprocess the generated code to remove the `XMLName`.

Suggestions?

ieure:
Currently, if you have nested structs, where the inner struct’s
XMLName doesn’t match the outer struct’s field name metadata, such as:

	type InnerStruct struct {
		XMLName Name   `xml:"innerStruct"`
		Name	string `xml:"name"`
	}

	type Outer struct {
		XMLName Name		 `xml:"outer"`
		Inner	*InnerStruct `xml:"inner"`
	}

Then decoding XML produces an error:

    xml: name "inner" in tag of xml.Outer.Inner conflicts with name "innerStruct" in *xml.InnerStruct.XMLName

This causes issues when interacting with SOAP web services, whose
WSDLs are expressed with XSD.  XSD allows defining reusable types;
each of those types has a unique name.  That name MAY be the name of
the tag containing the struct’s data, but isn’t ALWAYS.  An
xsd:element may specify a different tag name to contain an XSD type’s
values.  An illustration of a real-world case where this happens is
hooklift/gowsdl#219.

This PR uses the big hammer of removing the tests and errors which
cause the issue.  I’m reasonably certain that the structFieldInfo()
changes are fine, but I feel less good about Decoder.unmarshal().
Ideally, that code would look at the field metadata of the containing
struct, then the XMLName of the destination struct.  I’m not familiar
with this code, but it looks like the parent element isn’t accessible
at the time when this check would need to happen, so I opted to remove
it instead.

fixes hooklift/gowsdl#219

### DIFF
--- a/src/encoding/xml/marshal_test.go
+++ b/src/encoding/xml/marshal_test.go
@@ -2490,3 +2490,34 @@ func TestInvalidXMLName(t *testing.T) {
 		t.Errorf("error %q does not contain %q", err, want)
 	}
 }
+
+func TestMismatchedTagNames(t *testing.T) {
+	type InnerStruct struct {
+		XMLName Name   `xml:"innerStruct"`
+		Name    string `xml:"name"`
+	}
+
+	type Outer struct {
+		XMLName Name         `xml:"outer"`
+		Inner   *InnerStruct `xml:"inner"`
+	}
+
+	dec := NewDecoder(strings.NewReader(`<?xml version="1.0"?>
+<outer>
+    <inner>
+        <name>foo</name>
+    </inner>
+</outer>"`))
+
+	holder := &Outer{}
+	if err := dec.Decode(holder); err != nil {
+		t.Errorf("Unexpected error: %q", err)
+	}
+	if holder == nil {
+		t.Error("Holder should be non-nil")
+	} else if holder.Inner == nil {
+		t.Error("Inner should be non-nil")
+	} else if holder.Inner.Name != "foo" {
+		t.Errorf("Expected name to be %q, but was %q", "foo", holder.Inner.Name)
+	}
+}
--- a/src/encoding/xml/read.go
+++ b/src/encoding/xml/read.go
@@ -423,9 +423,6 @@ func (d *Decoder) unmarshal(val reflect.Value, start *StartElement) error {
 		// Validate and assign element name.
 		if tinfo.xmlname != nil {
 			finfo := tinfo.xmlname
-			if finfo.name != "" && finfo.name != start.Name.Local {
-				return UnmarshalError("expected element type <" + finfo.name + "> but have <" + start.Name.Local + ">")
-			}
 			if finfo.xmlns != "" && finfo.xmlns != start.Name.Space {
 				e := "expected element <" + finfo.name + "> in name space " + finfo.xmlns + " but have "
 				if start.Name.Space == "" {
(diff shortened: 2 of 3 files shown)


ieure:
@googlebot I signed it!

ieure:
welp

PR #48452, opened 2021-09-18
go/token: improve the performance of IsKeyword

chanxuehong:
This implementation is not perfect, but there is still a little performance improvement:

The current implementation is not the most efficient. 
Obviously, there is a high probability that we need to compare twice, and we can continue to optimize the hash function.

name                          old time/op  new time/op  delta
IsKeyword-12                   261ns ± 1%   112ns ± 0%  -57.05%  (p=0.000 n=9+9)
Lookup-12                      259ns ± 2%   113ns ± 3%  -56.28%  (p=0.000 n=7+9)
IsKeyword_WithNonKeywords-12   359ns ± 3%    84ns ± 6%  -76.50%  (p=0.000 n=8+10)
Lookup_WithNonKeywords-12      353ns ± 1%    81ns ± 1%  -77.19%  (p=0.000 n=8+10)

### DIFF
--- a/src/go/token/token.go
+++ b/src/go/token/token.go
@@ -283,19 +283,10 @@ func (op Token) Precedence() int {
 	return LowestPrec
 }
 
-var keywords map[string]Token
-
-func init() {
-	keywords = make(map[string]Token)
-	for i := keyword_beg + 1; i < keyword_end; i++ {
-		keywords[tokens[i]] = i
-	}
-}
-
 // Lookup maps an identifier to its keyword token or IDENT (if not a keyword).
 //
 func Lookup(ident string) Token {
-	if tok, is_keyword := keywords[ident]; is_keyword {
+	if tok, ok := isKeyword(ident); ok {
 		return tok
 	}
 	return IDENT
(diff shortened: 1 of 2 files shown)


chanxuehong:
> We found a Contributor License Agreement for you (the sender of this pull request), but were unable to find agreements for all the commit author(s) or Co-authors. If you authored these, maybe you used a different email address in the git commits than was used to sign the CLA ([login here](https://cla.developers.google.com/) to double check)? If these were authored by someone else, then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.
> In order to pass this check, please resolve this problem and then comment `@googlebot I fixed it.`. If the bot doesn't comment, it means it doesn't think anything has changed.
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2Fgolang%2Fgo%2Fpull%2F48452) for more info**.

@googlebot I fixed it.

chanxuehong:
@googlebot I fixed it.

Issue #8068, opened 2014-05-21
encoding/xml: empty namespace prefix definitions should be illegal


ghost:
<pre>What does 'go version' print?
go version go1.2 linux/amd64

What steps reproduce the problem?
<a href="http://play.golang.org/p/f8UWfdF7Hb">http://play.golang.org/p/f8UWfdF7Hb</a>

What happened?
No-error.

What should have happened instead?
Should generate an error.

Per <a href="http://www.w3.org/TR/REC-xml-names/#dt-prefix">http://www.w3.org/TR/REC-xml-names/#dt-prefix</a> a prefix must refer to a namespace
name, not the empty string. 

Please provide any additional information below.

Ran into this in implementing webdav, the litmus tests test this error condition [and
due to xml.Token hiding prefixing it is impossible to detect otherwise]</pre>

nigeltao:
It could be that, at the time of writing of the WebDAV litmus test suite, an empty namespace prefix was illegal, but that has since become legal. The grammar differs between
http://www.w3.org/TR/2006/REC-xml-names-20060816/#ns-decl and
http://www.w3.org/TR/REC-xml-names/#dt-prefix

iWdGo:
Namespace handling is not checking this case before adding the namespace to the namespace array of the decoder. A related fix is submitted which also rejects syntax xmlns:="..." which is invalid.

heschi:
As a heads up, this fix broke a couple of Google-internal tests that process XML from various sources. I believe that the added check is probably correct, but it seems very possible that this will cause some trouble. Should there be a way to opt out? Given that the fix waited 8 years, perhaps rolling back and taking some time to think about it would be reasonable?

ianlancetaylor:
At this stage in the release cycle we should roll back and try again next cycle.  I will send a revert CL.

ianlancetaylor:
Revert CL is 453996, reopening this issue.

ianlancetaylor:
Moving to backlog.

PR #48641, opened 2021-09-27
encoding/xml: support xmlns prefixes

### LINKED ISSUE #11431: encoding/xml: loss of xmlns= in encoding since Go 1.4
It's not obvious at first glance how, but https://golang.org/cl/2660 changed the behavior of the XML marshaler for people who were generating xmlns attributes "by hand". For example:

```
package main

import (
    "encoding/xml"
    "fmt"
    "log"
)

type T struct {
    Ns   string `xml:"xmlns,attr"`
    Body string
}

func main() {
    t := &T{Ns: "http://example.com/ns", Body: "hello world"}
    x, err := xml.Marshal(t)
    if err != nil {
        log.Fatal(err)
    }
    fmt.Printf("%s\n", x)
}
```

In Go 1.4 this program printed:

```
<T xmlns="http://example.com/ns"><Body>hello world</Body></T>
```

After this CL it prints:

```
<T><Body>hello world</Body></T>
```

In the absence of compelling justification for a breaking change, I think we should try to continue to support this way of setting the default xmlns.

I've seen tests break due to this change, and presumably real programs would break too, if the xmlns= were important.

What can we do to interpret these old programs correctly in the new more-namespace-aware world for Go 1.5?

@rogpeppe @nigeltao

ydnar:
Allow round-trip marshaling and unmarshaling with explicit namespace
prefixes. For example, this can be unmarshalled and re-marshalled
into this precise XML:

```xml
<epp xmlns="urn:ietf:params:xml:ns:epp-1.0">
        <command>
                <check>
                        <domain:check xmlns:domain="urn:ietf:params:xml:ns:domain-1.0">
                                <domain:name>golang.org</domain:name>
                                <domain:name>go.dev</domain:name>
                        </domain:check>
                </check>
        </command>
</epp>
```

For marshaling, a preferred namespace prefix can now be specified in a
struct tag or `XMLName` value by prefixing the local name:

`xml:"urn:ietf:params:xml:ns:domain-1.0 domain:check"`

Namespaced tag and attribute names are now strictly parsed and will
fail with an error if any are malformed, such as having a leading or
trailing colon, or more than 1 colon.

This change directly affects the tests introduced in 4d014e723
(https://golang.org/cl/277892). This may have security implications.
See #8068 and #43168.

An example playground that would be fixed with this change:

https://play.golang.org/p/-6Ee8tcLl2L

This is a rebase of https://golang.org/cl/11635 onto the current master branch.
It was originally written by Constantin Konstantinidis, Roger Peppe, and
Martin Hilton.

Original CL message:

When an xmlns="..." attribute was explicitly generated, it was being
ignored because the name space on the attribute was assumed to have been
explicitly set (to the empty name space) and it's not possible to have
an element in the empty name space when there is a non-empty namespace
set.

We fix this by recording when a default namespace has been explicitly
set and setting the name space of the element to that so
printer.defineNS can do its work correctly.

We do not attempt to add our own xmlns="..." attribute when one is
explicitly set.

We also add tests for EncodeElement, as that's the only way to attain
coverage of some of the changed behaviour. Some other test coverage is
also increased, although more work remains to be done in this area.

This change was jointly developed with Martin Hilton (@mhilton on
github).

Updates #11431
Updates #8068

### DIFF
--- a/src/encoding/xml/xml.go
+++ b/src/encoding/xml/xml.go
@@ -207,7 +207,7 @@ type Decoder struct {
 	toClose        Name
 	nextToken      Token
 	nextByte       int
-	ns             map[string]string
+	ns             map[string]string // url for a prefix ns[.Space]=.Value
 	err            error
 	line           int
 	linestart      int64
@@ -304,13 +304,22 @@ func (d *Decoder) Token() (Token, error) {
 		// to the other attribute names, so process
 		// the translations first.
 		for _, a := range t1.Attr {
-			if a.Name.Space == xmlnsPrefix {
-				v, ok := d.ns[a.Name.Local]
-				d.pushNs(a.Name.Local, v, ok)
+			if a.Name.Space == xmlnsPrefix { // name space attribute {.Space}xmlns:{.Local}={.Value}
+				if a.Value == "" {
+					d.err = d.syntaxError("empty namespace without prefix")
+					return nil, d.err
+				}
+				if a.Name.Local == "" {
+					d.err = d.syntaxError("empty prefix")
+					return nil, d.err
+				}
+				v, ok := d.ns[a.Name.Local] // Checking existence
+				// Recording the level of the name space by recording tag name
+				d.pushNs(a.Name.Local, v, ok) // Pushing tag, eventual value, and existence of namespace
 				d.ns[a.Name.Local] = a.Value
 			}
-			if a.Name.Space == "" && a.Name.Local == xmlnsPrefix {
-				// Default space for untagged names
+			if a.Name.Space == "" && a.Name.Local == xmlnsPrefix { // xmlns=".Value"
+				// Default space for non-prefixed names
 				v, ok := d.ns[""]
 				d.pushNs("", v, ok)
 				d.ns[""] = a.Value
@@ -335,8 +344,9 @@ func (d *Decoder) Token() (Token, error) {
 
 const (
 	xmlURL      = "http://www.w3.org/XML/1998/namespace"
-	xmlnsPrefix = "xmlns"
 	xmlPrefix   = "xml"
+	xmlnsURL    = "http://www.w3.org/2000/xmlns/"
+	xmlnsPrefix = "xmlns"
 )
 
 // Apply name space translation to name n.
@@ -345,6 +355,7 @@ const (
 func (d *Decoder) translate(n *Name, isElementName bool) {
 	switch {
 	case n.Space == xmlnsPrefix:
+		n.Space = xmlnsURL
 		return
 	case n.Space == "" && !isElementName:
 		return
@@ -797,6 +808,9 @@ func (d *Decoder) rawToken() (Token, error) {
 	for {
 		d.space()
 		if b, ok = d.mustgetc(); !ok {
+			if len(attr) > 0 && !d.Strict {
+				break // When not strict, an attribute might end with EOF
+			}
 			return nil, d.err
 		}
 		if b == '/' {
@@ -831,7 +845,6 @@ func (d *Decoder) rawToken() (Token, error) {
 				d.err = d.syntaxError("attribute name without = in element")
 				return nil, d.err
 			}
-			d.ungetc(b)
 			a.Value = a.Name.Local
 		} else {
 			d.space()
@@ -1024,6 +1037,11 @@ Input:
 			d.ungetc('<')
 			break Input
 		}
+		// This occurs only for an unquoted attr name.
+		if b == '>' && !cdata && quote < 0 { // Possible end of tag reached
+			d.ungetc('>') // Leaving end of tag available
+			break         // returning text
+		}
 		if quote >= 0 && b == byte(quote) {
 			break Input
 		}
@@ -1120,6 +1138,7 @@ Input:
 		}
 
 		// We must rewrite unescaped \r and \r\n into \n.
+		// End of line handling https://www.w3.org/TR/xml/#sec-line-ends
 		if b == '\r' {
 			d.buf.WriteByte('\n')
 		} else if b1 == '\r' && b == '\n' {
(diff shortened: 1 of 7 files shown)


ydnar:
@googlebot I consent.

ydnar:
:wave: @rogpeppe I hope this was OK!

ydnar:
Just pushed an additional commit that allows callers to specify a preferred XML name space prefix. For example, this XML can now be unmarshalled and re-marshalled byte for byte:

```xml
<epp xmlns="urn:ietf:params:xml:ns:epp-1.0">
        <command>
                <check>
                        <domain:check xmlns:domain="urn:ietf:params:xml:ns:domain-1.0">
                                <domain:name>golang.org</domain:name>
                                <domain:name>go.dev</domain:name>
                        </domain:check>
                </check>
        </command>
</epp>
```

rogpeppe:
@ydnar Thanks for lifting this from the archives.

Unfortunately, as hinted at [here](https://go-review.googlesource.com/c/go/+/12570/) I think that it's not enough to just resurrect the old fixes. We'd need a proposal for the changes and some discussion around them, because making these fixes will inevitably break some clients even as they fix others.

ydnar:
> @ydnar Thanks for lifting this from the archives.
> 
> 
> 
> Unfortunately, as hinted at [here](https://go-review.googlesource.com/c/go/+/12570/) I think that it's not enough to just resurrect the old fixes. We'd need a proposal for the changes and some discussion around them, because making these fixes will inevitably break some clients even as they fix others.

You're probably right. I guess my main question would be "is it worth it?" Given the alternative is to fork and vendor the encoding/xml package, which is relatively small and self-contained.

The API, for better or worse, is complex, but curiously missing some things that seem obvious in hindsight like an XMLName() interface that types could implement in lieu of struct tags or phantom struct fields.

I'm curious other than us (domainr.com), who's using this package at scale in production?

Would contributed broader test coverage help?

ydnar:
> @ydnar Thanks for lifting this from the archives.
> 
> Unfortunately, as hinted at [here](https://go-review.googlesource.com/c/go/+/12570/) I think that it's not enough to just resurrect the old fixes. We'd need a proposal for the changes and some discussion around them, because making these fixes will inevitably break some clients even as they fix others.

@rogpeppe thanks for the nudge. I made a proposal here: #48821

iwdgo:
@ydnar CLA was already approved when I mailed the first suggested fixes. To my knowledge, nothing moved since then. I obviously consented to contribute. It sounds odd that a new consent is requested.

ydnar:
> @ydnar CLA was already approved when I mailed the first suggested fixes. To my knowledge, nothing moved since then. I obviously consented to contribute. It sounds odd that a new consent is requested.

I think it's because I rebased your initial commit, so the CLA bot can't detect your consent:

> All (the pull request submitter and all commit authors) CLAs are signed, but one or more commits were authored or co-authored by someone other than the pull request submitter.

ydnar:
I haven’t rebased onto current `master` because of the change introduced in 4d8db00641 to use `strings.Cut`, which depends on Go 1.18.

thoro:
Hi,

sorry for posting this here on github, but I have never used Gerrit and it's a bit unclear how todo what!

I just had to use this to produce correct XML for UPNP without resorting to attribute tricks, and it works amazingly.

Since the prefixes are now tracked, it would be logical to allow to define prefixes via tags, it's 5 additional lines of code in the writeStart function (L845):

		if prefix == xmlnsPrefix {
			if prefix, prefixCreated := p.createPrefix(attr.Name.Space, local); prefixCreated {
				p.writePrefixAttr(prefix, attr.Name.Space)
			}
			continue
		}

This would allow to define certain namespace at the top-level object and reuse them everywhere:

	type DidlLite struct {
		XMLName xml.Name `xml:"urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/ DIDL-Lite"`
		UPNP string `xml:"urn:schemas-upnp-org:metadata-1-0/upnp/ xmlns:upnp,attr"`
		DC string `xml:"http://purl.org/dc/elements/1.1/ xmlns:dc,attr"`
		Items []Item `xml:"urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/ item"`
	}

This would allow to make unmarshaled / marshaled xml to be exactly the same!

But I do understand it would be a bit of magic (using the prefix xmlns) and therefore maybe undesirable.

ydnar:
@thoro happy to take a look at your proposal if you can share more context. Perhaps file an issue at @nbio/xml (the fork of `encoding/xml` I maintain)?

thoro:
@ydnar I guess you didn‘t see my edit, but I will suggest it there.

it would just be syntactic sugar, so you won‘t have to initialize the fields of the struct with the correct namespace.

src/encoding/xml/xml.go
```diff
@@ -1158,20 +1180,25 @@ func isInCharacterRange(r rune) (inrange bool) {
...
 		return name, false
-	} else if space, local, ok := strings.Cut(s, ":"); !ok || space == "" || local == "" {
+	} else if i := strings.Index(s, ":"); i < 1 || i > len(s)-2 { // Leading or trailing colon, not OK.
 		name.Local = s
+		return name, false
 	} else {
-		name.Space = space
-		name.Local = local
+		name.Space = s[0:i]
+		name.Local = s[i+1:]
```
DemiMarie, line 1201:
```suggestion
	if i := strings.IndexByte(s, 58); i == -1 { // No colons, no namespace. OK.
		name.Local = s
	} else if i < 1 || i > len(s)-2 { // Leading or trailing colon, not OK.
		name.Local = s
		return name, false
	} else {
		local := s[i+1:]
		if strings.IndexByte(local, 58) != -1 { // Second colon, not OK.
			name.Local = s
			return name, false
		} else {
			name.Space = s[0:i]
			name.Local = local
		}
```

Micro-optimization.

DemiMarie (commented):
Can you also check for #68294, #68295, #68296, and #68297?  These all include test strings.

3052:
it seems the example given here is poor:

https://play.golang.org/p/-6Ee8tcLl2L

in that it does not test marshaling namespaced attributes, such as:

~~~xml
<soap:Envelope xmlns:soap=""></soap:Envelope>
~~~

if you try this:

~~~go
var envelope struct {
   XMLName xml.Name `xml:"soap:Envelope"`
   Soap    string   `xml:"xmlns:soap,attr"`
}
~~~

the attribute namespace is ignored:

~~~xml
<soap:Envelope soap=""></soap:Envelope>
~~~

DemiMarie:
I recommend a much simpler approach: require namespaces to be explicitly represented, and make the encoder namespace-naive.

coyzeng:
6 years ago, why not fixed this bug!

luggesexe:
What is the progress on this one?

PR #48647, opened 2021-09-27
cmd/doc: generate docs for generic functions that return []T

### LINKED ISSUE #48485: cmd/doc: does not generate docs for generic functions that return []T

<pre>
$ gotip version
`devel go1.18-a83a558 Mon Sep 20 00:13:47 2021 +0000`
</pre>

Yes

linux amd64

### What did you do?

I wrote a generic function(inside package), run `gotip doc -all -c -u [package-name]`

expect to see the function signature in output, like `func Intersection[T comparable](slices... []T) []T`

### What did you see instead?

functions that return `[]T` are not printed, others like `func Some[T any](slice []T, pred func(T) bool) bool` are printed

I tried dig into the doc generator, found that
- these functions are now [considered](https://github.com/golang/go/blob/a83a558/src/go/doc/reader.go#L416) to be `factory functions of T`, their associations go `typ.funcs.set` branch instead of `r.funcs.set`
- `cleanupTypes()` [deleted](https://github.com/golang/go/blob/a83a558/src/go/doc/reader.go#L726) T from r.types because it has no definition
- then T is not [included](https://github.com/golang/go/blob/a83a558/src/cmd/doc/pkg.go#L177) in `docPkg.Types`, thus no results are printed

with following ~immature hack~ diff those functions will be printed:
```diff
@@ -429,8 +436,9 @@ func (r *reader) readFunc(fun *ast.FuncDecl) {
                 }
                 // If there is exactly one result type,
                 // associate the function with that type.
-                if numResultTypes == 1 {
+                if numResultTypes == 1 && typ.decl != nil {
```

ahxxm:
The current godoc can parse generic functions that return slice []T,
but will delete them afterwards, due to empty declaration of type T.

This change lets generic functions be top-level functions instead of 
type-T associated functions, making them appear in godoc output.

Fixes #48485

### DIFF
--- a/src/go/doc/example_test.go
+++ b/src/go/doc/example_test.go
@@ -747,3 +747,50 @@ func mustParse(fset *token.FileSet, filename, src string) *ast.File {
 	}
 	return f
 }
+
+func TestGenericFunctions(t *testing.T) {
+	const src = `
+package p
+
+func GenericF1[T any]() (t T)                     { return }
+func GenericF2[T any]() (t []T)                   { return }
+func GenericF3[D comparable, T any]() (m map[D]T) { return }
+
+type MyObject struct{}
+
+// MyObject constructor that use generic type T
+func MyConstructor[T any](i T) MyObject {
+        return MyObject{}
+}
+`
+	// Parse literal source code as a *doc.Package.
+	fset := token.NewFileSet()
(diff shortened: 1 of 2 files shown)


ahxxm:
@googlebot I signed it!

PR #48653, opened 2021-09-27
crypto/ecdsa: Adjust entropy bounds

soatok:
Matching the entropy for an ECDSA signature to the ECDLP attack cost of the curve might make intuitive sense, but the birthday bound for a random function is Q(sqrt(N)), so we're matching a square root of a square root. For P-224 (ECDLP security of 112 bits), this doesn't yield a birthday bound of 112 bits, it yields a birthday bound of **56 bits**.

This patch changes the formula to produce roughly double the expected entropy (by changing the divisor from `16` to `8`), while requiring at least 256 bits. In case of weird cure support, this patch also caps the entropy at 512 bits. The variability between the two will mostly be observed with curves in between the two bounds (i.e. NIST P-384, which many users expect to offer a 192-bit security level anyway).

the Gerrit change.

  + No trailing period
+ No Markdown


soatok:
@googlebot I signed it!

PR #48781, opened 2021-10-04
net/http: mention io.Seeker requirement in FS documentation

thatoddmailbox:
Both FileServer and NewFileTransport can try to seek a file, specifically
when MIME type sniffing is performed. This can be somewhat surprising to an
implementer of an fs.FS, as their filesystem will appear to work until a
user tries to access a file with an unrecognized extension (which requires
type sniffing and therefore seeking). With FileServer, this results in a
"seeker can't seek" message, which is not very clear for the developer.

The issue arises because fs.FS does not require Seek, while http.FileSystem
does. Therefore, this change adds a line to the documentation of net/http's
adapter function mentioning the requirement.

### DIFF
--- a/src/net/http/fs.go
+++ b/src/net/http/fs.go
@@ -815,6 +815,7 @@ func (f ioFile) Readdir(count int) ([]fs.FileInfo, error) {
 
 // FS converts fsys to a FileSystem implementation,
 // for use with FileServer and NewFileTransport.
+// The files provided by fsys must implement io.Seeker.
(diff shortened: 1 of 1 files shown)


oliverpool:
This has been addressed in afd792feb567f0848ddc05f6a90a7b7672fe4974, released in Go 1.20 and visible in the doc: https://pkg.go.dev/net/http#FS

Can probably be closed.

ianlancetaylor:
This change has been merged in Gerrit.  I don't know why this pull request didn't get closed.

PR #48796, opened 2021-10-05
Update url.go

moredure:


ianlancetaylor:
Thanks, but we aren't going to make this kind of change.  It makes the code harder to read and understand, for little benefit.

Issue #43631, opened 2021-01-11
HTTP client should not include raw header values in error messages


eli-darkly:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.15.6 darwin/amd64
</pre>

### Does this issue reproduce with the latest release?

Yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/elibishop/Library/Caches/go-build"
GOENV="/Users/elibishop/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/elibishop/.gvm/pkgsets/go1.15.6/global/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/elibishop/.gvm/pkgsets/go1.15.6/global"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/Users/elibishop/.gvm/gos/go1.15.6"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/Users/elibishop/.gvm/gos/go1.15.6/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/elibishop/work/test/goheaderbug/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/c8/g71crjpd18nc6jkdykcvp_ww0000gn/T/go-build182950335=/tmp/go-build -gno-record-gcc-switches -fno-common"

</pre></details>

### What did you do?

```go
package main

import (
	"fmt"
    "net/http"
)

func main() {
	req, _ := http.NewRequest("GET", "http://worldtimeapi.org/api/timezone", nil)
[...]

PR #48979, opened 2021-10-14
net/http: omit invalid header value from error message

AlexanderYastrebov:
Updates #43631


AlexanderYastrebov:
https://go-review.googlesource.com/c/go/+/355929 was merged.

PR #49024, opened 2021-10-17
net: use Done rather than comparing with context.Background

### LINKED ISSUE #49023: net: allow not creating a goroutine for netFD.connect in case of caller passes context which is not going to be cancelled manually
go1.17.2

Yes

`GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/mikefaraponov/Library/Caches/go-build"
GOENV="/Users/mikefaraponov/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/mikefaraponov/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/mikefaraponov/go"
GOPRIVATE=""
GOROOT="/usr/local/Cellar/go/1.15.6/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.15.6/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/mikefaraponov/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/97/d1hwd7t96mj0_jxmpngf2m5w0000gn/T/go-build319243062=/tmp/go-build -gno-record-gcc-switches -fno-common"`

### What did you do?
Creating TCPConn connection via net.Dialer

No netFD func2 goroutines to be spawned.

### What did you see instead?
Each new connection created create netFD interruptor goroutine

### Solution
net/fd_unix.go

```go
if deadline, hasDeadline := ctx.Deadline(); hasDeadline {
  fd.pfd.SetWriteDeadline(deadline)
  defer fd.pfd.SetWriteDeadline(noDeadline)
}

// Start the "interrupter" goroutine, if this context might be canceled.
// (The background context cannot)
//
// The interrupter goroutine waits for the context to be done and
// interrupts the dial (by altering the fd's write deadline, which
// wakes up waitWrite).
if ctxDone := ctx.Done(); ctxDone != nil {
  // skip creation of goroutine and allow fd.pfd.SetWriteDeadline(deadline) to gracefully close this connection on timeout, this allows to pass custom context.Context implementation with deadline but not cancelable or just check some ctx.Value("bypasscreationofgoroutine") instead of ctx.Done() != nil check may be prefereble
  // Wait for the interrupter goroutine to exit before returning
  // from connect.
  done := make(chan struct{})
  interruptRes := make(chan error)
  defer func() {
	  close(done)
	  if ctxErr := <-interruptRes; ctxErr != nil && ret == nil {
		  // The interrupter goroutine called SetWriteDeadline,
		  // but the connect code below had returned from
		  // waitWrite already and did a successful connect (ret
		  // == nil). Because we've now poisoned the connection
		  // by making it unwritable, don't return a successful
		  // dial. This was issue 16523.
		  ret = mapErr(ctxErr)
		  fd.Close() // prevent a leak
	  }
  }()
  go func() {
	  select {
	  case <-ctxDone:
		  // Force the runtime's poller to immediately give up
		  // waiting for writability, unblocking waitWrite
		  // below.
		  fd.pfd.SetWriteDeadline(aLongTimeAgo)
		  testHookCanceledDial()
		  interruptRes <- ctx.Err()
	  case <-done:
		  interruptRes <- nil
	  }
  }()
}
```
And then smth like this can be used to bypass context cancel mechanism and do not utilise interruptor goroutine:
```go
type deadlineCtx struct {
	deadline time.Time
}

func (m *deadlineCtx) Deadline() (deadline time.Time, ok bool) {
	return m.deadline, true
}

func (m *deadlineCtx) Done() <-chan struct{} {
	return nil
}

func (m *deadlineCtx) Err() error {
	return nil
}

func (m *deadlineCtx) Value(interface{}) interface{} {
	return nil
}
```

moredure:
Fixes golang#49023

### DIFF
--- a/src/net/fd_unix.go
+++ b/src/net/fd_unix.go
@@ -92,12 +92,11 @@ func (fd *netFD) connect(ctx context.Context, la, ra syscall.Sockaddr) (rsa sysc
 	}
 
 	// Start the "interrupter" goroutine, if this context might be canceled.
-	// (The background context cannot)
 	//
 	// The interrupter goroutine waits for the context to be done and
 	// interrupts the dial (by altering the fd's write deadline, which
 	// wakes up waitWrite).
-	if ctx != context.Background() {
+	if ctxDone := ctx.Done(); ctxDone != nil {
 		// Wait for the interrupter goroutine to exit before returning
 		// from connect.
 		done := make(chan struct{})
@@ -117,7 +116,7 @@ func (fd *netFD) connect(ctx context.Context, la, ra syscall.Sockaddr) (rsa sysc
 		}()
 		go func() {
 			select {
-			case <-ctx.Done():
+			case <-ctxDone:
 				// Force the runtime's poller to immediately give up
 				// waiting for writability, unblocking waitWrite
 				// below.


moredure:
Smth like that might be passed to connect
```go

type myCtx struct {
	deadline time.Time
}

func (m *myCtx) Deadline() (deadline time.Time, ok bool) {
	return m.deadline, true
}

func (m *myCtx) Done() <-chan struct{} {
	return nil
}

func (m *myCtx) Err() error {
	return nil
}

func (m *myCtx) Value(key interface{}) interface{} {
	return nil
}
```

PR #49119, opened 2021-10-22
net/http: enable TimeoutHandler custom error response

### LINKED ISSUE #22821: Document http.TimeoutHandler defaults
I was recently suprised to discover that [`http.TimeoutHandler`] returns a blob of HTML (`"<html><head><title>Timeout</title></head><body><h1>Timeout</h1></body></html>"`) as the default message body. It does not attempt to set the appropriate Content-Type header and if you write a response yourself there is no option to set the content type.

The documentation says:

> (If msg is empty, a suitable default message will be sent.)

It would be nice if the format of the default message could be mentioned in the documentation, the default could be removed, or the Content-Type could be set (or some combination of the above).

[`http.TimeoutHandler`]: https://golang.org/pkg/net/http/#TimeoutHandler

### LINKED ISSUE #48948: net/http: TimeoutHandler should distinguish between timeout and requests closed by clients.
> The new Handler calls h.ServeHTTP to handle each request, but if a call runs for longer than its time limit, the handler responds with a 503 Service Unavailable error and the given message in its body. (If msg is empty, a suitable default message will be sent.) After such a timeout, writes by h to its ResponseWriter will return ErrHandlerTimeout.
-- https://pkg.go.dev/net/http#TimeoutHandler

> For incoming server requests, the context is canceled when the client's connection closes, the request is canceled (with HTTP/2), or when the ServeHTTP method returns.
-- https://pkg.go.dev/net/http#Request.Context

If a call runs for **less** than its time limit, but the client cancels the request, TimeoutHandler will respond the same way -- with an ErrHandlerTimeout.

This makes it hard to tell if the code is performing poorly and hitting a timeout, or if the clients are ending requests.

I propose a conditional is added [here](https://cs.opensource.google/go/go/+/refs/tags/go1.17.2:src/net/http/server.go;l=3357), to check if the context's error is a context.DeadlineExceeded or context.Canceled, and return different errors.

AlexanderYastrebov:
Fixes #22821
Fixes #48948

### DIFF
--- a/src/net/http/serve_test.go
+++ b/src/net/http/serve_test.go
@@ -2531,6 +2531,58 @@ func TestTimeoutHandlerPanicRecovery(t *testing.T) {
 	testHandlerPanic(t, false, false, wrapper, "intentional death for testing")
 }
 
+type testHandlerWithTimeout struct{}
+
+func (h *testHandlerWithTimeout) ServeHTTP(ResponseWriter, *Request) {}
+
+func (h *testHandlerWithTimeout) ServeTimeout(w ResponseWriter, r *Request) {
+	if r.Context().Err() == context.DeadlineExceeded {
+		// Handler timeout, customize response
+		// https://github.com/golang/go/issues/22821
+		w.Header().Set("Content-Type", "application/problem+json")
+		w.WriteHeader(503)
+		fmt.Fprintf(w, `{"type": "https://example.net/timeout-error", "trace-id": "%s"}`, r.Header.Get("Trace-Id"))
+	} else {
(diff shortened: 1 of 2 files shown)


mvdan:
https://go-review.googlesource.com/c/go/+/356009/ was merged to fix the second issue, so I imagine this needs to be updated accordingly. As for fixing the first issue, see Damien's reply - new API needs to be discussed in the issue tracker first. I'm closing this for now, as there's nothing we can do here until that happens, and the PR isn't mergeable in its current state in any case.

PR #49127, opened 2021-10-23
crypto: support ECDHE when ec_point_formats is missing in ClientHello

### LINKED ISSUE #49126: crypto/tls: support ECDHE key exchanges when ec_point_formats is missing in ClientHello extension
### What did you do?

Per [rfc8422#section-5.1.2](https://datatracker.ietf.org/doc/html/rfc8422#section-5.1.2), 

> For backwards compatibility purposes, the point format list extension MAY still be included and contain exactly one value: the uncompressed point format (0).  RFC 4492 specified that if this extension is missing, it means that only the uncompressed point format is supported, so interoperability with implementations that support the uncompressed format should work with or without the extension

We are seeing TLS handshake failure (client and server failed to agree on ECDHE_ECDSA key exchange algorithem) when `ec_point_formats` is missing because we expect it to be listed in [tls/handshake_server.go](https://github.com/golang/go/blob/dc00dc6c6bf3b5554e37f60799aec092276ff807/src/crypto/tls/handshake_server.go#L317-L323)

```go
// supportsECDHE returns whether ECDHE key exchanges can be used with this
// pre-TLS 1.3 client.
func supportsECDHE(c *Config, supportedCurves []CurveID, supportedPoints []uint8) bool {
...
	supportsPointFormat := false
	for _, pointFormat := range supportedPoints {
		if pointFormat == pointFormatUncompressed {
			supportsPointFormat = true
			break
		}
	}

	return supportsCurve && supportsPointFormat
}
```

If `ec_point_formats` is missing in ClientHello, we will allow ECDHE key exchanges because  `it means that only the uncompressed point format is supported`

yang-wei:
As describe in rfc8422 5.1.2, we will support ECDHE in the case client does not
include ec_point_formats extension in ClientHello extension. This make sure ECDHE
will work with (uncompressed point format is listed explicitly) or without extension.

rfc8422 5.1.2: https://datatracker.ietf.org/doc/html/rfc8422#section-5.1.2.

Fixes #49126

### DIFF
--- a/src/crypto/tls/handshake_server.go
+++ b/src/crypto/tls/handshake_server.go
@@ -314,7 +314,9 @@ func supportsECDHE(c *Config, supportedCurves []CurveID, supportedPoints []uint8
 		}
 	}
 
-	supportsPointFormat := false
+	// RFC 8422, Section 5.1.2
+	// If this extension is missing, it means that only the uncompressed point format is supported
+	supportsPointFormat := len(supportedPoints) == 0
 	for _, pointFormat := range supportedPoints {
 		if pointFormat == pointFormatUncompressed {
 			supportsPointFormat = true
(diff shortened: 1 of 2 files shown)


torntrousers:
We've hit this in https://github.com/golang/go/issues/53750, any idea when the fix might make it out in a release?

PR #49131, opened 2021-10-23
go/types: propagate conversion error reason if not assignable

### LINKED ISSUE #200: Interface implementation errors are not descriptive.
Running [this program on play.golang.org](http://play.golang.org/p/dIjV7xedya) produces:

```
/tmp/sandbox161431897/main.go:14: cannot convert T literal (type T) to type FooBar:
	T does not implement FooBar (missing Bar method)
```

It tells us that `Bar` is the missing method (i.e. what the error actually is).

But running the same [program via gopherjs.org/play](http://www.gopherjs.org/play/#/nmf16uFkus) produces a error that is very hard to understand:

```
prog.go:14:13: cannot convert (T literal) (value of type T) to FooBar 
```

I actually have code that uses some very large composite interfaces -- and it's hard to track down exactly what I've done wrong (it's code that cannot be run via the standard Go compiler itself, only by gopherjs).

Any ideas on a potential workaround or improvement to this situation?

nevkontakte:
Prior to this change the code below would produce an error `cannot
convert (T literal) (value of type T) to I`:

```
type I interface{ Do() }
type T struct{}

var _ = I(T{})
```

After this change, the error is expanded to include `(missing method
Do)`, which is also what Go compiler outputs.

Fixes gopherjs/gopherjs#200.

### DIFF
--- a/src/go/types/conversions.go
+++ b/src/go/types/conversions.go
@@ -86,7 +86,7 @@ func (check *Checker) conversion(x *operand, T Type) {
 // exported API call, i.e., when all methods have been type-checked.
 func (x *operand) convertibleTo(check *Checker, T Type, reason *string) bool {
 	// "x is assignable to T"
-	if ok, _ := x.assignableTo(check, T, nil); ok {
+	if ok, _ := x.assignableTo(check, T, reason); ok {
 		return true
 	}
 


nevkontakte:
https://go-review.googlesource.com/c/go/+/359137 has fixed this issue while this PR was in review, so I'm retracting it :)

PR #49236, opened 2021-10-30
log: add Llevel flag and OutputL() function

### LINKED ISSUE #49239: log: add Llevel flag and OutputL() function
Some code I am working with take level string:

* Add flag : Llevel
* Add function OutputL:
  OutputL(calldepth int, level, s string) error
  outputl(calldepth int, level, s string) error

I have worked on log4go and nxlog4go for few years. The 3rd log alway face three problem:

1. The interface function. More in stdlog, more need to be modified.
2. Override Output function and replace with 3rd formatter.
3. Building and using own Writer easy.

After compare few solutions, I found:

The level type should be string. It is better than int.
Do not implement too much interface functions in std log. Let the 3rd log do it.
Do not implement level filter in std log. Let the 3rd log do it.

New OutputL function receives the level string and writes it to the beginning of the log line.

ccpaging:
Some code I am working with takes level string:

* Add flag : Llevel
* Add function OutputL:
  OutputL(calldepth int, level, s string) error
  outputl(calldepth int, level, s string) error

Fixes #49239

### DIFF
--- a/src/log/log.go
+++ b/src/log/log.go
@@ -266,7 +266,7 @@ func (l *Logger) Panic(v ...interface{}) {
 // Panicf is equivalent to l.Printf() followed by a call to panic().
 func (l *Logger) Panicf(format string, v ...interface{}) {
 	s := fmt.Sprintf(format, v...)
-	l.OutputL(2, "Panic", s)
+	l.OutputL(2, "Panic ", s)
 	panic(s)
 }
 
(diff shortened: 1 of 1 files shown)


ccpaging:
I have worked on log4go and nxlog4go for few years. The 3rd log alway face three problem:

1. The interface function. More in stdlog, more need to be modified.
2. Override Output function and replace with 3rd formatter.
3. Using own Writer.

After compare few solutions, I found:
1. The level type should be string. It is better than int.
2. Do not implement too much interface functions in std log. Let the 3rd log do it.
3. Do not implement level filter in std log.  Let the 3rd log do it.
3. New OutputL function receives the level string and writes it to the beginning of the log line, so it is the key.

Issue #47694, opened 2021-08-13
doc: write Go 1.18 release notes


cagedmantis:
Tracking bug for writing the Go 1.18 Release Notes.

The latest state on tip can be viewed at https://tip.golang.org/doc/go1.18.

When the Go 1.18 Release Notes are complete, this should be closed, and a similar issue should be made for the next release. Previous issues are #44513, #40700, #37419, #36878, #17929, #15810, #5929, etc.

ianlancetaylor:
> There is a small issue in the first paragraph stating the changes to the testing package: In the last case showing the new precedence rules the operator in the middle should be a pipe '|' and not a slash '/' like in the other cases.

Thanks, https://golang.org/cl/371616 fixes this.

carlmjohnson:
The release notes should mention http.MaxBytesHandler. Is there an issue for this?

heschi:
Only thing left to do is remove the draft notice, i.e. submit https://go.dev/cl/390375, which we'll do during the release process. Removing as a release blocker.

PR #49258, opened 2021-11-01
doc/go1.18: document http.Transport.Dial* being used in js/wasm

neilalexander:
This PR adds a note into the Go 1.18 changelog for CL 330852. 

Updates #46923.


neilalexander:
Related: golang/go#47694

PR #49359, opened 2021-11-04
net/http: add MaxBytesError

earthboundkid:
Fixes #30715


protiumx:
Hi all!
Any update when this will be merged and in which version will it be released?

ianlancetaylor:
@protiumx As discussed over at https://go.dev/cl/361397, this should be in the 1.19 release.  Note that 1.18 is not yet released; 1.19 is scheduled for August 1, 2022.

PR #49409, opened 2021-11-06
runtime: use gcMiscSys for module's gc ptr mask

MoZhonghua:
Memory allocated to materialize gcprog for modules .data and .bss should
be added to memstat.gcMsicSys, not memstats.buckhash_sys.


MoZhonghua:
@googlebot I signed it!

PR #49420, opened 2021-11-07
database/sql:fix SetConnMaxLifetime may not effect immediately

henrykin:
when wake up the cleaner, `db.maxLifetime = d` may not be set. cleaner will determine to clean or not will the old maxLifetime


mvdan:
The CLA wasn't signed here, so I'm closing this for now. Feel free to reopen if/when you've signed it.

PR #49492, opened 2021-11-10
Use common error on `exec.LookPath`

felipecrs:
LookPath is already expected to return the common `exec.ErrNotFound` error, but it does not do that when it is called with a leading `/`.

This treats the error returned by `io/fs`, so it gets properly translated to the expected and common error for such not found scenarios.

This is to prevent workarounds like https://github.com/twpayne/chezmoi/pull/1608.


felipecrs:
@googlebot I signed it!

felipecrs:
Closing as per the comments in Gerrit.

PR #49706, opened 2021-11-21
os: improve file readability

ZuoFuhong:


mvdan:
There was a reply on the CL, argumenting that the change doesn't actually improve the code. There wasn't a response, so I'm closing this for now.

PR #49799, opened 2021-11-25
syscall/js: allocate makeArgs slices on stack

### LINKED ISSUE #39740: syscall/js: Increase performance of Call, Invoke, and New by not allowing new slices to escape onto the heap

<pre>
$ go version
go version go1.14.4 linux/amd64
</pre>

Yes, in go1.15rc1

### What did you do?

I used `js.Call`, `js.New`, and `js.Invoke` with multiple arguments, hundreds of times per frame, 60 frames per second.

The above functions are absolutely essential for getting work done in js/wasm, as they are analogous to a simple function call. They should be optimized as such. No excess go heap memory should be allocated, no extra go garbage collection. After implementing a hack to fix the issue I will go on to describe (see below), here is a profile of memory allocation:
image

Importantly, memory allocation is 8% of the total time and does **not** include `makeArgs` allocating any slices that end up on the heap

### What did you see instead?

2+ new slices allocated on the heap per call to one of the above three functions, as the first few lines of the makeArgs function...
https://github.com/golang/go/blob/60f78765022a59725121d3b800268adffe78bde3/src/syscall/js/js.go#L361-L363

...possibly are assumed to escape to the heap via...
https://github.com/golang/go/blob/60f78765022a59725121d3b800268adffe78bde3/src/syscall/js/js.go#L405-L406
...or `valueNew`/`valueInvoke`, or if go's escape analysis optimization doesn't work on the caller's stack.

A profile shows the significant penalty associated with allocating too many slices (mallocgc now accounts for over 20%), and that doesn't even include the extra garbage collector load that claims a few FPS.
image

I thought of adding `//go:noescape` before each of the above and potentially other functions implemented in javascript, but I didn't get any improvement. Only my hack to `makeArgs` worked consistently:
```go
const maxArgs = 9 // A better hack/fix would grow the slices automatically
var argVals = make([]Value, 0, maxArgs)
var argRefs = make([]ref, 0, maxArgs)

func makeArgs(args []interface{}) ([]Value, []ref) {
	for i, _ := range argVals {
		argVals[i] = Value{}
	}
	for i, _ := range argRefs { // I don't think this is strictly necessary
		argRefs[i] = 0
	}

	argVals = argVals[:len(args)]
	argRefs = argRefs[:len(args)]

	for i, arg := range args {
		v := ValueOf(arg)
		argVals[i] = v
		argRefs[i] = v.ref
	}
	return argVals, argRefs
}
```

Another potential solution, building on the above hack, would be to make `makeArgs` take slices, to put the values/refs in, as arguments (instead of returning new slices). The caller could deal with efficiently handling the memory. Maybe this could help go's escape analysis, in conjunction with `//go:noescape`.

finnbear:
This change optimizes the `syscall/js` package such that:
- Calling a JavaScript function with 16 arguments or fewer will not induce two additional heap allocations, at least with the current Go compiler
- Using `syscall/js` features with slices and strings of statically-known length will not cause them to be escaped to the heap, at least with the current Go compiler
- The reduction in allocations has the additional benefit that the garbage collector runs less often, blocking WebAssembly's one and only thread less often

There should be no observable difference in the `syscall/js` API.

These changes have been discussed and mostly agreed upon in the linked issue.

Fixes #39740

I am aware that this change falls outside a release cycle and during a development freeze. I will leave it up to maintainers to decide whether this qualifies for a freeze exception.

### DIFF
--- a/src/syscall/js/js.go
+++ b/src/syscall/js/js.go
@@ -209,6 +209,7 @@ func ValueOf(x interface{}) Value {
 	}
 }
 
+//go:noescape
 func stringVal(x string) ref
 
 // Type represents the JavaScript type of a Value.
@@ -292,6 +293,7 @@ func (v Value) Get(p string) Value {
 	return r
 }
 
+//go:noescape
 func valueGet(v ref, p string) ref
 
 // Set sets the JavaScript property p of value v to ValueOf(x).
@@ -306,6 +308,7 @@ func (v Value) Set(p string, x interface{}) {
 	runtime.KeepAlive(xv)
 }
 
+//go:noescape
 func valueSet(v ref, p string, x ref)
 
 // Delete deletes the JavaScript property p of value v.
@@ -318,6 +321,7 @@ func (v Value) Delete(p string) {
 	runtime.KeepAlive(v)
 }
 
+//go:noescape
 func valueDelete(v ref, p string)
 
 // Index returns JavaScript index i of value v.
@@ -347,15 +351,29 @@ func (v Value) SetIndex(i int, x interface{}) {
 
 func valueSetIndex(v ref, i int, x ref)
 
-func makeArgs(args []interface{}) ([]Value, []ref) {
-	argVals := make([]Value, len(args))
-	argRefs := make([]ref, len(args))
+func makeArgs(args []interface{}) (argVals []Value, argRefs []ref) {
+	// value chosen for being power of two, and enough to handle all web APIs
+	// in particular, note that WebGL2's texImage2D takes up to 10 arguments
+	const maxStackArgs = 16
+	if len(args) <= maxStackArgs {
+		// as long as makeArgs is inlined, these will be stack-allocated
+		argVals = make([]Value, len(args), maxStackArgs)
+		argRefs = make([]ref, len(args), maxStackArgs)
+	} else {
+		// allocates on the heap, but exceeding maxStackArgs should be rare
+		argVals = make([]Value, len(args))
+		argRefs = make([]ref, len(args))
+	}
+	return
+}
+
+func storeArgs(args []interface{}, argValsDst []Value, argRefsDst []ref) {
+	// would go in makeArgs if the combined func was simple enough to inline
 	for i, arg := range args {
 		v := ValueOf(arg)
-		argVals[i] = v
-		argRefs[i] = v.ref
+		argValsDst[i] = v
+		argRefsDst[i] = v.ref
 	}
-	return argVals, argRefs
 }
 
 // Length returns the JavaScript property "length" of v.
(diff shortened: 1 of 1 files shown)


src/syscall/js/js.go
```diff
@@ -580,4 +604,5 @@ func CopyBytesToJS(dst Value, src []byte) int {
 	return n
 }
 
+//go:noescape
```
finnbear, line 607: Justification: this will affect whether byte slices with dynamic contents but statically known length to not escape anymore.

src/syscall/js/js.go
```diff
@@ -306,6 +308,7 @@ func (v Value) Set(p string, x interface{}) {
 	runtime.KeepAlive(xv)
 }
 
+//go:noescape
```
finnbear, line 311: Justification: this will affect whether strings with dynamic contents but statically known length to not escape anymore.

src/syscall/js/js.go
```diff
@@ -292,6 +293,7 @@ func (v Value) Get(p string) Value {
 	return r
 }
 
+//go:noescape
```
finnbear, line 296: Justification: this will affect whether strings with dynamic contents but statically known length to not escape anymore.

src/syscall/js/js.go
```diff
@@ -209,6 +209,7 @@ func ValueOf(x interface{}) Value {
 	}
 }
 
+//go:noescape
```
finnbear, line 212: Justification: this will affect whether strings with dynamic contents but statically known length to not escape anymore.

src/syscall/js/js.go
```diff
@@ -318,6 +321,7 @@ func (v Value) Delete(p string) {
 	runtime.KeepAlive(v)
 }
 
+//go:noescape
```
finnbear, line 324: Justification: this will affect whether strings with dynamic contents but statically known length to not escape anymore.

src/syscall/js/js.go
```diff
@@ -410,13 +431,15 @@ func (v Value) Invoke(args ...interface{}) Value {
 	return makeValue(res)
 }
 
+//go:noescape
```
finnbear, line 434: Justification: this is essential for the slice optimization

src/syscall/js/js.go
```diff
@@ -429,6 +452,7 @@ func (v Value) New(args ...interface{}) Value {
...
```
finnbear, line 455: Justification: this is essential for the slice optimization

src/syscall/js/js.go
```diff
@@ -391,13 +410,15 @@ func (v Value) Call(m string, args ...interface{}) Value {
...
```
finnbear, line 413: Justification: this is essential for the slice optimization

finnbear:
This is superseded by https://github.com/golang/go/pull/66684 :rocket:

PR #49818, opened 2021-11-27
math/rand: improve performance by adding rand.Int31n(n) into rand.Int63n(n)

songoow:
Int31n(n) is faster than Int63n(n) when 0<n<=1-31-1 . while we use Int63n(n) to generate a pseudo-random number in this scenario which 0<n and n<=1<<31-1,using Int31n(n) is more recommended.


liu-song:
@googlebot  I signed it!

PR #50054, opened 2021-12-09
net/http: Add rawurl request support, it can act as `curl --path-as-is`.

purpleroc:
**net/http: Add rawurl request support, it can act as `curl --path-as-is`.**

I can't send a request to server with uri like: 
"http://127.0.0.1/aabb/#/..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f/etc/hosts".
The url.Path will be: /aabb/

When I forced set url.Path="/aabb/#/..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f..%2f/etc/hosts"

Server will recive req.path like : 
"/aabb/%23/..%252f..%252f..%252f..%252f..%252f..%252f..%252f..%252f..%252f/etc/hosts"

But I just want send a request like `curl --path-as-is`, without any fragment handle and urlencode.

```
Usage:
	req, err := http.NewRequest("GET", _url, nil)
	req.PathAsIs.Flag = true
	req.PathAsIs.RawUrl = _url
```

### DIFF
--- a/src/net/http/request.go
+++ b/src/net/http/request.go
@@ -94,6 +94,11 @@ var reqWriteExcludeHeader = map[string]bool{
 	"Trailer":           true,
 }
 
+type PathAsIs struct {
+	Flag   bool
+	RawUrl string
+}
+
 // A Request represents an HTTP request received by a server
 // or to be sent by a client.
 //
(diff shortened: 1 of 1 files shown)


jimen0:
This should go through the Proposal process first as it is a public facing API change.

https://github.com/golang/proposal#the-proposal-process

mvdan:
Closing per https://github.com/golang/go/pull/50054#issuecomment-991644459.

PR #50089, opened 2021-12-10
cmd/go/internal/load: prevent calling ImportErrorf when the err is *module.InvalidPathError

SilverRainZ:
1. The call of setError is unnecessary bacause setError do nothing when
p.Error is not nil.
2. The call of ImportErrorf may panic in this situation
(double quotes appear in the import path).

Fixes #49137


SilverRainZ:
Why this PR is not imported to Gerrit?

ianlancetaylor:
> Why this PR is not imported to Gerrit?

See https://golang.org/issue/49696.

SilverRainZ:
/comments Regression test has been added : )

SilverRainZ:
@gopherbot /comments Regression test has been added : )

SilverRainZ:
@bcmills Any progress?

PR #50128, opened 2021-12-13
runtime: typo fix name of comment

clamyang:
becuse --> because


clamyang:
@googlebot I signed it!

clamyang:
this pr commit info was wrong email and username so cla was faied..

PR #50171, opened 2021-12-14
all: gofmt -w -r 'interface{} -> any' src

JacksonKearl:
Example PR to see what this commit would look like in github.dev.


JacksonKearl:
Ahh, my bad. I meant to base against my own repo sorry all.

PR #50240, opened 2021-12-17
crypto/x509: Add android user trusted CA folder

jbpin:
User can trust new CA on android but it seems that go build package are not able to use it.

This PR will add the folder where user CA trusted certificate is added to.


jbpin:
I've signed CLA please rescan :)

PR #50258, opened 2021-12-18
src/fmt: misuse of different kind of ticks together is corrected

ufukty:
misuse of different kind of ticks together is corrected


ufukty:
I closed #50257 after I signed the CLA and opened this PR.

PR #50330, opened 2021-12-23
database/sql: Remove redundant type conversion

sashamelentyev:


sashamelentyev:
@googlebot I signed it!

PR #50355, opened 2021-12-27
net/http: add client option to retain the auth/cookie headers

notrynosuccess:
I added an option attribute RetainHeaders to the client struct, it indicates whether to keep the auth/cookie headers(Authorization, Www-Authenticate, Cookie, Cookie2 in function shouldCopyHeaderOnRedirect) when redirect across different primary domains.

For example, if you want to retain the auth/cookie headers after redirecting from a.com to b.com, set it to true.

In fact, this is just needed when the system migrates to a new primary domain name and is compatible with downstream applications (redirection is configured), especially for some interfaces that require secure header parameters, such as oauth2 / token, these headers cannot be discarded.

### DIFF
--- a/src/net/http/client.go
+++ b/src/net/http/client.go
@@ -104,6 +104,10 @@ type Client struct {
 	// RoundTripper implementations should use the Request's Context
 	// for cancellation instead of implementing CancelRequest.
 	Timeout time.Duration
+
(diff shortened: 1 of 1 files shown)


hu-md:
the great ...

carlmjohnson:
This is already handled by CookieJars.

PR #50358, opened 2021-12-27
database/sql: predictable request queue satisfaction

alzasr:
change random satisfaction of the query queue to sequential

test simulate situation when query append faster that execution

Sitaution:

I have 2 source of query:
  - the first source generates queries often. These queries are not important and can be canceled by timeout. Contexts have timeout
  - the second source generates queryes less often. These queries are important and must be executed. Contexts dont have timeout;
 
Trouble:
The connection poll satisfied request by random. This behavior does not guarantee the execution of the request in a finite time. I am faced with situation an unlimited growth of latency.

I make test to demonstration my situation.

Results of random satisfaction:

delay(diff = 5m30.844350115s, expected = 1m58.8s, max = 7m29.644350115s), queue(len=298, max=1486, min=364, diff=1122)
sql_db_test.go:51: current delay 7m34.568549607s is 2 times greater than expected 1m59.6s
--- FAIL: TestDB_queryQueue (600.00s)

FAIL

Results of sequenced satisfaction:

delay(diff = 344.00638ms, expected = 1m58.8s, max = 1m59.14400638s), queue(len=298, max=1486, min=1189, diff=297)
--- PASS: TestDB_queryQueue (600.00s)
PASS

queue.len - length of request to execution
queue.max - max index of request from queue
queue.min - min index of request from queue
queue.diff = queue.max - queue.min
If queue.diff is much longer that queue.len then queue has request with big latency

delay.expected = (queue.len - 1) * appendReuqestDelay
delay.max - latency of older request
delay.diff = delay.max - delay.expected

Solution:
Change the connection behavior from random to sequenced.

### DIFF
--- a/src/database/sql/sql.go
+++ b/src/database/sql/sql.go
@@ -464,11 +464,12 @@ type DB struct {
 	// connections in Stmt.css.
 	numClosed uint64
 
-	mu           sync.Mutex // protects following fields
-	freeConn     []*driverConn
-	connRequests map[uint64]chan connRequest
-	nextRequest  uint64 // Next key to use in connRequests.
-	numOpen      int    // number of opened and pending open connections
+	mu                   sync.Mutex // protects following fields
+	freeConn             []*driverConn
+	connRequests         map[uint64]chan connRequest
+	nextRequest          uint64 // Next key to use in connRequests.
+	nextRequestToSatisfy uint64 // Next key to satisfy from connRequests.
+	numOpen              int    // number of opened and pending open connections
 	// Used to signal the need for new connections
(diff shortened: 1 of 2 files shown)


dmitshur:
This PR is currently targeting the "release-branch.go1.17" branch. Go 1.17 is no longer supported per the [release policy](https://go.dev/doc/devel/release#policy) and so we don't use the `release-branch.go1.17` branch for anything now.

If this is a change you'd still like to send, please see https://go.dev/doc/contribute#before_contributing and send it again targeting the branch named `master` instead. Thanks.

PR #50370, opened 2021-12-27
os/env: provide a way to get a default value if env variable is not present

pedromassango:
This is useful when loading non-strings and we need to convert that same value to something else (usually int64), to avoid methods like strconv.ParseInt(...) from crashing if the string is empty, we can use this method to automally asign a default value (let's say zero or 123)


mvdan:
Closing per Ian's reply on the CL.

PR #50443, opened 2022-01-05
cmd/go: improve documentation for relative import paths

jatron:
Relative imports do not work when using Go modules. Update the
documentation to reflect this.


jatron:
Let's abandon this change and chat on https://go-review.googlesource.com/c/go/+/375535 instead.

PR #50650, opened 2022-01-17
regexp: add the missing is

luochuanhang:


luochuanhang:
I signed it!

luochuanhang:
Yes, this may not be a fix, but adding the is makes it more readable for  non-native speakers

PR #50739, opened 2022-01-21
strings/clone: pretty comments

callthingsoff:
Add commas for better reading.


bejandiana (commented):
The comma is not mandatory, a better way of achieving a better readability would be to use 
`The string "" will be returned for strings of length zero`

mvdan:
There was a reply on the CL, argumenting that the change doesn't actually improve the code. There wasn't a response, so I'm closing this for now.

PR #50771, opened 2022-01-24
pprof: support limit the max number for dumping goroutines.

doujiang24:
pprof may take too much time when there are many goroutines.
The new SetMaxDumpGoroutineNum API set the max goroutine number for dumping.


doujiang24:
Now, the `SetMaxDumpGoroutineNum` API introduced a new global variable.
This may not be a good idea, but I don't know the better way.

PR #50843, opened 2022-01-27
io: add WriterTo support to MultiReader

Jorropo:
This fixes #50842.

### DIFF
--- a/src/io/multi_test.go
+++ b/src/io/multi_test.go
@@ -63,6 +63,26 @@ func TestMultiReader(t *testing.T) {
 	})
 }
 
+func TestMultiReaderWriterTo(t *testing.T) {
+	mr := MultiReader(
+		strings.NewReader("foo "),
+		strings.NewReader(""),
+		strings.NewReader("bar"),
+	)
+	mrAsWriterTo := mr.(WriterTo) // Checked at compilation time in multi.go
+	sink := bytes.NewBuffer(nil)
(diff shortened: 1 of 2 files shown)


Jorropo:
I don't really like the way `Copy` is internally used.

It would better if we could call `copyBuffer` with a buffer to avoid allocating a buffer per sub `Reader`, reusing a single one instead.
However blindly allocating something would lead to useless allocations if all sub `Reader`s supports zero copy.

I think `copyBuffer` should take a pointer to slice for a buffer or return the slice it has grown inside.
Then we could lazily allocate the buffer and reuse it.
I have not made that change because I think it is outside of the scope of this PR.

src/io/multi_test.go
```diff
@@ -63,6 +63,26 @@ func TestMultiReader(t *testing.T) {
...
+	sink := bytes.NewBuffer(nil)
+	n, err := mrAsWriterTo.WriteTo(sink)
+	if err != nil {
+		t.Fatalf("expected no error; got %v", err)
+	}
+	if n != 7 {
+		t.Errorf("expected red 7 bytes; got %d", n)
+	}
+	if result := string(sink.Bytes()); result != "foo bar" {
+		t.Errorf("expected red 7 bytes; got %d", n)
```
TheGUNNER13, line 82: should the error message say that the output is mis matched ?

Jorropo:
I'll resend a fixed patch for the go1.19 merge window.

Issue #19420, opened 2017-03-06
net/http: ServeContent doesn't set Content-Length when Content-Encoding is set


mbertschler:
### What version of Go are you using (`go version`)?
`go version go1.8 darwin/amd64`

### What did you do?

A HTTP request served by [caddy](https://github.com/mholt/caddy) for `/myfile`. If the request is sent with accept encoding gzip, caddy has a mechanism that looks up if a file `/myfile.gz` is present in the folder, and serves that directly instead of gzipping the uncompressed `myfile`.

Relevant lines are starting here:
https://github.com/mbertschler/caddy/blob/2649f0330e3d4ab3395618eeb2c6705a5d240efa/caddyhttp/staticfiles/fileserver.go#L203

### What did you expect to see?

A HTTP response that has a Content-Length header set.

### What did you see instead?

The HTTP response doesn't have a Content-Length header set.

This happens because the serveContent function only sets Content-Length if Content-Encoding is empty. In caddy's case it is set to gzip before ServeContent is called. This code was written 5 years ago by @bradfitz when ServeContent was first added. Do you still know why this check was added?

https://github.com/golang/go/blob/a610957f2e48c5b917126b1d0b5f930f05ca8841/src/net/http/fs.go#L284

### What operating system and processor architecture are you using (`go env`)?
```bash
$ go env
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/mbertschler/go:/Users/mbertschler/wcd/go"
GORACE=""
GOROOT="/usr/local/Cellar/go/1.8/libexec"
GOTOOLDIR="/usr/local/Cellar/go/1.8/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/d0/zv0zzgf90r9bds00x6rm_qbr0000gn/T/go-build936143375=/tmp/go-build -gno-record-gcc-switches -fno-common"
CXX="clang++"
CGO_ENABLED="1"
PKG_CONFIG="pkg-config"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
[...]

PR #50904, opened 2022-01-29
net/http: set/override Content-Length for encoded range requests

mitar:
Currently, http.ServeContent returns invalid Content-Length header if:

* Request is a range request.
* Content is encoded (e.g., gzip compressed).
* Content-Length of the encoded content has been set before calling
  http.ServeContent, as suggested in https://github.com/golang/go/issues/19420.

Example:

	w.Header().Set("Content-Type", "application/json")
	w.Header().Set("Content-Length", strconv.Itoa(len(compressedJsonBody)))
	w.Header().Set("Content-Encoding", "gzip")
	w.Header().Set("Etag", etag)
	http.ServeContent(
		w, req, "", time.Time{},
		bytes.NewReader(compressedJsonBody),
	)

The issue is that http.ServeContent currently sees Content-Length as
something optional when Content-Encoding is set, but that is a problem
with range request which can send a payload of different size. So this
reverts https://go.dev/cl/4538111 and makes Content-Length be set
always to the number of bytes which will actually be send (both for
range and non-range requests).

Without this fix, this is an example response:

	HTTP/1.1 206 Partial Content
	Accept-Ranges: bytes
	Content-Encoding: gzip
	Content-Length: 351
	Content-Range: bytes 100-350/351
	Content-Type: application/json; charset=UTF-8
	Etag: "amCTP_vgT5PQt5OsAEI7NFJ6Hx1UfEpR5nIaYEInfOA"
	Date: Sat, 29 Jan 2022 14:42:15 GMT

As you see, Content-Length is invalid and should be 251.

### DIFF
--- a/src/net/http/fs.go
+++ b/src/net/http/fs.go
@@ -343,9 +343,7 @@ func serveContent(w ResponseWriter, r *Request, name string, modtime time.Time, 
 	}
 
 	w.Header().Set("Accept-Ranges", "bytes")
-	if w.Header().Get("Content-Encoding") == "" {
-		w.Header().Set("Content-Length", strconv.FormatInt(sendSize, 10))
-	}
+	w.Header().Set("Content-Length", strconv.FormatInt(sendSize, 10))
 
 	w.WriteHeader(code)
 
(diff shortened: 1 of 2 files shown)


mitar:
@neild @rsc: Can I help to get this change somehow in? Is there anything more from me to do?

PR #50908, opened 2022-01-29
os.types: fixup ModePerm comment

jtagcat:
Unix permission bits: 0o777 → 0777


mvdan:
The current syntax is correct, see https://go.dev/doc/go1.13#language.

PR #50983, opened 2022-02-02
cmd/compile: correct doc comment name for OrderedOrEqual

pierwill:


pierwill:
I did a force push. Not sure if that was the right thing to do.

PR #51396, opened 2022-02-28
io: add WriterTo to MultiReader

Jorropo:
Fixes #50842.

Second patch with comments from #50843 fixed and lazy buffer reusing.


Jorropo:
Replaced by #51502

PR #51612, opened 2022-03-11
syscall: Add double quotes if parameter contain special symbols

pzx521521:
i want to run cmd:`test.exe "a&b&c"`

if i run exec.Command("./test.exe", '"a&b&c"'):
appendEscapeArg will change it to `test.exe \"a&b&c\"`

if i run exec.Command("./test.exe", "a&b&c")
appendEscapeArg will not change it: `test.exe a&b&c`
it shoule be `test.exe "a&b&c"`

### DIFF
--- a/src/syscall/exec_windows.go
+++ b/src/syscall/exec_windows.go
@@ -106,7 +106,7 @@ func makeCmdLine(args []string) string {
 		if len(b) > 0 {
 			b = append(b, ' ')
 		}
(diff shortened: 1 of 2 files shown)


pzx521521:
syscall/exec_windows/exec_windows.go: Fix the EscapeArg of makeCmdLine

pzx521521 (commented):
Add double quotes if parameter contain special symbols

PR #51905, opened 2022-03-24
os/env: Getenv add defaultValue

batara666:
```go
os.Getenv("MONGO_URI", "mongodb://127.0.0.1:3306")
```

reason: 

1. Python has it, and It's really useful
2. this changes will not break old code (but may make it more tight in the future)


seankhliao:
This is a breaking api change, and as such won't be accepted.
Any api change should go through the proposal process: https://go.dev/s/proposal

batara666:
Hi, nice to know you @seankhliao

PR #51961, opened 2022-03-26
io/ioutil: add deprecation markers to ioutil

jeffwidman:
All the code in ioutil just forwards functionality to code
in either the io or os packages, per issue 42026.

This change adds the "Deprecated" marker to all the
functions in this package.

For #42026

Fixes #51927

### DIFF
--- a/src/io/ioutil/ioutil.go
+++ b/src/io/ioutil/ioutil.go
@@ -4,7 +4,7 @@
 
 // Package ioutil implements some I/O utility functions.
 //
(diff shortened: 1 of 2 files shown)


jeffwidman:
@dlsniper FYI I cherry-picked your original take on this from https://github.com/golang/go/pull/43799 to ensure you get credit and then added a commit for the deprecations in `ioutil/tempfile.go` since that file changed a little since your original PR.

PR #52121, opened 2022-04-03
net/http: trim cookie names

darmiel:
The current implementation ignores cookies where the 
cookie name starts or ends with a space. For example,

name =value 

is ignored.

I have come across pages that send cookies in this weird format.
I tested with the latest versions of Firefox, Safari and Chrome, 
all of which accept cookies in this format. 

To do this, I remove leading and trailing spaces from the 
cookie name after cutting at '='.


TomRomeo (approved):

darmiel:
Thanks! I've moved `textproto.TrimString` after checking for `strings.Cut` and added a comment to the tests

PR #52122, opened 2022-04-03
Update go_spec.html: Added a comma for clarification

jessicasmith666:


zamicol:
Thanks for pointing that out. 

I'm still not clear on the meaning of the sentence even with the comma.  

Removing the parenthetical expression results in a sentence that doesn't make sense, "For a type parameter which is always an interface." 

I suspect it was suppose to say: 

"The underlying type of a type parameter that is the underlying type of its type constraint is always an interface." That would make sense considering the later sentence in the section: "The underlying type of P is interface{}"

PR #52136, opened 2022-04-04
encoding/json: add `...` json tag to spread it in struct

maru44:
I added new json tag.
This can be instead of embedded generics.

As you know, we cannot embed generics to struct.

**this is invalid**
```go
type bad[T any] struct{
    Name string
    T
}
```

I want use embed in json, so I added new json tag to spread struct in its parent.

ex)
```go
type good[T any] struct {
    Name string
    Free T `json:"..."`
}

type more struct {
    Age int
    Country string
}

var foo = good[more]{
    Name: "Foo",
    more: more{
        Age: 20,
        Country: "U.S.A",
    },
}

```

If you encode it in json. It'll be like this.
```
{"Name": "Foo", "Age": 20, "Country": "U.S.A"}
```

### DIFF
--- a/src/encoding/json/encode.go
+++ b/src/encoding/json/encode.go
@@ -1239,6 +1239,8 @@ func typeFields(t reflect.Type) structFields {
 
 			// Scan f.typ for fields to include.
 			for i := 0; i < f.typ.NumField(); i++ {
+				var spread bool
+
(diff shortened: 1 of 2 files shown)


mvdan:
Closing per the CL comments - this needs a proposal, but one hasn't been filed yet.

PR #52147, opened 2022-04-04
bytes: limit allocation in SplitN

catenacyber:
So that bytes.SplitN("", "T", int(144115188075855872)) does not panic.


catenacyber:
As was done for strings in https://github.com/golang/go/pull/51755 https://go-review.googlesource.com/c/go/+/393654

catenacyber:
Found by oss-fuzz with ngolo-fuzzing : https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46117

PR #52186, opened 2022-04-06
crypto/tls: optimise handshakeContext

moredure:
Looks like context.WithCancel only needed in some corner cases when some callbacks from tls.Config executed, maybe it makes sense to the user to cancel such a context by them selves?


moredure:
But no https://github.com/golang/go/commit/860704317e02d699e4e4a24103853c4782d746c1

PR #52274, opened 2022-04-11
crypto/tls: avoid creating handshake context on each read/write

moredure:
handshakeCtx and cancel function always allocated on read and write operation not depending on whether the handshake is completed


dt:
I believe there is a similar CL here: https://golang.org/cl/379034

moredure:
@dt, I guess so

PR #52362, opened 2022-04-14
cmd/dist: use gohostarch for ssa rewrite check 

paralin:
Fix a build failure when bootstrapping the Go compiler with go-bootstrap 1.4
while the environment contains GOARCH=riscv64.

Building Go toolchain1 using go-1.4-bootstrap-20171003.
src/cmd/compile/internal/ssa/rewriteRISCV64.go:4814
invalid operation: y << x (shift count type int64, must be unsigned integer)

This is because:

 - buildtool.go:198: calls bootstrapRewriteFile(src)
 - bootstrapRewriteFile: buildtool.go:283 calls:
 - isUnneededSSARewriteFile: checks os.Getenv("GOARCH")
 - isUnneededSSARewriteFile: returns "", false
 - bootstrapRewriteFile: calls bootstrapFixImports
 - boostrapFixImports: generates code go1.4 cannot compile

Instead of checking "GOARCH" in the environment, use the gohostarch variable.


paralin:
Fixes https://github.com/golang/go/issues/52583

paralin:
Edited PR title & body to match commit message.

Issue #41949, opened 2020-10-13
crypto/x509: IsEncryptedPEMBlock returns false on valid encrypted keys. ParseRawPrivateKeyWithPassphrase fails on PKCS8 format encrypted key.


HarikrishnanBalagopal:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.15.2 darwin/amd64
</pre>

### Does this issue reproduce with the latest release?
YES

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/harikrishnanbalagopal/Library/Caches/go-build"
GOENV="/Users/harikrishnanbalagopal/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/harikrishnanbalagopal/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/harikrishnanbalagopal/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.15.2/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.15.2/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/harikrishnanbalagopal/go/src/github.com/konveyor/testkey/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/09/5yjxv27n6njfskvvmkv9v8m40000gn/T/go-build525969776=/tmp/go-build -gno-record-gcc-switches -fno-common"
GOROOT/bin/go version: go version go1.15.2 darwin/amd64
[...]

Issue #32777, opened 2019-06-25
crypto/x509: replace MD5 in PEM encryption?


micahhyman1:
The `deriveKey` function in crypto/x509 uses an md5 hash to derive the encryption key. This hash is out-of-date, insecure and non FIPS compliant. It should be replaced with sha-256.

FiloSottile:
That's not a change we can make unilaterally. It would make it impossible to decrypt PEM files encrypted by OpenSSL or by previous versions of Go.

Thankfully, it's not being used as a collision-resistant hash, but as a key derivation function, and MD5 is not broken for that purpose.

I will look into the spec to see if there's anything we can change.

micahhyman1:
Could it be made an option on the encryption function to choose which hash is used, the same way one can choose which cipher to use?

lucasmoten:
I think that could be added as _additional_ convenience functions in crypto/x509/pem_decrypt.go. 
Similar to how openssl extended to support password based key derivation function. Theoretically, the current implementation of DecryptPEMBlock wouldn't work with a cert encrypted via 
`openssl enc -aes-256-cbc -md sha512 -pbkdf2 -iter 1000 -salt -in plaintext.pem -out encrypted.pem`
Establishing additional functions that support the inputs openssl offers for specifying message digest and iteration count etc would be an improvement.

PR #52384, opened 2022-04-16
crypto/x509: removed oracle from DecryptPEMBlock. removed deprecations

bustedware:
removed IsEncryptedPEMBlock, EncryptPEMBlock, and DecryptPEMBlock
from becoming deprecated


bustedware:
This PR is an attempt at scoping the problem of oracle attacks and to prevent IsEncryptedPEMBlock, EncryptPEMBlock, and DecryptPEMBlock from becoming deprecated in future releases of golang. I will address the deprecation message and attempt to resolve any concerns with keeping it in future releases

The message included alongside DecryptPEMBlock deprecation indicates the routine is insecure since it does not authenticate the ciphertext and is vulnerable to padding oracle attacks.
```
// Deprecated: Legacy PEM encryption as specified in RFC 1423 is insecure by
// design. Since it does not authenticate the ciphertext, it is vulnerable to
// padding oracle attacks that can let an attacker recover the plaintext.
```

Oracle attacks rely on 2 assumptions ([page 9](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.711.2723&rep=rep1&type=pdf))
- An adversary can eavesdrop the communication and intercept the CBC-encrypted ciphertext
- An adversary has access to a padding oracle O, which means he or she can distinguish between VALID and INVALID paddings

The attack relies on leaking the ciphertext. The routine DecryptPEMBlock should not be deprecated because it still adds value. Without encrypted PEM blocks we would be left with plaintext - which is what the deprecation message mentions an attacker would be able to recover. Defense in depth, also known as layered defense, is a security principle where single points of complete compromise are eliminated or mitigated by the incorporation of a series or multiple layers of security safeguards and risk-mitigation countermeasures ([reference](https://github.com/OWASP/DevGuide/blob/master/02-Design/01-Principles%20of%20Security%20Engineering.md)). Encrypting PEM blocks, although sometimes thwarted by a clever attacker, adds to our layers of defense.

1. Included in this PR is the removal of the "Invalid Padding" error message
Removing the oracle (in this case the "Invalid Padding" error message) will strengthen the protection among the golang community who leverage this routine. This helps mitigate an attacker from determining VALID from INVALID padding for an applications usage of this routine. An attacker may still choose to use another oracle but will still need to have access to or depend on leaked ciphertext. This effectively address the second point "it is vulnerable to padding oracle attacks" from the deprecation message.

References for oracle attacks:
"The error messages are the crucial basis for the attack" ([page 6](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.711.2723&rep=rep1&type=pdf))
"Removing the oracle would also prevent the attack" ([under "Defenses" second paragraph](https://research.nccgroup.com/2021/02/17/cryptopals-exploiting-cbc-padding-oracles/))
"The oracle could be something as simple as returning a value that says "Invalid padding" or something more complicated like taking a measurably different time to process a valid block as opposed to an invalid block." ([under "Introduction" fourth paragraph](https://docs.microsoft.com/en-us/dotnet/standard/security/vulnerabilities-cbc-mode))

2. To address the first point in the deprecation message "Since it does not authenticate the ciphertext"
PEM blocks are not supplemented with any metadata to perform said authentication. Consider that authentication of the ciphertext relies on a signature which can only be produced by a signer. For example, TLS relies on both the client and server to agree upon a common CA for cipher authentication. A clever golang developer can still authenticate the ciphertext outside the scope of DecryptPEMBlock routine and may choose to not DecryptPEMBlock against a ciphertext which is not accompanied with a verifiable signature. Golang developers should be empowered with the choice of whether the integrity of the ciphertext is important to their application and whether or not they should invoke the DecryptPEMBlock routine. This should also address the concerns with EncryptPEMBlock deprecation message

ianlancetaylor:
@bustedware We don't use GitHub for code review, so almost nobody will see your comments on this conversation.  We use Gerrit, as explained in the automatically added comment above.  The place to reach out to reviewers is https://go.dev/cl/400655.  I'll ping that change now.

ianlancetaylor:
Actually, this needs more info.  I've commented on https://go.dev/cl/400655; please follow up there.  Thanks.

bustedware:
@ianlancetaylor thanks for helping point me in the right direction. i think some of the formatting got lost when I added it there but I wanted to make a comment here for completeness. I provided my first full description in the comments here (4th comment)

Here is what I just added regarding the 2 issues you linked. 
https://github.com/golang/go/issues/32777
FiloSottile mentions that md5 is used in the key derivation function, and that MD5 is not broken for that purpose.

https://github.com/golang/go/issues/41949
FiloSottile mentions in one comment "That encryption format is legacy and broken by design, so we should deprecate it, not mix it with newer formats that encourage its use" but does not provide any context about why its broken by design. I can only assume those concerns are reflected in the deprecation message which I will address 

Thanks again. I'll continue correspondence on https://go-review.googlesource.com/c/go/+/400655

Issue #13492, opened 2015-12-05
runtime: c-shared builds fail with musllibc


mdempsky:
Currently, some of the init_array functions provided by a c-shared build expect to be called with (argc, argv, envp) arguments as is done by glibc, but this isn't specified by the ELF gABI for DT_INIT_ARRAY (http://www.sco.com/developers/gabi/latest/ch5.dynamic.html#init_fini), and isn't done with other libc implementations like musllibc.

CC @ianlancetaylor @mwhudson

jamesr:
musl does expose getauxval() for querying the auxiliary vector but does not provide any way to get the main program's argc/argv.  The musl maintainers argue that this isn't something a shared library should have access to anyway.

mdempsky:
Somewhat relatedly, C99 says [section 5.1.2.2.1]:

> The parameters `argc` and `argv` and the strings pointed to by the `argv` array shall be modifiable by the program, and retain their last-stored values between program startup and program termination.

So in the shared library cases where a C main function runs, the Go runtime should conservatively assume the C program may legitimately mutate the argv strings.  In particular, it can't use gostringnocopy on them like it currently does in goargs.

mdempsky:
So I think in the cases where we need to play nicely with arbitrary C code:

  - syscall's {Clear,Get,Put,Set}env and Environ functions should all treat C as the source of truth for the environment, rather than keeping a local copy of the environment to manipulate and just trying to copy mutations to C.  (E.g., even today, if cgo code calls setenv, it won't be visible to os.Getenv; whereas os.Setenv is visible to getenv.)

  - package runtime should use getauxval for accessing the aux value array, to avoid needing to locate the array in memory.

That would just leave needing to figure out a solution for os.Args.  It would kinda suck, but maybe we could just leave it nil in the case where Go isn't acting as the program's entry point?

ianlancetaylor:
If we can't get os.Args, then we have to leave it nil.  But I think we should only do that for environments where we can't get it.  When we can get it, as for glibc, we should.

mdempsky:
I just looked at glibc, Bionic, uClibc, musl, and dietlibc, as well as the dynamic linkers for FreeBSD, NetBSD, and OpenBSD.  It looks like only glibc passes (argc, argv, envp) to the DSO init functions.

[...]

PR #52541, opened 2022-04-25
runtime: add check before using arguments with -buildmode=c-archive and -buildmode=c-shared on non glibc systems such as musl/uclinux

1800alex:
This PR fixes using go .so and .a libraries with non glibc systems such as musl/uclinux when linking at compile time. This fix modifies the solution found here https://go-review.googlesource.com/c/go/+/37868

This fix does not fix the issues encountered when using dlopen on musl systems, I'm opening this PR in hopes of getting some help from the go team to fix this crippling issue using musl based systems.

Previous PR here -> https://github.com/golang/go/pull/47249

Tested with go 1.18 on 386/armv7/aarch64 with both libc and musl


1800alex:
Linking with https://github.com/golang/go/issues/13492

PR #52581, opened 2022-04-26
syscall/js: enable a JS Promise pattern for Funcs

TiemenSch:
Methods using js.PromiseOf return a js.Func that is callable using async/await in JS.


TiemenSch:
I was working with Go and WASM today, and registering functions with 
```go
js.Global().Set("someAlias",  FuncOf(someFunc))
```
works, but without error handling the program quickly exits after one wrong call from the JS side to ``global.someAlias()``.

This PR proposes to adopt a pattern that is already in use in the community to resolve requests to the Go code with promises, and could be useable by: 
```go
js.Global().Set("someAlias", PromiseOf(somePromiseAbleFunc))
```
where the only difference should be is that ``somePromiseAbleFunc`` should return an interface and error, which it can then use to either resolve or reject the promise safely with ``await global.someAlias()``.

carlmjohnson:
I saw this in https://brunoluiz.net/blog/2022/aug/imagewand-privacy-first-image-conversion-experiment-with-golang-and-wasm/. I've written basically [the same helper](https://github.com/carlmjohnson/shitpic/blob/master/shitpicweb/wasm.go) in my WASM project, so this serves a very common need, but it's new API, so you need to go through the Go proposal process. Go to the Issue tab and create a new issue called "proposal: syscall/js: add PromiseOf" and start a discussion there. If approved, your PR can be merged afterwards.

PR #52595, opened 2022-04-28
cmd: replace `git.io` in the message

SukkaW:
Replace the `git.io` link with its actual URL.

x-ref: https://github.blog/changelog/2022-04-25-git-io-deprecation/


hitzhangjie:
This `git.io/JfYMW' is a short URL generated by `git.io`, the request will be redirected to the true URL, which points to https://github.com/google/pprof/blob/master/doc/README.md#interpreting-the-callgraph.

git.io will not accept new link, but it will still works for previous shortened links.

Issue #27741, opened 2018-09-18
compress/flate: Close is not idempotent


dsnet:
The `flate.Writer.Close` should be an idempotent operation. Furthermore, it should disallow further calls to `Write`:
```go
b := new(bytes.Buffer)
zw, _ := flate.NewWriter(b, 6)
fmt.Println(zw.Write([]byte("hello, world!")))
fmt.Println(zw.Close())
fmt.Println(zw.Write([]byte("hello, world!")))
fmt.Println(zw.Close())
fmt.Printf("%x\n", b.String())
```

Currently, this prints:
```
13 <nil>
<nil>
13 <nil>
<nil>
ca48cdc9c9d75128cf2fca495104040000ffff42e100020000ffff
```

I expect it to print something like:
```
13 <nil>
<nil>
0 flate: write on closed stream
<nil>
ca48cdc9c9d75128cf2fca495104040000ffff
```

Currently, allowing a `Write` after a `Close` results in a corrupted stream. The trailing `42e100020000ffff` is indicative of a backwards reference to the previous block. However, this is invalid since the previous block terminated the stream with the FINAL bit set.

This hasn't been much of a problem in practice since almost all usages of `flate` are through `gzip`, which does have this property.

bcmills:
Do you have a change pending, or should this be “help wanted”?

Liberatys:
I would like to have a go on this issue if it hasn't been picked up by someone already.

PR #52620, opened 2022-04-29
compress/flate: make Close idempotent

teivah:
Fixing https://github.com/golang/go/issues/27741.

This PR follows the same approach as in `gzip` using a `closed` flag. I also had to update two existing tests that were failing once `Close` became idempotent.


teivah:
Duplicate of https://go-review.googlesource.com/c/go/+/136475

PR #52643, opened 2022-05-01
strings: change in trim function documentation

D7682:
Documentation could cause confusion about whether the functions are returning a slice of bytes or a slice of separated string characters at a glance, changed to use similar phrasing as other trim functions.

the Gerrit change.


seankhliao:
The original wording is correct and the change is wrong

Issue #52563, opened 2022-04-26
cmd/compile: No BCE for len(x)&y in slice expression


greatroar:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.18 linux/amd64
$ gotip version
go version devel go1.19-96c8cc7fea9 Sun Apr 24 01:22:21 2022 +0000 linux/amd64
</pre>

### What did you do?

	const blocksize = 16
	fullBlocks := len(p) &^ (blocksize - 1)
	blk := p[:fullBlocks]
	rem := p[fullBlocks:]

### What did you expect to see?

I expected the bounds checks on the final two lines to be eliminated, since len(p)&x <= cap(p) always.

### What did you see instead?

Bounds checks. This is similar to #38476, but presumably easier to tackle. GOSSAFUNC output shows one of these checks as

```
v6 = SliceLen <int> v5
v8 = Const64 <int> [-16]
v9 = And64 <int> v6 v8 (fullBlocks[int])
v12 = SliceCap <int> v5
v14 = IsSliceInBounds <bool> v9 v12
```

greatroar:
#38717 is also similar, but for IsInBounds instead of IsSliceInBounds.

greatroar:
I just noticed that the examples from #38476, ChaCha20 and Poly1305, would reduce to this if the compiler also knew that

    x-(x&y) = x&^y

PR #52665, opened 2022-05-03
cmd/compile: ssa rewrite x-(x&y) => x&^y

Jorropo:
This allows to rewrite:
  AND BX, AX
  SUB AX, BX

Into (with GOAMD64=v3):
  ANDN AX, BX, AX

Which is twice as fast, and smaller.

Into (without GOAMD64=v3):
  NOT BX
  AND AX, BX

Which if AX's depency isn't resolved yet but BX is in most cases twice as
fast because now the NOT BX and AX's depency can execute in parallel.
In other cases, it has negligeable positive impact.

Other instructions set will also benefits from the pipelining adventage.
And also benefit from the smaller ANDN instruction if they have something
similar.

Help reduce ChaCha20 and Poly1305 to #52563

### DIFF
--- a/src/cmd/compile/internal/ssa/gen/AMD64.rules
+++ b/src/cmd/compile/internal/ssa/gen/AMD64.rules
@@ -2266,6 +2266,7 @@
 
 // CPUID feature: BMI1.
 (AND(Q|L) x (NOT(Q|L) y))           && buildcfg.GOAMD64 >= 3 => (ANDN(Q|L) x y)
+(SUB(Q|L) x (AND(Q|L) x y))         && buildcfg.GOAMD64 >= 3 => (ANDN(Q|L) x y)
 (AND(Q|L) x (NEG(Q|L) x))           && buildcfg.GOAMD64 >= 3 => (BLSI(Q|L) x)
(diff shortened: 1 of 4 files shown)


greatroar:
Are you sure the separate rule for amd64 is needed? It looks redundant given the generic rule.

Jorropo:
> Are you sure the separate rule for amd64 is needed? It looks redundant given the generic rule.

@greatroar 

Good point, this is a remanent from my first patch which was:
> ```
> (Sub(64|32|16|8) x z:(And(64|32|16|8) <t> x y)) && z.Uses == 1 => (And(64|32|16|8) x (Com(64|32|16|8) <t> y))
> ```
To avoid generating extra instructions where we wouldn't need them.
Turns out, this is not needed, will fix.

> NOT(L|Q) isn't generated for much else than Com(8|16|32|64).

I don't understand that, I do not emit or match any `NOT` outside of the `Com` in the generic.
The `(AND x (NOT y))` rule is one that already exists.

> Also, ChaPoly has constant y, so I expect that would be an `AND(L|Q)const` by the time that rule is checked.

That why the generic rule exists. :upside_down_face:

greatroar:
> I don't understand that, I do not emit or match any NOT outside of the Com in the generic.

Never mind, brain fart.

PR #52771, opened 2022-05-08
database/sql: add SetMaxBadConnRetries

jinzhu:
Add the ability to sets the number of maximum retries if the driver returns driver.ErrBadConn.

This can allow us to skip retry those long running SQL queries killed by db server.

Fixes #52886

### DIFF
--- a/src/database/sql/sql.go
+++ b/src/database/sql/sql.go
@@ -487,6 +487,11 @@ type DB struct {
 	maxIdleTimeClosed int64 // Total number of connections closed due to idle time.
 	maxLifetimeClosed int64 // Total number of connections closed due to max connection lifetime limit.
 
+	// maxBadConnRetries is the number of maximum retries if the driver returns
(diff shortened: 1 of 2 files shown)


src/database/sql/sql.go
```diff
@@ -1048,6 +1048,49 @@ func (db *DB) SetConnMaxIdleTime(d time.Duration) {
...
+	maxBadConnRetries := db.maxBadConnRetries
+	if maxBadConnRetries == 0 {
+		maxBadConnRetries = defaultMaxBadConnRetries
+	}
+
+	for i := 0; i < maxBadConnRetries; i++ {
+		err = fn(cachedOrNewConn)
+		isBadConn = err != nil && errors.Is(err, driver.ErrBadConn)
+		if !isBadConn {
+			break
```
tzzed, line 1080: I suggest to return `err` here and `nil` at the end of the function.

jinzhu: refactored the retry method, thank you for your review.

PR #52849, opened 2022-05-11
io/io.copy add custom  copy buffer size

ningmingxiao:
Signed-off-by: ningmingxiao <ning.mingxiao@zte.com.cn>

### DIFF
--- a/src/io/io.go
+++ b/src/io/io.go
@@ -382,7 +382,11 @@ func CopyN(dst Writer, src Reader, n int64) (written int64, err error) {
 // Otherwise, if dst implements the ReaderFrom interface,
 // the copy is implemented by calling dst.ReadFrom(src).
 func Copy(dst Writer, src Reader) (written int64, err error) {
-	return copyBuffer(dst, src, nil)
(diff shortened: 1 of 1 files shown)


Jorropo:
@ningmingxiao All changes the the std need to go [the proposal process](https://github.com/golang/proposal#readme) first.
You need to open a new issue proposing to add this.

Secondly, about the proposal, this already possible in the current API doing:
```go
io.CopyBuffer(t, f, make([]byte, size))
```
If you pass a buffer to `CopyBuffer` it's gonna make the copy in thoses chunks, replicating the same behaviour as your patch but without any new API.

Also `size` is a really confusing parameter name, I would expect it to be the total size of the copy, but no infact it's the chunk size.

ningmingxiao:
an exapmle , io.CopyBuffer(dst,src,make([]byte,16*1024)) ，16*1024  is custom cache size
is my pr redundant ?

mvdan:
Closing per @Jorropo's comment - API changes need to be discussed on the issue tracker first, so there isn't anything to do here.

Issue #43961, opened 2021-01-27
encoding/json: disable HTML escaping and omit final newline


89z:
I am encoding some JSON. I need to disable HTML escaping, and I need no trailing
newline. It seems no easy way to do this.

If I try `json#Marshal`, it does HTML escaping with no option to change. If I
try `json#Encoder.Encode`, it adds newline with no option to change. I came up
with this as a workaround:

~~~go
package main

import (
   "bytes"
   "encoding/json"
   "fmt"
   "log"
)

func compact(dst *bytes.Buffer, v interface{}) error {
   var src bytes.Buffer
   enc := json.NewEncoder(&src)
   enc.SetEscapeHTML(false)
   err := enc.Encode(v)
   if err != nil {
      return err
   }
   return json.Compact(dst, src.Bytes())
}

func main() {
   var dst bytes.Buffer
   e := compact(&dst, []string{"May", "June & July"})
   if e != nil {
      log.Fatal(e)
   }
   fmt.Printf("%q", dst.String())
}
~~~

but its pretty awkward. Also if you wanted to indent, it is difficult, since you
are now calling `json#Compact`, using `json#Encoder.SetIndent` would be pointless, you
would have to call `json#Indent` after everything is done.

https://golang.org/pkg/encoding/json

seankhliao:
This is possible with the current api (as you've demonstrated), see
https://golang.org/doc/faq#x_in_std

89z:
@seankhliao my workaround is really inefficient. The `Compact` function is overkill to essentially remove a trailing newline, when it would be better if option to just not add the newline in the first place.

89z:
@seankhliao my workaround is really inefficient. The `Compact` function is overkill to essentially remove a trailing newline, when it would be better if option to just not add the newline in the first place.

Also that link doesnt really relate to this issue. Im not asking to add a new package.

seankhliao:
You can just slice off the last element of the byte slice.

89z:
@seankhliao yeah, but again, it would be better if it wasnt added in the first place, even as an option.

89z:
Here is another approach, based on your suggestion:

~~~go
func marshal(v interface{}) ([]byte, error) {
   var dst bytes.Buffer
   enc := json.NewEncoder(&dst)
   enc.SetEscapeHTML(false)
   err := enc.Encode(v)
   if err != nil {
      return nil, err
   }
   return dst.Bytes()[:dst.Len() - 1], nil
}
~~~

[...]

PR #52903, opened 2022-05-13
encoding/json: add terminate newline as option

sshaplygin:
I added terminate newline as option because exists case by me encode input data to single line.


seankhliao:
API changes should go through the [proposal process](https://go.dev/s/proposal-process)

Closing as this particular change has been declined before, ref https://github.com/golang/go/issues/43961#issuecomment-769722443

PR #53073, opened 2022-05-25
debug/buildinfo: fix buildinfo data section location on elf and pe

### LINKED ISSUE #53071: affected/package: BuildInfo
1.18

Yes

PE go binary, any host OS

### What did you do?
The go buildinfo tool fails to locate the build info because the bit test here is wrong https://github.com/golang/go/blob/119da6358bf8cc01bb8bc7556d616bd0fe723ad4/src/debug/buildinfo/buildinfo.go#L317. When the `.data` section characteristics flags are value `0xC0500040`. 

image

The `.data` section to be found.

### What did you see instead?
The `.data` section is not found and the PE is seen as malformed by the buildinfo command, and information extraction fails.

stevemk14ebr:
Fixes the section flag checks in buildinfo. The elf flag check was wrong too but it's never hit due to the symbol resolution just above, which is usually valid for non-stripped binaries.

Fixes https://github.com/golang/go/issues/53071


stevemk14ebr:
This is a mess now, I'm just going to re-open this with the correct commit author.

PR #53086, opened 2022-05-25
io: use errors.Is in reader and writer helpers

dominicbarnes:
This adds broad usage of errors.Is within io helper functions to make these helpers more resilient to implementations that choose to wrap errors. Currently, if an io.Reader wraps an io.EOF error in any way, helpers like io.ReadAll will not behave as expected, since they seem to expecting a naked EOF error to be returned.


seankhliao:
Closing as declined

Issue #36819, opened 2020-01-28
net/http: graceful shutdown race condition


zerkms:
**Update**: I have found a super trivial way to reproduce it, see the step at the very bottom

Preamble: I think it's relevant to https://github.com/golang/go/issues/23829

### What version of Go are you using (`go version`)?

<pre>
$ go _version_
go version go1.13.6 linux/amd64
</pre>

### Does this issue reproduce with the latest release?
Yes

### What operating system and processor architecture are you using (`go env`)?
GOHOSTARCH="amd64"
GOHOSTOS="linux"

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/ivan/.cache/go-build"
GOENV="/home/ivan/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/srv/work/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/home/ivan/.local/share/umake/go/go-lang"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/home/ivan/.local/share/umake/go/go-lang/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build903857222=/tmp/go-build -gno-record-gcc-switches"

</pre></details>

### What did you do?

```go
package main

import (
	"context"
	"fmt"
	"net/http"
	"os"
	"os/signal"
	"syscall"
)

func main() {
	sig := make(chan os.Signal, 1)
	signal.Notify(sig, syscall.SIGTERM, os.Interrupt)

	mux := http.NewServeMux()
	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {})

	srv := &http.Server{
[...]

Issue #33313, opened 2019-07-27
net/http: Shutdown may lose an active connection


monkey92t:
### What version of Go are you using (`go version`)?

<pre>
go version go1.12.4 darwin/amd64

</pre>

### Does this issue reproduce with the latest release?
yes go 1.12

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
GOARCH="amd64"
GOBIN="/usr/local/go/bin"
GOCACHE="/var/root/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/var/root/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/ax/Desktop/ArcTestHttpServer/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build931930631=/tmp/go-build -gno-record-gcc-switches -fno-common"

</pre></details>

my english is not very good. I try to describe it.

i found a problem in http shutdown(), when the shutdown function is executed, srv.mu.Lock() is executed. but at this time the net accept function can still be executed normally, execute c.setState(c.rwc, StateNew) put a new connection into activeConn and wait for srv.mu.Unlock(). 

when shutdown() finishes processing all activeConn connections, it will execute srv.mu.Unlock() and end the shutdown function, c.setState() gets the lock and puts a new connection into activeConn. At this point, there will still be an active connection after the shutdown() is completed.

medusar:
I have encountered the same problem.  
When Shutdown method is called, it will first close the Listener and then close all the active connections, but the code in method Serve will give a chance for a connection returned by Accept but has not been added to active connection array when shutdown is closing all the active connections, then the newly established connection will not be closed.

[...]

PR #53118, opened 2022-05-27
net/http: close accepted connection when Closed

### LINKED ISSUE #48642: net/http: Server.Close may not able to close all connections
I encountered lingering connection after calling Server.Close and it's not closed by Server.Close

Here is reproducible example: https://play.golang.org/p/9aLVDwWPQhH

The problem is http/server.go: Server.Serve function didn't check if Server is already closed before calling handler, and if Server.Close been called after l.Accept return but before c.setState(c.rwc, StateNew, runHooks), the new connection is not registered in the activeConn map when Server.Close try to close all connections.

The example code simulate this condition by sleep one second during ConnContext call. Program can exit corrected if this sleep is removed.

AlexanderYastrebov:
Previous attempt https://golang.org/cl/353714 also incorrectly closed
accepted connection while in shutdown.

Fixes #48642
Updates #33313, #36819

### DIFF
--- a/src/net/http/serve_test.go
+++ b/src/net/http/serve_test.go
@@ -6757,3 +6757,124 @@ func TestProcessing(t *testing.T) {
 		t.Errorf("unexpected response; got %q; should start by %q", got, expected)
 	}
 }
+
+type wrapConnListener struct {
+	net.Listener
+	wrap func(net.Conn) net.Conn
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
Superseded by https://go.dev/cl/409537

PR #53242, opened 2022-06-05
os: don't check for IsExist in MkdirAll example

waldreiter:
If a directory already exists, then MkdirAll returns nil. Therefore the
check with IsExist is not necessary.

### DIFF
--- a/src/os/example_test.go
+++ b/src/os/example_test.go
@@ -255,7 +255,7 @@ func ExampleMkdir() {
 
 func ExampleMkdirAll() {
 	err := os.MkdirAll("test/subdir", 0750)
-	if err != nil && !os.IsExist(err) {
+	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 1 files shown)


AlexanderProschek:
I think you are miss-understanding what this check actually does. The if statement essentially checks if there is an error, and that error occurred because something occurred that was not because the folder already existed. Mkdir returns an error and everything might still be ok (the folder might have just existed before).

Think about it like this:

| Error | Type  | Action |
| --- | --- | --- |
| nil | | skip condition |
| non-nil | folder already exists | skip condition |
| non-nil | something else | execute condition + fatal (something weird happened) |

waldreiter:
@AlexanderProschek 
What you are saying is correct for `Mkdir`, but `MkdirAll` is different. The documentation says: _"If path is already a directory, MkdirAll does nothing and returns nil."_ Therefore `IsExist` gives us no information. 

The example and the documentation contradict each other. I appreciative the discussion, because part of the motivation for this pull request is to find out if my understanding is correct.

PR #53441, opened 2022-06-18
bytes/buffer: add Buffer.Init([]byte)

cwbhhjl:
currently bytes.NewBuffer can be used to initialize Buffer with a custom buf, but in performance-sensitive scenarios, I hope to reuse Buffer to avoid heap allocation and memcpy, so I added the Init method so that I can replace a Buffer's buf.

### DIFF
--- a/src/bytes/buffer.go
+++ b/src/bytes/buffer.go
@@ -452,6 +452,15 @@ func (b *Buffer) ReadString(delim byte) (line string, err error) {
 	return string(slice), err
 }
 
+// Init initializes the buffer using buf as its initial contents.
+// the buffer takes ownership of buf, and the caller should not use
+// buf after this call.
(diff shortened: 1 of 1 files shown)


Jorropo:
@cwbhhjl fyi, you need to create a proposal issue and it to be accepted before sending a PR.
This wont be accepted.

You should read: https://github.com/golang/proposal#the-proposal-process

Jorropo:
@cwbhhjl fyi once you closed a PR, the bot will close the matching one on gerrit, but afaik gerrit doesn't support reopening.

You need to make a new PR.

cwbhhjl:
@Jorropo 

Thanks, I created an issue, if the proposal can be accepted, I will create a new pr.

* https://github.com/golang/go/issues/53443

PR #53503, opened 2022-06-22
crypto/aes: speedup CTR mode on AMD64 and ARM64

### LINKED ISSUE #20967: crypto/aes: implement ctrAble when using AES-NI on amd64

1.8.3

linux/amd64

### What did you do?

Benchmarked CTR-AES128 backed by `crypto/aes` on a system with AES-NI.

Acceptable performance.

### What did you see instead?

~400 MB/s for 1 KiB writes.  For reference the same system using `crypto/aes`'s `gcmAble` does GCM-AES128 Seal() at ~1300 MB/s.  A cursory look at the source suggests that there is no special case `ctrAble` implementation for AES-NI.

starius:
The implementation runs up to 8 AES instructions in different registers
one after another in ASM code. Because CPU has instruction pipelining
and the instructions do not depend on each other, they can run in
parallel with this layout of code. This results in significant speedup
compared to the regular implementation in which blocks are processed in
the same registers so AES instructions do not run in parallel.

GCM mode already utilizes the approach.

The ASM implementation of ctrAble has most of its code in XORKeyStreamAt
method which has an additional argument, offset. It allows to use it
in a stateless way and to jump to any location in the stream. The method
does not exist in pure Go and boringcrypto implementations.

AES CTR benchmark delta.

$ go test crypto/cipher -bench 'BenchmarkAESCTR*'

AMD64. Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
name                   old time/op    new time/op   delta
BenchmarkAESCTR1K-2       1259ns        266.9ns     -78.8%
BenchmarkAESCTR8K-2       9859ns         1953ns     -80.1%

ARM64. ARM Neoverse-N1 (AWS EC2 t4g.small instance)
name                   old time/op    new time/op   delta
BenchmarkAESCTR1K-2       1098ns        481.1ns     -56.2%
BenchmarkAESCTR8K-2       8447ns         3452ns     -59.1%

ARM64. Apple M1
name                   old time/op    new time/op   delta
BenchmarkAESCTR1K-2      455.3ns        154.3ns     -66.1%
BenchmarkAESCTR8K-2       3491ns         1116ns     -68.0%

Fixes #20967

Investigation and initial implementation:
https://github.com/mmcloughlin/aesnix/
Full implementation in external repo:
https://github.com/starius/aesctrat


starius:
@andig I'm rewriting the generators in Go. Want to make a single Go program producing for all architectures.

starius:
@andig I finished with generators and added correctness test. Waiting for review in Gerrit :-)

PR #53506, opened 2022-06-22
cmd/test2json: parse entire output lines for pass/fail/other update messages

### LINKED ISSUE #26325: testing: stdout output with no newline produces confusing/corrupt -v output
```go
package out_test

import (
	"fmt"
	"testing"
)

func TestOut(t *testing.T) { fmt.Print("TestOut") }
```

```sh
$ go test .
ok  	testout	0.016s
$ go test . -v
=== RUN   TestOut
Test--- PASS: TestOut (0.00s)
PASS
ok  	testout	0.017s
```

Output to stdout during tests is captured and printed along with `t.Logf` et al. output when `-v` is supplied. If the output doesn't end with a newline, the output is as above.

We have some tools that parse test output which get confused in this situation. It's easy enough to avoid (just print a newline in your test) but it would also be pretty trivial for the testing package could add a newline when one is missing.

MisterSquishy:
Fixes #26325

### DIFF
--- a/src/cmd/internal/test2json/test2json.go
+++ b/src/cmd/internal/test2json/test2json.go
@@ -191,13 +191,16 @@ func (c *Converter) handleInputLine(line []byte) {
 	// "=== RUN   "
 	// "=== PAUSE "
 	// "=== CONT  "
-	actionColon := false
+	start := -1
+	end := -1
 	origLine := line
 	ok := false
 	indent := 0
 	for _, magic := range updates {
-		if bytes.HasPrefix(line, magic) {
+		start = bytes.Index(line, magic)
+		if start > -1 {
 			ok = true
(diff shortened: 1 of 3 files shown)


MisterSquishy:
@ianlancetaylor I'm not completely sure I understand what Gerrit is trying to tell me... are you reviewing this change?

ianlancetaylor:
@MisterSquishy I asked a question about the change, and you answered it, but I am not reviewing it.  We are currently in a release freeze for the 1.19 release.  This change can be reviewed for 1.20.  But it's a hard change to review because the effects are difficult to analyze.  We have two unlikely scenarios; which is less likely?

That said it's much better for our process if discussion can take place in Gerrit, not here.  Thanks.

Issue #53664, opened 2022-07-03
go/types/testdata/fixedbugs: git rejects file with carriage returns


willfaught:
### What version of Go are you using (`go version`)?

1.18.3

### Does this issue reproduce with the latest release?

Yes

### What operating system and processor architecture are you using (`go env`)?

N/A

### What did you do?

```
~/Developer/proj/internal/go/src main
❯ git add ./go/types/testdata/fixedbugs # copied from ~/Developer/go/src, the go repo (version go1.18.3)
fatal: CRLF would be replaced by LF in internal/go/src/go/types/testdata/fixedbugs/issue39754.go2

~/Developer/go/src tags/go1.18.3 24s
❯ xxd ./go/types/testdata/fixedbugs/issue39754.go2
00000000: 2f2f 2043 6f70 7972 6967 6874 2032 3032  // Copyright 202
00000010: 3020 5468 6520 476f 2041 7574 686f 7273  0 The Go Authors
00000020: 2e20 416c 6c20 7269 6768 7473 2072 6573  . All rights res
00000030: 6572 7665 642e 0d0a 2f2f 2055 7365 206f  erved...// Use o
```

You can see the CR before the LF in the hex view.

### What did you expect to see?

`git add` succeed.

### What did you see instead?

`git add` fail.

ianlancetaylor:
I don't know enough about git to understand what is going on, but crlf is valid in a Go file.  I don't know why this test case is using crlf, but we shouldn't have to change the file to work around git.  Clearly git is able to handle crlf files, as evidenced by the fact that the file exists in Go's git repository.  So it seems to me that this must have something to do with some local git configuration on your system.

willfaught:
Wow, fast response! :)

You're right, I just realized I have Git config that has CRLF policy:

```
[core]
autocrlf = input
safecrlf = true
```

Setting either of those to false makes it work.

The issue isn't about removing CRs from valid Go. I'd say it's about making the implementation compatible with typical Git config people use to keep line endings consistent across platforms. For instance, without these policies, if I had added a new line to this file on macOS, then committed it, the line endings would be inconsistent within the file, and it would probably appear malformed in Windows Notepad. If Go intends to support Windows, and accept contributions from Windows users, then it seems to me like a good idea to use one line ending everywhere.

alexbrainman:
[...]

PR #53665, opened 2022-07-03
testdata: normalize line ending for go files

ZekeLu:
According to .gitattributes, all go files in this repository should
use LF as the line ending.

I have checked the commit history, it turns out that there is not
intention to use CRLF for these files.

Updates #53664.


ZekeLu:
Abandoned due to https://github.com/golang/go/issues/53664#issuecomment-1173167979.

PR #53677, opened 2022-07-04
Update path.go

Tomator01:
Solve the problem of missing files while using package filepath.Walk .

the Gerrit change.

  + No trailing period
+ No Markdown

### DIFF
--- a/src/path/filepath/path.go
+++ b/src/path/filepath/path.go
@@ -456,7 +456,7 @@ func walk(path string, info fs.FileInfo, walkFn WalkFunc) error {
 		} else {
 			err = walk(filename, fileInfo, walkFn)
 			if err != nil {
-				if !fileInfo.IsDir() || err != SkipDir {
+				if err != SkipDir {
 					return err
 				}
 			}


Tomator01:
image

In this path ,i use filepath.Walk to get all directory or file under "/home/pengpai/p2" ,but when  the variable skipobj is file not directory ,the return will lost some files.
 You can repet this issue by the following demo.

```
package main

import (
	"fmt"
	"io/fs"
	"os"
	//"github.com/path/filepath"
	"path/filepath"
)

func main() {
	dir := "/home/pengpai/p2"
	skipobj := "/home/pengpai/p2/test2.c"
	//skipobj := "/home/pengpai/p2/p1"
	var objlist []string
	filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
		if path == skipobj {
			return fs.SkipDir
		} else {
			objlist = append(objlist, path)
			return nil
		}

	})
	for _, obj := range objlist {
		fmt.Println(obj)
	}
}

```

When skipobj := "/home/pengpai/p2/test2.c" , the print :
image
Has lost  the test3.c and test.c .

When /skipobj := "/home/pengpai/p2/p1" ,the print:
image
is right.

This issue could lead to some file lost.

ZekeLu:
I think you should change your walkFunc like this:

```diff
  		if path == skipobj {
- 			return fs.SkipDir
+ 			return nil
  		} else {
```

Tomator01:
> I think you should change your walkFunc like this:
> 
> ```diff
>   		if path == skipobj {
> - 			return fs.SkipDir
> + 			if info.IsDir() {
> + 				return fs.SkipDir
> + 			}
> + 			return nil
>   		} else {
> ```

yes ,You are right ,but i think we can implement this in path package ,or why not？

ZekeLu:
The behavior is documented: https://pkg.go.dev/path/filepath#WalkFunc

> The error result returned by the function controls how Walk continues. If the function returns the special value SkipDir, Walk skips the current directory (path if info.IsDir() is true, **otherwise path's parent directory**). Otherwise, if the function returns a non-nil error, Walk stops entirely and returns that error.

PR #53901, opened 2022-07-15
go/time : Added json tags in struct field

Pradumn-18:
This change modifies Go's time package to be used as a type in CRD generations.

Since the json tag were missing for struct Time {} in go/src/time/time.go . It was not allowing time.Time to be used as a variable in CRD genetration in k8s.


mvdan:
Sorry but this doesn't make sense. These fields are not exported, so they are not available via reflection for libraries like encoding/json.

An API change like this should be discussed in the issue tracker first, as well - we shouldn't make changes like this lightly.

PR #53978, opened 2022-07-21
net/textproto: add extra field to indicate whether EOF is reached

### LINKED ISSUE #53858: net/textproto: swallows errors returned by underlying reader

<pre>
$ go version
go version go1.18.3 linux/amd64

</pre>

Yes.

### What did you do?

See example: https://go.dev/play/p/Bn8H6p4QUf6

I expect that bufio package react on underlying reader errors and stop the reading.

### What did you see instead?

I see that bufio skips an error multiple times, and the example code causes OOM error becuase reader is rewinded every time:

here are cases when an error is not tracked:
https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/net/textproto/reader.go;l=148;drc=27794c4d4a18c61d8c158d253421d72b5a6a8673
https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/net/textproto/reader.go;l=496;drc=27794c4d4a18c61d8c158d253421d72b5a6a8673
https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/net/textproto/reader.go;l=566;drc=27794c4d4a18c61d8c158d253421d72b5a6a8673
https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/net/textproto/reader.go;l=571;drc=27794c4d4a18c61d8c158d253421d72b5a6a8673

the error, returned by autoRewind is just ignored and set to nil:
https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/bufio/bufio.go;l=156;drc=27794c4d4a18c61d8c158d253421d72b5a6a8673
https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/bufio/bufio.go;l=124;drc=27794c4d4a18c61d8c158d253421d72b5a6a8673

I know that it's something that I have to expect with autoRewind, but this behavior caused an unpleasant surprise.

see also:
* https://github.com/golang/go/issues/53854
* https://github.com/golang/go/issues/53855

cz2h:
The existing implementation of ReadMIMEHeader() relies on the assumption that bufio will keep the EOF error. In the case where a rewind reader is use, bufio will always return a string and ignore the EOF. So a variable is used to indicate that EOF is reached.

Corresponding test is provided.

Fix #53858

### DIFF
--- a/src/net/textproto/reader.go
+++ b/src/net/textproto/reader.go
@@ -17,10 +17,9 @@ import (
 // A Reader implements convenience methods for reading requests
 // or responses from a text protocol network connection.
 type Reader struct {
-	R        *bufio.Reader
-	dot      *dotReader
-	buf      []byte // a re-usable buffer for readContinuedLineSlice
-	reachEnd bool
+	R   *bufio.Reader
+	dot *dotReader
+	buf []byte // a re-usable buffer for readContinuedLineSlice
 }
 
 // NewReader returns a new Reader reading from r.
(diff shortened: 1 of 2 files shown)


zengchu2:
Commit message does not align with latest fixes. Will open another PR for this

PR #54281, opened 2022-08-05
net/textproto: dont't ignore error in skipSpace function

cz2h:
skipSpace ignores the error it encounters while using a reader 
which seems like a bug. Refactor the function and have it 
returns the errors. 

Modified readContinuedLineSlice and have it ignore io.EOF 
while something is ever read in the function to keep its 
behaviour unchanged.

Update  #53858

### DIFF
--- a/src/net/textproto/reader.go
+++ b/src/net/textproto/reader.go
@@ -157,7 +157,15 @@ func (r *Reader) readContinuedLineSlice(validateFirstLine func([]byte) error) ([
 	r.buf = append(r.buf[:0], trim(line)...)
 
 	// Read continuation lines.
-	for r.skipSpace() > 0 {
+	for {
+		n, err := r.skipSpace()
+		if n == 0 || err != nil {
+			// If function reads something, it ignores io.EOF.
+			if err == io.EOF && len(r.buf) > 0 {
+				err = nil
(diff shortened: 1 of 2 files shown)


kayrus:
@zengchu2 why this PR was closed?

heavenisme:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

zengchu2:
> @zengchu2 why this PR was closed?

It seems that changes to skipSpace will affect the behaviour of readContinuedLine, making it return a EOF after reading the last line of text. On the other hand, if we somehow ignore this EOF if we are returning something, then this change seems not that necessary cause it cannot fix corner case mentioned in the issue.

Probably there is another way of fixing that issue?

kayrus:
> Probably there is another way of fixing that issue?

not sure, but silently swallowing the err is also not good :\

zengchu2:
> 

Agree with that. But I also found it little bit hard to provide a proper example test case for this refactoring. 

For regular io readers, the error can be detected during the next time when readContinuedLine is called. I guess this is why the error is ignored in skipSpace

PR #54290, opened 2022-08-05
crypto/x509: err instead of panic on invalid curve

### LINKED ISSUE #54288: crypto/x509: panics on invalid curve instead of returning error

<pre>
$ go version 1.19

</pre>

### What did you do?
In go 1.19 `crypto/elliptic` will now panic when operating on invalid curve points. This is great as it avoids undefined behavior, but `crypto/x509` will now panic in `marshalPublicKey` in the `case *ecdsa.PublicKey:` where it previosly would return `errors.New("x509: unsupported elliptic curve")` as it calls `elliptic.Marshal(pub.Curve, pub.X, pub.Y)` before it checks `oidFromNamedCurve(pub.Curve)`.

So simply switching those two calls around should fix the issue.

Meanwhile clients have to manually check if their key satisfies `IsOnCurve` before calling something like `x509.MarshalPKIXPublicKey` where in previous versions of go you could rely on an error being returned instead of a panic.

FnuGk:
in marshalPublicKey case *ecdsa.PublicKey:
because elliptic.Marshal() was called before oidFromNamedCurve we would panic
if the curve was invalid.
By calling oidFromNamedCurve first we can return early with a meaningfull error

Fixes #54288

### DIFF
--- a/src/crypto/x509/x509.go
+++ b/src/crypto/x509/x509.go
@@ -84,11 +84,11 @@ func marshalPublicKey(pub any) (publicKeyBytes []byte, publicKeyAlgorithm pkix.A
 		// RFC 3279, Section 2.3.1.
 		publicKeyAlgorithm.Parameters = asn1.NullRawValue
 	case *ecdsa.PublicKey:
-		publicKeyBytes = elliptic.Marshal(pub.Curve, pub.X, pub.Y)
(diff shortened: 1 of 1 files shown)


seankhliao:
superceded by https://go.dev/cl/422298

PR #54387, opened 2022-08-11
doc/go_mem.html: delete a sentence that is ambiguous

shichuzhu:
It is unclear what the statement
"then any read r on x has only one possible W(r)" means:

A. If the context here is about: In a given program execution
(aka given a specific mapping W),
then this statement is trivial to mention,
as W(r) is by definition a write operation w,
therefore the only one.

B. If the context here is about: Among all possible
program execution (aka all possible mappings {W})
that satisfies requirement 1,2,3,
then this statement is not true.
Because W(r) can be different write operations
in different program execution
(data-race-free doesn't imply deterministic execution).

Alternatively, a less ambiguous statement could be like:

"if there are no read-write or write-write data races on memory location x,
then any read r on x has only one possible W(r), among all program
executions {W} that share the same synchronized before relation."

### DIFF
--- a/doc/go_mem.html
+++ b/doc/go_mem.html
@@ -183,13 +183,7 @@ which are unordered by happens before.
 </p>
 
 <p>
-Note that if there are no read-write or write-write data races on memory location <i>x</i>,
-then any read <i>r</i> on <i>x</i> has only one possible <i>W</i>(<i>r</i>):
-the single <i>w</i> that immediately precedes it in the happens before order.
-</p>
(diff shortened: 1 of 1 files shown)


shichuzhu:
Alternatively, a less ambiguous statement could be
'''
if there are no read-write or write-write data races on memory location x, then any read r on x has only one possible W(r), among all _program executions_ {W} that share the same _synchronized before_ relation.
'''

PR #54417, opened 2022-08-12
expvar: use atomic types+methods instead of atomic functions to guarantee 8-byte alignments for Int and Float values

zigo101:
If the Int and Float types are embedded in other struct types,
the alignments of the Int and Float fields might be not always
8-byte on 32-bit architectures.

This PR tries to avoid this problem.


go101:
Close for done in other PRs.

PR #54721, opened 2022-08-28
context: make use of a rw-mutex to protect cancelCtx fields

ortuman:
Reduce mutex contention on `cancelCtx` when cancellation error is queried from multiple goroutines.

Signed-off-by: Miguel Ángel Ortuño <ortuman@gmail.com>


seankhliao:
This is strictly worse than a plain mutex

Issue #49080, opened 2021-10-20
Using cgo built DLL on Visual Studio Project


rikkei-minh:
### What version of Go are you using (`go version`)?
go version go1.15.6 

### Does this issue reproduce with the latest release?
I have not checked on go1.17 yet.

### What operating system and processor architecture are you using (`go env`)?
windows/amd64

### What did you do?
I want to use a golang module as c-shared library and then call it from a MFC application.

1. I built the go module as c-shared library(example.dll) and example.h file successfully.
2. I created simple C program, included example.h file and called my library function, I built the program with gcc command, everything worked as expected.
3. On MFC project, included example.h file and called my library function, then used VS2019 Community to build. Error came out.

```
error E0020: identifier "__SIZE_TYPE__" is undefined
error E0065: expected ';'
error C2143: syntax error : missing '}' before 'enum [tag]'
```
I checked on Internet and found out that the problem is about compiler.
On VS2019, they use msvc as compiler for C program so when I include a cgo built library (which is using gcc by default) on VS2019 project, there are some conflicts between them.

I tried to find a way to build my golang module to c-shared library with msvc compiler and found this.
https://go-review.googlesource.com/c/go/+/133946

Unfortunately, this [cmd/compile: add support for MSVC toolchain to go build] change is still in review.

### What did you expect to see?
So if there is an official guide about using cgo built library on VS2019, I would be happy.

### What did you see instead?
Included in [What di you do?] section.

Issue #42146, opened 2020-10-22
build: cgo with clang and -buildmode=c-shared failures on Windows


tungnh28:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.15.3 windows/amd64
</pre>

### Does this issue reproduce with the latest release?

Yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
set GO111MODULE=
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\nguyenha\AppData\Local\go-build
set GOENV=C:\Users\nguyenha\AppData\Roaming\go\env
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOINSECURE=
set GOMODCACHE=C:\Users\nguyenha\go\pkg\mod
set GONOPROXY=
set GONOSUMDB=
set GOOS=windows
set GOPATH=C:\Users\nguyenha\go
set GOPRIVATE=
set GOPROXY=https://proxy.golang.org,direct
set GOROOT=c:\go
set GOSUMDB=sum.golang.org
set GOTMPDIR=
set GOTOOLDIR=c:\go\pkg\tool\windows_amd64
set GCCGO=gccgo
set AR=ar
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=D:\work\projects\mimic\mimic-core\go.mod
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -mthreads -fmessage-length=0 -fdebug-prefix-map=C:\Users\nguyenha\AppData\Local\Temp\go-build611343782=/tmp/go-build -gno-record-gcc-switches
</pre></details>

### What did you do?

set CGO_ENABLED=1
set CC=clang
set PATH=C:\Program Files\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin;%PATH%
go build -buildmode=c-shared -o bin/crawler.dll -ldflags "-v" -v my/pathp/export

### What did you expect to see?
Build success.

### What did you see instead?
HEADER = -H10 -T0xffffffffffffffff -R0xffffffff
[...]

PR #54811, opened 2022-09-01
cmd/link: support LLVM/Clang for MSVC

kkqy:
I find that cgo can compatible with LLVM/Clang for MSVC well actually.
The only problem I have found is that some arguments for linker of Mingw/GCC can't be used in linker of MSVC, so I write a judgement.
If go detect that the toolchain is LLVM/Clang for MSVC, do not pass those arguments that are not supported by linker of MSVC.

### DIFF
--- a/src/cmd/link/internal/ld/lib.go
+++ b/src/cmd/link/internal/ld/lib.go
@@ -1295,6 +1295,18 @@ func (ctxt *Link) archive() {
 }
 
 func (ctxt *Link) hostlink() {
+	// Determine if we are using Clang for MSVC.
+	// Some args for the linker of Mingw can not be passed to the linker of MSVC,
+	// and some args for the linker of MSVC are defferent from that of Mingw.
+	extld := ctxt.extld()
+	name := extld[0]
+	usingClangForMSVC := false
+	if out, err := exec.Command(name, "--version").CombinedOutput(); err == nil {
+		if bytes.Contains(out, []byte("msvc")) {
+			usingClangForMSVC = true
+		}
+	}
+
 	if ctxt.LinkMode != LinkExternal || nerrors > 0 {
 		return
 	}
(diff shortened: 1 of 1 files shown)


KuoKongQingYun:
Updates https://github.com/golang/go/issues/42146
Updates https://github.com/golang/go/issues/49080

KuoKongQingYun:
I find the second problem today that is the program using Clang for MSVC can't be debugged.
I have no idea to fix it because it seems like Golang supports debug info of GCC only.
so this commit only help sb to link those libraries of built-by-MSVC and closed source.

Mingw/GCC has defferent C++ ABI from MSVC,so we can't use C++ library of MSVC in Mingw/GCC,
but there are so many commercial libraries are closed source and built by MSVC.
It is regretful if Golang have no easy solution to be compatible with them.

We can build at least now....hm....

seankhliao:
please use Gerrit for discussing the change, we don't use GitHub pull requests for this

gatspy:
how use msvc complie cgo?

ianlancetaylor:
@gatspy Please see https://go.dev/wiki/Questions.

kkqy:
@gatspy you need:
1. Edit the source like this commit, and remove those flags that are available only in mingw.
2. Then install the Clang for MSVC (which use MSVC's compiler as its backend), and add the Clang into the PATH.
3. set environment：
```
CC="clang"
CXX="clang++"
```

now, your code in CGO can be compiled by compiler of MSVC.

vault-thirteen:
If Google does not want to lose this war, they will have to support LLVM sooner or later.

asaka1234:
> kkqy

Hi kkqy, Now golang has been update to version  1.23.4,   if we want to use msvc for cgo,  do we still need to modify the source code ?

kkqy:
> > kkqy
> 
> Hi kkqy, Now golang has been update to version 1.23.4, if we want to use msvc for cgo, do we still need to modify the source code ?

Yes, because msvc's clang has many different parameters, but I have given up this pull request, because I feel that compatibility with MSVC is something that will not be maintained in the long term. It is better to use https://github.com/llvm/llvm-project directly.

asaka1234:
> > > kkqy
> > 
> > 
> > Hi kkqy, Now golang has been update to version 1.23.4, if we want to use msvc for cgo, do we still need to modify the source code ?
> 
> Yes, because msvc's clang has many different parameters, but I have given up this pull request, because I feel that compatibility with MSVC is something that will not be maintained in the long term. It is better to use https://github.com/llvm/llvm-project directly.

Can you give me you email? i want to discuss with you deeply. pls

PR #54945, opened 2022-09-08
misc/cgo/gmp: fix a method and panic

sjnam:
misc/cgo/gmp, 
- the result of Mod() is not like the Go's.
- Exp() if m == nil, panic occurs.


sjnam:
- misc/cgo/gmp, the result of `Mod()` is not like the Go's.
- `Exp()` if m == nil, panic occurs.

PR #55119, opened 2022-09-17
image/png: optimise RGBA encoding

alx-ef:
Optimised RGBA image encoding to PNG. Improved test coverage. Reworked benchmark.

Performance improvement with both old and new versions of the benchmark:

name                           old speed      new speed      delta
EncodeRGBA_OriginalVersion-10   115MB/s ± 1%   308MB/s ± 1%  +166.70%  (p=0.000 n=19+17)
EncodeRGBA_NewVersion-10       40.3MB/s ± 1%  51.1MB/s ± 2%   +26.93%  (p=0.000 n=18+20)

name                           old allocs/op  new allocs/op  delta
EncodeRGBA_OriginalVersion-10      614k ± 0%        0k ± 0%   -99.99%  (p=0.000 n=20+20)
EncodeRGBA_NewVersion-10           614k ± 0%        0k ± 0%   -99.99%  (p=0.000 n=20+20)

### DIFF
--- a/src/image/png/writer.go
+++ b/src/image/png/writer.go
@@ -450,22 +450,27 @@ func (e *encoder) writeImage(w io.Writer, m image.Image, cb int, level int) erro
 			if nrgba != nil {
 				offset := (y - b.Min.Y) * nrgba.Stride
 				copy(cr[0][1:], nrgba.Pix[offset:offset+b.Dx()*4])
+			} else if rgba != nil {
+				for x := b.Min.X; x < b.Max.X; x++ {
+					offset := rgba.PixOffset(x, y)
+					a := rgba.Pix[offset+3]
+					if a == 0 {
+						// Do nothing, keep next 4 bites as 0, it's a transparent pixel
+					} else if a == 0xff {
+						// opaque pixel, no need to un-premultiply
+						copy(cr[0][i:i+4], rgba.Pix[rgba.PixOffset(x, y):rgba.PixOffset(x, y)+4])
+					} else {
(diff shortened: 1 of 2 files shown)


alx-ef:
## Cove coverage
### Before
```
ok  	std/image/png	0.702s	coverage: 85.3% of statements
```
### After 
```
ok  	std/image/png	0.791s	coverage: 86.2% of statements
```

## Benchmark results
### Before
```
/usr/local/go/bin/go test -v ./... -bench . -run ^$
goos: darwin
goarch: arm64
pkg: std/image/png
 . . .
BenchmarkEncodeRGBA-10           100	  10812776 ns/op	 113.64 MB/s	 3321754 B/op	  614432 allocs/op
PASS
ok  	std/image/png	19.135s
```

### After optimizations
Almost 3x faster
```
BenchmarkEncodeRGBA-10           301	   3980006 ns/op	 308.74 MB/s	  864128 B/op	      32 allocs/op
```
### After benchmark reworking
Actually, the original benchmark tested only one corner case, so I reworked BenchmarkEncodeRGBA, making it closer to real-world usage. Thus, the benchmark became much slower 
```
BenchmarkEncodeRGBA-10            46	  25082351 ns/op	  48.99 MB/s	 1084532 B/op	   55117 allocs/op
```
but anyway, it's still faster than without optimizations:
```
BenchmarkEncodeRGBA-10    	      34	  30837667 ns/op	  39.85 MB/s	 3321862 B/op	  614436 allocs/op
```

Issue #55889, opened 2022-09-27
cmd/compile: some x/sys versions no longer build due to "go:linkname must refer to declared function or variable"


mvdan:
Using an `x/sys` version that is too old for a Go version may result in build errors, and that has clearly confused at least hundreds of Go users in the past year or so:

* https://stackoverflow.com/questions/71507321/go-1-18-build-error-on-mac-unix-syscall-darwin-1-13-go253-golinkname-mus
* https://github.com/golang/go/issues/51706

Given that `x/sys` uses `//go:linkname` directives into unexported Go APIs, it is fairly tightly bound to a set of Go versions. If the Go version is too old or too new, it's entirely possible that the build may fail like the example above on `GOOS=darwin`.

If the `x/sys` version is too old, the solution is pretty easy: `go get golang.org/x/sys` to update the dependency. That is not obvious given the error message, though. And I bet that a good number of those end users are only indirectly depending on `x/sys`, so they might not even know what the library does or why it's being built.

Similarly, if one tries to use a recent version of `x/sys` with an ancient version of Go, I imagine the build may also fail in weird and confusing ways, though this side of the coin doesn't seem to be as common of a problem.

I think we should try to make these errors more intuitive. As a strawman idea, I suggest that we force the current master version of `x/sys` to fail to build on Go 1.17 and older, as well as Go 1.21 and newer, via build tags such as:

```
$ cat go_older.go
//go:build !go1.18
// +build !go1.18

"this version of golang.org/x/sys supports Go 1.18 and Go 1.19; upgrade your Go version or, if you cannot, downgrade x/sys"
$ cat go_newer.go
//go:build go1.21

"this version of golang.org/x/sys supports Go 1.18 and Go 1.19; upgrade x/sys via: go get golang.org/x/sys"
```

The string literals would cause a parse error which would show the string to the user, as a way of customizing the error message:

```
$ go1.16.15 build go_older.go 
go_older.go:4:1: expected 'package', found "this version of golang.org/x/sys supports Go 1.18 and Go 1.19; upgrade your Go version or, if you cannot, downgrade x/sys"
```

We would not error on Go 1.20, as that is the current tip/master version, and we are continuously supporting and testing it. Erroring on Go 1.21 today is perhaps unnecessary, but it's the only way to prepare the current master versions for a future when Go 1.21 is widely used and may not work with today's versions of `x/sys`. When Go 1.20 is released and we start working towards 1.21, the build tags in `x/sys` master would be advanced by one version.

[...]

Issue #51706, opened 2022-03-16
go:linkname must refer to declared function or variable


littlejiancc:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.18 darwin/amd64
</pre>

### Does this issue reproduce with the latest release?
yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/wxj/Library/Caches/go-build"
GOENV="/Users/wxj/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/wxj/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/wxj/go"
GOPRIVATE=""
GOPROXY="https://goproxy.io"
GOROOT="/Users/wxj/sdk/go1.18"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/Users/wxj/sdk/go1.18/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.18"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/wxj/go/src/demo/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/85/j24r836d5ygfhjgsn4j5zr0m0000gn/T/go-build178809849=/tmp/go-build -gno-record-gcc-switches -fno-common"

</pre></details>

### What did you do?

<pre>
$ tree
.
├── 1.s
├── go.mod
├── go.sum
├── inner
│   └── inner.go
└── main.go
</pre>

<pre>
$ cat inner.go
package inner

type name struct {
	name string
}

[...]

PR #55891, opened 2022-09-27
`go:linkname must refer to declared function or variable` is not an error in 1.17

chengjoey:
In version 1.17, `go:linkname must refer to declared function or variable` is not an error
may be should change the comment to 1.18

### DIFF
--- a/src/cmd/compile/internal/noder/noder.go
+++ b/src/cmd/compile/internal/noder/noder.go
@@ -296,8 +296,8 @@ func (p *noder) processPragmas() {
 		}
 		n := ir.AsNode(typecheck.Lookup(l.local).Def)
 		if n == nil || n.Op() != ir.ONAME {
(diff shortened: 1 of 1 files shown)


chengjoey:
#51706

mvdan:
You are targeting a 1.17 branch which hasn't been updated for some time. The right fix is being discussed in https://github.com/golang/go/issues/55889.

chengjoey:
> You are targeting a 1.17 branch which hasn't been updated for some time. The right fix is being discussed in #55889.

I don't quite understand what is the correct fix

In the comments for the 1.17 branch,`go:linkname must refer to declared function or variable` will become an error in version 1.17, but in fact, version 1.18 only throws it as an error

@bcmills cound you take a look when you have time

mdempsky:
Thanks for your interest in helping, but this CL is unnecessary. We don't care about internal comments in obsolete versions of Go.

PR #55932, opened 2022-09-29
container/heap: replace !h.Less(j, i) with h.Less(i, j)

zhiqiangxu:
Seems there's no need to use neg here, it also simplifies understanding a bit.


zhiqiangxu:
Found that !h.Less(j, i) and h.Less(i, j) differs when they're equal.

PR #55984, opened 2022-10-01
net/http: check length of input ipv4 address before for loop

asheikm:


asheikm:
Closing this will raise another PR

PR #55987, opened 2022-10-01
net: check length of input ip address before for loop

asheikm:
It would be good if length of the given string is checked before for loop and return appropriate return value.


asheikm:
As per discussion closing this PR

PR #56069, opened 2022-10-06
sort: O(N/2) for IsSorted function

al8n:
This PR changes the `sort.IsSorted` and `sort.SliceIsSorted` to O(N/2).


carlmjohnson:
This is doing the same amount of work as before. Why should it be any faster? Have you run any benchmarks?

PR #56085, opened 2022-10-06
syscall/js: expose a scope argument in JS interop

romaindoumenc:
Updates #56084

This exposes the scope object in Javascript Go constructor, and makes it available in syscall/js.

Not to be merged as such, this is the basis for a proposal.


romaindoumenc:
Will re-submit with single version, integrating existing comments.

PR #56127, opened 2022-10-10
cmd/go/internal/load: improve comments

sivchari:
isTestFunc checked whether the parameter type is B, M or T before the version of Go1.17.
But, after Go1.18, isTestFunc checks F, too.


gabrielbussolo (approved):
good catch 👍

PR #56248, opened 2022-10-15
archive/zip: fix a bug with zip64 check

kleeon:
Seems to be a bug here. Size of central directory is a 32-bit value so it should be compared to 0xffffffff


kleeon:
okay I'm going to create an issue instead.

PR #56282, opened 2022-10-17
os: document that WriteFile is not atomic

### LINKED ISSUE #56173: os: document that WriteFile is not atomic
As I was replying to https://github.com/golang/go/issues/56172, I was looking at the docs for https://pkg.go.dev/os#WriteFile, which say:

```
// WriteFile writes data to the named file, creating it if necessary.
// If the file does not exist, WriteFile creates it with permissions perm (before umask);
// otherwise WriteFile truncates it before writing, without changing permissions.
func WriteFile(name string, data []byte, perm FileMode) error {
```

I know that this operation isn't atomic: for example, if a write to the file fails half-way through, we may be left with a partial file. That is why alternatives like https://pkg.go.dev/github.com/google/renameio#WriteFile exist.

However, by just reading the docs as a new user, it could be an easy misconception. A lot of the other `os` APIs like `Create`, `Mkdir`, or `Chmod` are atomic at the system level - they only perform a single system call which modifies the filesystem, so we can't be left with partial changes.

Perhaps we should add a godoc note that the API is not atomic. That is, that if an error occurs, the file might be left behind. I imagine this won't matter for many cases, but the warning would be very useful for any Go program which cares about invalid or incomplete states.

cc @ianlancetaylor @stapelberg

RichardoC:
Fixes #56173

### DIFF
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -730,8 +730,7 @@ func ReadFile(name string) ([]byte, error) {
 // WriteFile writes data to the named file, creating it if necessary.
 // If the file does not exist, WriteFile creates it with permissions perm (before umask);
 // otherwise WriteFile truncates it before writing, without changing permissions.
-// Note that WriteFile does not guarantee atomicity, if file writing fails halfway through,
-// the file content will be incomplete
+// Note that WriteFile is not atomic and can have side effects.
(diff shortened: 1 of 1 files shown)


RichardoC:
This adds the wording suggested by @KaithyRookie

Issue #33121, opened 2019-07-15
debug/elf: don't check section size


gregory-m:
### What version of Go are you using (`go version`)?
go1.12.7
also reproducible on tip
### Does this issue reproduce with the latest release?
yes

### What did you do?

https://play.golang.org/p/_5bUxWBh14g

```
package main

import (
	"debug/elf"
	"fmt"
	"strings"
)

func main() {
	data := "\u007fELF\x02\x01\x010000000000000" +
		"\x010000000000000000000" +
		"\x02\x00\x00\x00\x00\x00\x00\x0000000000\x00\x00\x00\x00" +
		"000\x0000"

	_, err := elf.NewFile(strings.NewReader(data))
	if err != nil {
		fmt.Println(err)
	}
}
```

### What did you expect to see?
no panic

### What did you see instead?
```
panic: runtime error: makeslice: len out of range

goroutine 1 [running]:
debug/elf.(*Section).Data(0x44b2c0, 0x43e300, 0x7f6900, 0x206590, 0x109440, 0x7fb880, 0x0, 0x0)
	/usr/local/go/src/debug/elf/file.go:105 +0x40
debug/elf.NewFile(0x15d640, 0x43e2e0, 0x43e2e0, 0x1293, 0x432070, 0x11)
	/usr/local/go/src/debug/elf/file.go:448 +0x1660
main.main()
	/tmp/sandbox898898450/prog.go:15 +0xa0
```

Fund with https://github.com/dvyukov/go-fuzz

odeke-em:
Thank you for this report @gregory-m!

Kindly paging @tklauser @heschik @aclements.

ianlancetaylor:
I'm not really sure how hard debug/elf should work to handle invalid input files.

gregory-m:
I have same question as @ianlancetaylor, because I have 2 more panics in different places.

rockdaboot:
> I'm not really sure how hard debug/elf should work to handle invalid input files.

Here is at least one reason why NewFile() should not crash because of invalidated input.

We have a service that extracts BuildIDs from ELF files provided from different sources (distro packages or arbitrary upload from user).
For this we use NewFile() to load the ELF headers and then extract the BuildID from ".note.gnu.build-id" (or for kernel object we use ".notes").

We already saw some crashes of the service (luckily, it automatically recovers =)).

Now I ran into this again while experimenting with the Go fuzzer.
Here are some details, let me know if you prefer a separate issue.

[...]

PR #56405, opened 2022-10-24
debug/elf: use saferio.SliceCap when decoding ELF sections

ZekeLu:
This avoids allocating an overly large slice for corrupt input.

No test case because the problem can only happen for invalid data. Let
the fuzzer find cases like this.

Updates #33121.

### DIFF
--- a/src/debug/elf/file.go
+++ b/src/debug/elf/file.go
@@ -467,8 +467,12 @@ func NewFile(r io.ReaderAt) (*File, error) {
 	}
 
 	// Read section headers
-	f.Sections = make([]*Section, shnum)
-	names := make([]uint32, shnum)
+	c := saferio.SliceCap((*Section)(nil), uint64(shnum))
+	if c < 0 {
+		return nil, &FormatError{0, "too many sections", shnum}
+	}
+	f.Sections = make([]*Section, 0, c)
+	names := make([]uint32, 0, c)
 	for i := 0; i < shnum; i++ {
 		off := shoff + int64(i)*int64(shentsize)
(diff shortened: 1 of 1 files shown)


rockdaboot (approved):
:+1:

I confirm that this change fixes the issue for me.

rockdaboot:
@ZekeLu Just out of curiosity (and for planning): Will the fix go into 1.19.3 or only into 1.20 ?
Because without the fix we can not always recover. Depending on the `shnum` value, there is either a recoverable panic or a fatal error (fatal error: runtime: out of memory). The error can not be recover()'ed from.

ZekeLu:
Hi @rockdaboot, no backport request is created, so it won't go into 1.19.x.

If you want to make it capable for fuzzing, I think the following commits should be backported too:

- dbd0ce84d7c1da2e788c516c72fef44d5b760337
- b6d5831caa219aa00574fbacb1026585cf11d6b4
- fa13731a45049eca149bdfd807c35d18ece6a85a

I'm not sure whether they are qualified for backporting to the 1.19 branch. Maybe @ianlancetaylor can answer your question.

ianlancetaylor:
I don't think these kinds of changes meet our backport criteria (https://go.dev/wiki/MinorReleases).  Sorry.

rockdaboot:
@ZekeLu Thanks GTN. I fuzz using latest `master`. But I think we have to roll out our own code to parse ELF files (we are mostly interested in getting symbols).

ianlancetaylor:
it shouldn't be too hard to use your own copy of the master version of debug/elf for now.

PR #56406, opened 2022-10-25
net/netip: Need to unmap the Addr in AddrFromSlice due to that there may be an IPv4-mapped IPv6 address

zhlsunshine:
Generally, the code should not assert an IP address is IPv4 or IPv6 via the slice length due to there may be an IPv4-mapped IPv6 address, so need to unmap the Addr in `AddrFromSlice` in `net/netip` package.

This can be reproduced by calling `net.DefaultResolver.LookupNetIP` in `net` package to obtain all interfaces and then each interface address may be in an IPv4-mapped IPv6 address, and this will cause the problem when checking an address's IP family.

### DIFF
--- a/src/net/lookup.go
+++ b/src/net/lookup.go
@@ -255,9 +255,11 @@ func (r *Resolver) LookupNetIP(ctx context.Context, network, host string) ([]net
 	}
 	ret := make([]netip.Addr, 0, len(ips))
 	for _, ip := range ips {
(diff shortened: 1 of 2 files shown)


ZekeLu:
Hi @zhlsunshine, now that https://go.dev/cl/415580 is merged and `net.DefaultResolver.LookupNetIP` won't return IPv4-mapped IPv6 address. Can you verify whether it addresses your  issue? Thanks!

PR #56456, opened 2022-10-27
crypto/tls: exposed underlying net.Conn

dEeds83:
Exposed the underlying net.Conn connection. This can be useful for setting e.g. TCP params.


ZekeLu:
Such kind of change needs a proposal. See https://github.com/golang/proposal.

mateusz834:
I don't know your specific use case, but what is wrong with the [`*Conn.NetConn`](https://pkg.go.dev/crypto/tls#Conn.NetConn) method

PR #56480, opened 2022-10-28
net/http: use Copy in ServeContent if CopyN not needed

costela:
This small PR allows optimizations made in io.Copy (like the use of 
io.WriterTo) to be used in one possible path of http.ServeContent
(in case of a non-Range request).
This, in turn, allows us to skip the buffer allocation in io.Copy.


costela:
Benchmark code: https://gist.github.com/costela/44079c1c742007d355cb3e382d0ac0ef

PR #56604, opened 2022-11-06
sync/atomic: disallow type conversions of atomic.Pointer[T]

### LINKED ISSUE #56603: sync/atomic: atomic.Ponter[T] can be misused with type conversions.
```go
package main

import (
	"math"
	"sync/atomic"
)

type small struct {
	small [64]byte
}

type big struct {
	big [math.MaxUint16 * 10]byte
}

func main() {
	a := atomic.Pointer[small]{}
	a.Store(&small{})

	b := atomic.Pointer[big](a) // type conversion
	big := b.Load()

	for i := range big.big {
		big.big[i] = 1
	}
}

```
This sample (on my system) produces an segfault: 
```
[mateusz@arch aa ]$ go run main.go
signal: segmentation fault (core dumped)
```
So we basically done a unsafe type conversion without using the unsafe package.

mateusz834:
Fixes #56603

### DIFF
--- a/src/sync/atomic/type.go
+++ b/src/sync/atomic/type.go
@@ -35,12 +35,18 @@ func b32(b bool) uint32 {
 	return 0
 }
 
+type noTypeConversion[T any] struct{}
+
 // For testing *Pointer[T]'s methods can be inlined.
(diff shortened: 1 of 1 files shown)


mateusz834:
There was some issue with github, I posted this PR like an hour ago, but it shows that I posted it around now.

PR #56827, opened 2022-11-18
Slightly more efficient heap functions in heap.go

lorenzotinfena:
Sorry I think I duplicated (2 times) this pr, I'm noob.
I have done only a little benchmark, but maybe not coverings all possible cases.
In my benchmark (initialize a a heap) was faster around of 3-4% with this code.

### DIFF
--- a/src/container/heap/heap.go
+++ b/src/container/heap/heap.go
@@ -100,13 +100,19 @@ func up(h Interface, j int) {
 func down(h Interface, i0, n int) bool {
 	i := i0
 	for {
-		j1 := 2*i + 1
-		if j1 >= n || j1 < 0 { // j1 < 0 after int overflow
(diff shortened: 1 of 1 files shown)


ishwar00:
https://github.com/golang/go/blob/d5de62df152baf4de6e9fe81933319b86fd95ae4/src/container/heap/heap.go#L25-L27

Hey @lorenzotinfena, please update the comments accordingly.

Thanks

lorenzotinfena:
> 

Hi, sorry but I don't know what you mean. It seems right to me that comment.
Maybe to be more precise (to count int overflow) I think it can be:
`!h.Less(j, i) for 0 <= i < h.Len() and 2*i+1 <= j <= 2*i+2 and 0 <= j < h.Len()`

PR #56867, opened 2022-11-21
syscall/js: add Scope function

romaindoumenc:
Fixes #56084

This changes allow to pass an arbitrary object from the JS world to the WASM program.

 - if no object is provided, globalThis is returned instead.
 - the scope object is derive from internal jsGo object, avoiding a possible ID clash when the scope is unset (so left as globalThis)
 - test uses a dummy object to perform a simple get/set action.


romaindoumenc:
@neelance : just checking if the patch set in Gerrit is correctly attributed for review – first time there, I find it utterly confusing.
If it’s on your radar, all good. Thank you

PR #56971, opened 2022-11-29
regexp: fix incorrect name of MatchString test

anuraaga:
Currently, the test that checks regexp.MatchString is called TestMatchFunction.
This was probably a typo - there is no behavior change in this commit.


anuraaga:
Sorry nevermind, reread the code and made sense

Issue #48621, opened 2021-09-25
unicode: update to version 14.0


180909:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.17.1 darwin/amd64
</pre>

### Does this issue reproduce with the latest release?
yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN="/Users/wangjiahua/go/bin"
GOCACHE="/Users/wangjiahua/Library/Caches/go-build"
GOENV="/Users/wangjiahua/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/wangjiahua/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/wangjiahua/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.17.1"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/vj/b75wqh2n5l324xzv_88g14vm0000gn/T/go-build2079872506=/tmp/go-build -gno-record-gcc-switches -fno-common"
GOROOT/bin/go version: go version go1.17.1 darwin/amd64
GOROOT/bin/go tool compile -V: compile version go1.17.1
uname -v: Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64
[...]

PR #57265, opened 2022-12-12
all: upgrade Unicode from 13.0.0 to 15.0.0

cowtoolz:
Update unicode/tables.go to reflect changes in the Unicode Standard up to
Unicode 15.0.0, released 13 Sept 2022.

In order to accommodate this update, strconv/isPrint has been updated to
reflect changes in printable characters.

Also changed is template/exec_test.go for both text and html packages- in
the test "TestJSEscaping", rune U+FDFF was used as a placeholder for an
unprintable character. This codepoint was assigned and made printable in
Unicode 14.0.0, breaking this test. It has been replaced with the assigned
and never-printable U+FFFE to fix the test and provide resiliency in the
future.

This upgrade bypasses Unicode 14.0.0, but is compatible.

Updates https://github.com/golang/go/issues/48621
Fixes https://github.com/golang/go/issues/55079


Imacheaterandliar21 (changes requested):
c8885cab7a0c23632e1e5a433b1e8d5634a45a30

PR #57403, opened 2022-12-20
net/http: fix when writeLoop exited still has request in writech

### LINKED ISSUE #49621: net/http: Request body is not closed by Transport.RoundTrip on some net.Conn errors

<pre>
$ go version
go version go1.17.3 linux/amd64
</pre>

Yes

$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/eg/.cache/go-build"
GOENV="/home/eg/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/eg/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/eg/go"
GOPRIVATE=""
GOROOT="/usr/lib/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOVCS=""
GOVERSION="go1.17.3"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/usr/lib/go/src/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build2074510398=/tmp/go-build -gno-record-gcc-switches"
</pre></details>

### What did you do?

1. Wrap `net.Listener` with error injections
2. Run `http.Server` using such listener
3. Run many concurrent requests targeting such server
4. On every request, at the client side, wait for request body to be closed

Kind of a minimal example to run described steps: https://play.golang.com/p/lku8lEgiPu6
Yeah, I know that reproducer feels big, but the most of it is to create error injections actually, but
main part (see main fn) is straightforward and simple.

I expect that `net/http` eventually closes `Request.Body` as it is told in documentation for RoundTripper: https://pkg.go.dev/net/http#RoundTripper

### What did you see instead?

Specified example program hangs indefinitely while waiting for `Request.Body` to be closed.

### Investigation

* It is possible that encountered behavior has something to deal with `errServerClosedIdle` error inside `net/http`, because we have met described hangs only after receiving that error from `Client.Do`.
* Also the fact that I am doing `Post` request may be important because `net/http` does not retry `Post` (non-idepotent) request on such error.

In case next assumption is true:

Once request execution got to `func (pc *persistConn) roundTrip(req *transportRequest)` call, `Request.Body` closure is expected to be done by `func (pc *persistConn) writeLoop()`.

Then it seems like there is a possibility of not closing body in the `writeLoop` in case `pc` is closed concurrently.
Imagine:

```
                   ---------------------------------------------------------------------------------------> time
gor1               pc.writech <- writeRequest{...}                                   
gor2                   pc.closeWithError(...)
writeLoop                                                              select { case <-pc.closech: return }
                                                                                              
```

So request was sent successfully into `pc.writech` (channel has buffer of size 1), but won't be ever read from it because `writeLoop` exited.

Related places in the code:
1. https://github.com/golang/go/blob/f6591839727e09cc5cb11d08b333fd2386e8aa1b/src/net/http/transport.go#L2383
2. https://github.com/golang/go/blob/f6591839727e09cc5cb11d08b333fd2386e8aa1b/src/net/http/transport.go#L2595

Aoang:
When there is still a request in writech after
writeLoop exits, the request body will not be
closed and will not be retried.
The request is not sent to the server, so it can
be safely retried without regard to idempotence.

Fixes #49621

### DIFF
--- a/src/net/http/transport.go
+++ b/src/net/http/transport.go
@@ -719,6 +719,13 @@ func (pc *persistConn) shouldRetryRequest(req *Request, err error) bool {
 		// can "rewind" the body with GetBody.
 		return req.outgoingLength() == 0 || req.GetBody != nil
 	}
+	if len(pc.writech) != 0 {
+		// Request was sent successfully into writech, but isn't read from it
+		// because writeLoop exited.
+		// In this case, you can retry without considering idempotence.
+		// see https://github.com/golang/go/issues/49621
+		return true
+	}
 	if !req.isReplayable() {
 		// Don't retry non-idempotent requests.
 		return false
--- a/src/net/http/transport.go
+++ b/src/net/http/transport.go
@@ -719,13 +719,6 @@ func (pc *persistConn) shouldRetryRequest(req *Request, err error) bool {
 		// can "rewind" the body with GetBody.
 		return req.outgoingLength() == 0 || req.GetBody != nil
 	}
-	if len(pc.writech) != 0 {
-		// Request was sent successfully into writech, but isn't read from it
-		// because writeLoop exited.
-		// In this case, you can retry without considering idempotence.
-		// see https://github.com/golang/go/issues/49621
-		return true
-	}
 	if !req.isReplayable() {
 		// Don't retry non-idempotent requests.
 		return false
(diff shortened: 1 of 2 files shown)


Aoang:
https://golang.org/cl/461675

PR #57460, opened 2022-12-25
cmd/link: remove objIdx structure

jquirke:
This indirection appears to be unnecessary for linking or linker debugging, and therefore hinders readability.

Since all *oReaders are added to loader.objs *only* via the Preload -> addObj path, before any symbols are examined, there is no possible way the "i" member of this structure is still useful; and is likely a remnant of an earlier design.

### DIFF
--- a/src/cmd/link/internal/loader/loader.go
+++ b/src/cmd/link/internal/loader/loader.go
@@ -91,11 +91,6 @@ type oReader struct {
 // non-package symbols).
 func (r *oReader) NAlldef() int { return r.ndef + r.nhashed64def + r.nhasheddef + r.NNonpkgdef() }
 
-type objIdx struct {
-	r *oReader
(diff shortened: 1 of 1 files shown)


src/cmd/link/internal/loader/loader.go
```diff
@@ -184,7 +179,7 @@ type symAndSize struct {
 //     of the symbol gets the same global index as the first appearance.
 type Loader struct {
 	start       map[*oReader]Sym // map from object file to its start index
-	objs        []objIdx         // sorted by start index (i.e. objIdx.i)
+	objs        []*oReader       // sorted by start index (i.e. objIdx.i)
```
omdxp, line 182: comments should be updates too for readability

omdxp (approved):

PR #57717, opened 2023-01-10
invalid parameters were omitted mode

wjw1758548031:
invalid parameters were omitted mode


T2Player (commented):
If you remove "mode" param from "importFrom", so you should remove it also from "ImportFrom" at string 619.
P.S. Naming different functions the same name with only register of one symbol difference is very bad practice.

PR #57809, opened 2023-01-16
os: add test for UserCacheDir

### LINKED ISSUE #57638: os: UserCacheDir and UserConfigDir have no tests
`os.UserCacheDir` was added in https://go.dev/cl/78835 for #22536.

In examining #57630, it seems that this function is not covered by any tests at all. There were none added in https://go.dev/cl/78835, nor in https://go.dev/cl/120757.

It seems like we should at least have some test that verifies that `UserCacheDir` reports either an error or a directory in which the user is allowed to create files. 😅

(CC @rsc @robpike @ianlancetaylor @bradfitz @griesemer for `os`)

ja7ad:
Fixes #57638

### DIFF
--- a/src/os/os_test.go
+++ b/src/os/os_test.go
@@ -2554,17 +2554,8 @@ func TestUserCacheDir(t *testing.T) {
 		t.Fatal("UserCacheDir return empty string without error")
 	}
 
-	if err != nil {
+	if err != nil && IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


bcmills:
I notice that a new PatchSet was uploaded — if this is ready for another review pass, please mark the Gerrit comment threads as resolved and use the Gerrit `Reply` button to publish your replies.

PR #57833, opened 2023-01-17
os: use GetTempPath2 on Windows if available

nwnt:
This generates GetTempPath2 together with RtlGetNtVersionNumbers. The latter is needed to determine if the running Windows has GetTempPath2 by comparing it against the minimum build number that has the API. 

RtlGetNtVersionNumbers was generated into syscall/windows since syscall is locked down.

Fixes #56899


kaixuanfeng:
好滴，已收到

PR #57980, opened 2023-01-25
os: use GetTempPath2 on Windows if available

nwnt:
This generates GetTempPath2. Go now tries to determine if the windows it runs on has GetTempPath2 by finding it only once at the loading time. If GetTempPath2 exists, it sets the flag so that any calls to tempDir will use it. If it doesn't exist, Go then uses GetTempPath.

GetTempPath2 was generated into internal/syscall/windows since syscall is locked down.

Fixes #56899


bcmills:
If the Gerrit comments have been addressed, please mark them as resolved in Gerrit (and don't forget to publish the replies).

Issue #58237, opened 2023-02-02
net/http: ResponseController panic


komuw:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.20 linux/amd64
</pre>

### Does this issue reproduce with the latest release?
yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home//.cache/go-build"
GOENV="/home//.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home//go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home//go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.20"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home//Downloads/cool/go.mod"
GOWORK=""
CGO_CFLAGS="-O2 -g"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-O2 -g"
CGO_FFLAGS="-O2 -g"
CGO_LDFLAGS="-O2 -g"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=/run/user/1000/go-build839916512=/tmp/go-build -gno-record-gcc-switches"
</pre></details>

### What did you do?
Run the program; https://go.dev/play/p/Alu2e-LdzqX?v=gotip
and then use curl to access it `curl -vkL https://127.0.0.1:56782/`

### What did you expect to see?
No panic, possibly an error

### What did you see instead?
```sh
listening on  :56782
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7471f5]

goroutine 22 [running]:
[...]

PR #58282, opened 2023-02-02
net/http: add ResponseController http2 request without body read deadline test

AlexanderYastrebov:
Requires CL 464936

For #58237

### DIFF
--- a/src/net/http/responsecontroller_test.go
+++ b/src/net/http/responsecontroller_test.go
@@ -326,3 +326,18 @@ func testResponseControllerEnableFullDuplex(t *testing.T, mode testMode) {
 	}
 	pw.Close()
 }
+
+func TestIssue58237(t *testing.T) {
+	cst := newClientServerTest(t, http2Mode, HandlerFunc(func(w ResponseWriter, req *Request) {
+		ctl := NewResponseController(w)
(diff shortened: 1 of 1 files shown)


src/net/http/responsecontroller_test.go
```diff
@@ -263,3 +263,18 @@ func testWrappedResponseController(t *testing.T, mode testMode) {
...
 	defer res.Body.Close()
 }
+
+func TestIssue58237(t *testing.T) {
+	cst := newClientServerTest(t, http2Mode, HandlerFunc(func(w ResponseWriter, req *Request) {
+		ctl := NewResponseController(w)
+		if err := ctl.SetReadDeadline(time.Now().Add(1 * time.Millisecond)); err != nil {
+			t.Errorf("ctl.SetReadDeadline() = %v, want nil", err)
+		}
+		time.Sleep(10 * time.Millisecond)
```
komuw, line 273: since the `1ms` and `10ms` are related, my I suggest:
```suggestion
	cst := newClientServerTest(t, http2Mode, HandlerFunc(func(w ResponseWriter, req *Request) {
		ctl := NewResponseController(w)
		duration := 1 * time.Millisecond
		if err := ctl.SetReadDeadline(time.Now().Add(duration)); err != nil {
			t.Errorf("ctl.SetReadDeadline() = %v, want nil", err)
		}
		time.Sleep(10 * duration)
```

PR #58287, opened 2023-02-03
cmd/go: mod vendor: fixed checking for vendor directory

barshow:
In our case I was using a "vendor-replace" directory which was incorrectly flagged as being in the vendor directory.


bcmills:
I notice that a new patch was uploaded. If this is ready for review, please mark the comment threads in Gerrit as resolved and publish the replies using the Gerrit `Reply` button.

PR #58463, opened 2023-02-10
add introduction & how to use

jimmysoccer:
1. add a brief introduction of what is the Go language
2. add two tutorials about how to print a hello world message and how to build a simple web server

PR #58479, opened 2023-02-11
update Readme file with more reability

jimmysoccer:
When I read this readme, I felt confused and didn't quite understand what the Go Langauge is and no idea on how to start using it. In order to make new users like me to understand it quickly, I propose the following changes. 

1. An introduction part describing what the Go Language is with a more specific functionalities. It briefly introduce the similarity between Go and C, advantages of Go, and several main functionalities of using Go. In addtion, it also includes a hint to remind users to choose the correct version and set up the correct envrionment in case of unexpected errors while in use.

2. A simple tutorial teaching new users how to run and use Go language, including a guide to print the 'hello world' message to console, and a guide to build a simple local web server. 

New users would have a better understanding of what the Go language is and using methods of it by applying these modifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,14 @@ There is a mirror of the repository at https://github.com/golang/go.
 Unless otherwise noted, the Go source files are distributed under the
 BSD-style license found in the LICENSE file.
 
+### Introduction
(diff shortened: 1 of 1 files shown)


seankhliao:
Duplicate of #58463

PR #58497, opened 2023-02-13
syscall/js: correct the wrong name in error for Value.Length method

xoba:
when i call "Length" method on a js.Value object, and an error results, that error claims to originate from the "SetIndex" method, which is incorrect. instead, it actually originates from the "Length" method. thus, i changed the string from "Value.SetIndex" to "Value.Length". thanks for considering!


xoba:
this bug still exists as of 2024-06-10, over one year later. can we get this fix merged? if so, how? thanks!

PR #58779, opened 2023-02-28
io/ioutil: add doc links for deprecated function

SilverRainZ:
Go 1.19 introduced doc links (https://go.dev/doc/comment#doclinks).
It will be convenient when we can directly jump to the suggested
function when the original function is deprecated.


dmitshur:
Closing because this PR has been merged via [CL 471915](https://go.dev/cl/471915) as commit dfe4591fc4d2d66da7740de18b32761e7a20691c.

(Normally `gerritbot` does this, but sometimes it needs help due to #35867.)

PR #58847, opened 2023-03-03
net/http: use initial request protocol version for redirect requests

AlexanderYastrebov:
This is useful for a logging RoundTripper that uses httputil.DumpRequest and it also matches NewRequest behavior.


AlexanderYastrebov:
Closing, see gerrit discussion

PR #58913, opened 2023-03-07
crypto/tls: set const maxUselessRecords to 32 (the same with OpenSSL)

RPRX:
In OpenSSL:

https://github.com/openssl/openssl/blob/ac57336cd258e0432ffa485615d11c7c7ecfe81a/ssl/record/methods/tls_common.c#L513

In BoringSSL:

https://github.com/google/boringssl/blob/082e953a134ad423a00b8859f9daf5708e729260/ssl/tls_record.cc#L128


RPRX:
In BoringSSL：

https://github.com/google/boringssl/blob/082e953a134ad423a00b8859f9daf5708e729260/ssl/tls_record.cc#L128

PR #59014, opened 2023-03-13
runtime/debug: replace "GOARM64" with "GOARM" in docs

tianon:
The former does not exist (yet) but the latter very much exists and is in active use.


tianon:
As far as I can find, https://github.com/golang/go/issues/38485#issuecomment-647762275 is the only reference beyond this doc to `GOARM64` as a variable (which isn't a thing yet, and that comment is just the suggestion that it could be made a thing if it were necessary). :smile:

PR #59018, opened 2023-03-14
syscall: add support for setns after fork

cpuguy83:
This adds a Namespaces field to Linux's SysProcAttr type.
When set, these namespaces will be entered after fork and before exec.
The format for this is [ns name]=[path], e.g. mnt=/some/path.
    
This allows users to exec a new process in a pre-defined set of
namespaces without having to resort to hacks or re-execs to bootstrap
these namespaces.

Closes #56680


cpuguy83:
I noticed that `SYS_SETNS` is not defined on amd64 (but it is on other arches) so I generated it just so I could test.
I'm not sure what the right approach is here.

PR #59301, opened 2023-03-28
errors: improve the godoc for Join and Unwrap

dnephin:
Document that errors returned by Join always implement Unwrap []error.

Explicitly state that Unwrap does not unwrap errors
with an Unwrap() []error method.

### DIFF
--- a/src/errors/join.go
+++ b/src/errors/join.go
@@ -10,6 +10,8 @@ package errors
 // The error formats as the concatenation of the strings obtained
 // by calling the Error method of each element of errs, with a newline
 // between each string.
+//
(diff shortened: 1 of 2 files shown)


src/errors/join.go
```diff
@@ -10,6 +10,8 @@ package errors
 // The error formats as the concatenation of the strings obtained
 // by calling the Error method of each element of errs, with a newline
 // between each string.
+//
+// A non-nil error returned by Join implements the Unwrap() []error method.
```
dnephin, line 14: Would it be appropriate to be more explicit here, and say something about "the returned error can not be unwrapped by [Unwrap]" ?

Issue #58001, opened 2023-01-25
security: fix CVE-2022-41724


rolandshoemaker:
This is a PRIVATE issue for CVE-2022-41724, tracked in http://b/261456191 and fixed by http://tg/1679436.

/cc @golang/security and @golang/release

rolandshoemaker:
@gopherbot please open backport issues.

catenacyber:
Is there a reproducer available for this issue @rolandshoemaker ?

PR #59425, opened 2023-04-04
crypto/tls: fix PSK binder calculation

### LINKED ISSUE #59424: crypto/tls: TLSv1.3 connection fails with invalid PSK binder

The bug impacts following versions
 
* Go1.19.6 or later 
* Go1.20.1 or later

Yes

The bug is not OS dependent

### What did you do?

1. I have written TLSv1.3 client and server written in go.
2. In the server I have set `CurvePreferences` with non-default list of curves.
3. In the client I'm using default list of curves.
4. In the client I have enabled TLS session resumption by setting `ClientSessionCache`

When the client establishes second TLS session, I expected the TLS handshake to succeed.

### What did you see instead?

When the client establishes second TLS session, the server will reject it with alert `invalid PSK binder`.

tsaarni:
When server and client have mismatch in curve preference, the server will 
send HelloRetryRequest during TLSv1.3 PSK resumption. There was a bug
introduced by Go1.19.6 or later and Go1.20.1 or later, that makes the client
calculate the PSK binder hash incorrectly. Server will reject the TLS 
handshake by sending alert: invalid PSK binder.

Fixes #59424

### DIFF
--- a/src/crypto/tls/handshake_client_test.go
+++ b/src/crypto/tls/handshake_client_test.go
@@ -1028,6 +1028,27 @@ func testResumption(t *testing.T, version uint16) {
 	deleteTicket()
 	testResumeState("WithoutSessionTicket", false)
 
+	// In TLS 1.3, HelloRetryRequest is sent after incorrect key share.
+	// See https://www.rfc-editor.org/rfc/rfc8446#page-14.
+	if version == VersionTLS13 {
+		deleteTicket()
+		serverConfig = &Config{
+			// Use a different curve than the client to force a HelloRetryRequest.
+			CurvePreferences: []CurveID{CurveP521, CurveP384, CurveP256},
+			MaxVersion:       version,
+			Certificates:     testConfig.Certificates,
+		}
+		testResumeState("InitialHandshake", false)
+		testResumeState("WithHelloRetryRequest", true)
+
+		// Reset serverConfig back.
+		serverConfig = &Config{
+			MaxVersion:   version,
+			CipherSuites: []uint16{TLS_RSA_WITH_RC4_128_SHA, TLS_ECDHE_RSA_WITH_RC4_128_SHA},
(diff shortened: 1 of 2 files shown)


tsaarni:
The bug that this PR attempts to fix is on the TLSv1.3 client handshake code.  Only TLSv1.3 clients are impacted.  Servers are not impacted.  TLSv1.2 implementation is not impacted.

The bug was introduced by https://github.com/golang/go/issues/58001

The bug impacts TLS clients compiled with with Go1.19.6 or later or Go1.20.1 or later. 

For the client to be impacted, client must have explicitly enabled TLS resumption by setting e.g. 
ClientSessionCache: tls.NewLRUClientSessionCache(num) 
in client's tls.Config structure.  Client also needs to hold the SSL context and re-use it for multiple requests to triggers resumption

The bug happens in PSK resumption when client and server having incompatible preferred EC curve.  For example, it can happen when the server has set 

CurvePreferences: []tls.CurveID{tls.CurveP521, tls.CurveP384, tls.CurveP256}

but client uses the Go default list: client will send ClientHello with x25519.  Since server preferences does not include x25519, server will respond with HelloRetryRequest and set secp512r1 from CurvePreferences according to chapter "2.1.  Incorrect DHE Share" https://www.rfc-editor.org/rfc/rfc8446#page-14. 

The reception of HelloRetryRequest response from the server will trigger the bug in client when above conditions are true.  There is a typo in processHelloRetryRequest() which causes ClientHello message to be sent to server with invalid PSK binders hash. 

When server will process the new ClientHello with invalid PSK binders field, it will abort the handshake with "invalid PSK binder" alert.

dmitshur:
Merged as [CL 481955](https://go.dev/cl/481955) (commit 2c70690451f1484607a9172a4c24f78ae832dcb0).

PR #59457, opened 2023-04-05
slices: add Transform to slices.go

FGasper:


FGasper:
Maintainers: I’ll add a test if you’re agreeable to this addition.

Thank you for your consideration.

PR #59644, opened 2023-04-14
doc: remove extra comma

crisman:
Removed an extra comma before an or.


crisman:
I can not find the reference I though I remembered on why that comma was bad.  Closing.

PR #59752, opened 2023-04-21
runtime: exit if madvise is a nonexistent system call

ioworker0:
madvise is not specified by any standards.

Since Linux 3.18, support for this system call is optional,
depending on the setting of the CONFIG_ADVISE_SYSCALLS
configuration option. See https://github.com/torvalds/linux/commit/d3ac21cacc24790eb45d735769f35753f5b56ceb

If we attempt to invoke a nonexistent system call, then it
will return -1, and errno will be set to ENOSYS. See https://github.com/torvalds/linux/blob/2af3e53a4dc08657f1b46f97f04ff4a0ab3cad8d/include/uapi/asm-generic/errno.h#L18

### DIFF
--- a/src/internal/testenv/exec.go
+++ b/src/internal/testenv/exec.go
@@ -19,11 +19,65 @@ import (
 // HasExec reports whether the current system can start new processes
 // using os.StartProcess or (more commonly) exec.Command.
 func HasExec() bool {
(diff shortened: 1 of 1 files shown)


ioworker0:
1

PR #59793, opened 2023-04-24
runtime: add a alignment check

ioworker0:
The Linux implementation requires that the address addr be
page-aligned, and allows length to be zero.

See Linux notes:
https://man7.org/linux/man-pages/man2/madvise.2.html


ioworker0:
It's a nice experience, I really appreciate for everyone!

PR #60026, opened 2023-05-06
errors: optimize *joinError's Error method for less allocation and faster execution

callthingsoff:
Handle the case of one error at the beginning.
Use unsafe.String to avoid memory allocation when converting byte slice to string.

### DIFF
--- a/src/errors/join.go
+++ b/src/errors/join.go
@@ -38,11 +38,14 @@ type joinError struct {
 }
 
 func (e *joinError) Error() string {
-	var b []byte
-	for i, err := range e.errs {
-		if i > 0 {
-			b = append(b, '\n')
-		}
+	// Assert len(e.errs) > 0.
+	if len(e.errs) == 1 {
+		return e.errs[0].Error()
+	}
+
+	b := []byte(e.errs[0].Error())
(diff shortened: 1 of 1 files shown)


src/errors/join.go
```diff
@@ -38,11 +38,14 @@ type joinError struct {
...
-			b = append(b, '\n')
-		}
+	// Assert len(e.errs) > 0.
+	if len(e.errs) == 1 {
+		return e.errs[0].Error()
+	}
+
+	b := []byte(e.errs[0].Error())
+	for _, err := range e.errs[1:] {
+		b = append(b, '\n')
```
willfaught, line 48: If errs len is 0, the result will not be "".

gocurr: "Join returns nil if every value in errs is nil." 
```go
	if n == 0 {
		return nil
	}
	e := &joinError{
		errs: make([]error, 0, n),
	}
```

gocurr: The return of errors.Join:
1. nil, the Error method returns ```"<nil>"```.
2. a *joinError, the "errs" in it must not be empty.

Vivekgaddigi (approved):
LGTM

PR #60124, opened 2023-05-11
net/http/httptest: remove unnecessary creation of http.Transport

ZekeLu:
In (*Server).StartTLS, it's unnecessary to create an http.Client
with a Transport, because a new one will be created with the
TLSClientConfig later.


mimani68 (approved):
Sound current modification could helps to improve performance.

Why is `test` result?

ZekeLu:
> Why is `test` result?

Sorry, but I don't understand your question. Can you elaborate the question?

And please note that the golang project uses Gerrit for code review.

PR #60155, opened 2023-05-12
testing: clarify comments on runCleanup

evanj:
The comment on runCleanup states "If catchPanic is true ...", but there is no catchPanic argument or variable. This was introduced in commit 998cbe29832a, which also introduced the panicHandling type. This was changed during code review, but the comment was missed. See:

https://go-review.googlesource.com/c/go/+/214822/4..6/src/testing/testing.go#b797

### DIFF
--- a/src/testing/testing.go
+++ b/src/testing/testing.go
@@ -1300,7 +1300,7 @@ func (c *common) Setenv(key, value string) {
 	}
 }
 
-// panicHanding is an argument to runCleanup.
+// panicHanding controls the panic handling used by runCleanup.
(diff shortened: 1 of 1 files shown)


evanj:
Github appears to have mangled my initial commit message (it had line breaks in it originally!) Rather than try to reverse engineer that, I started a change using Gerrit, since I needed to set it up for another documentation change I just made. I'm abandoning this in favor of https://go-review.googlesource.com/c/go/+/538695 

My apologies for the duplication due to me not understanding these tools very well.

PR #60194, opened 2023-05-15
os: add a FindProcess implementation for UNIX

### LINKED ISSUE #34396: os: FindProcess should document how to test if a process is alive since on UNIX it always returns a process

<pre>
$go version go1.12.6 linux/amd64
</pre>

Yes, Documentation issue, see live doc at:

https://golang.org/pkg/os/#FindProcess

$ go env
$ go env
GOARCH="amd64"
GOHOSTARCH="amd64"
GOHOSTOS="linux"
</pre></details>

### What did you do?

Tried to find out if a process exists using os.FindProcess()

Read the documentation to find that the FindProcess call always returns a nil error on unix systems so it doesn't really "find" the process:

```
On Unix systems, FindProcess always succeeds and returns a Process for the given pid, regardless of whether the process exists.
```

Found closed documentation ticket saying you can send a signal to find if the process really exists or not:

https://github.com/golang/go/issues/14146#issuecomment-176888204

More helpful information in the documentation about how someone might actually test for a process existing, for example as a starter on Linux:

```
	// error is always nil on Unix systems, just creates a process object
	p, _ := os.FindProcess(pid)
	// send SIGCONT (on Linux) which is a safe signal to the process to test if it exists
	err = p.Signal(syscall.SIGCONT)
```

### What did you see instead?

No help without how the user might find out that a process really exists. Just a comment that it doesn't actually tell you if the process exists:

```
On Unix systems, FindProcess always succeeds and returns a Process for the given pid, regardless of whether the process exists.
```

mpldr:
The current implementation of always returning a process can lead to unexpected behavior and is generally not very helpful.

This commit adds a simple check whether the process is still alive or not, as described in the POSIX standard, by attempting to send a null signal to the process.

Fixes #34396

Link: https://pubs.opengroup.org/onlinepubs/9699919799/functions/kill.html

### DIFF
--- a/src/os/exec.go
+++ b/src/os/exec.go
@@ -84,9 +84,6 @@ func Getppid() int { return syscall.Getppid() }
 //
 // The Process it returns can be used to obtain information
 // about the underlying operating system process.
-//
-// On Unix systems, FindProcess always succeeds and returns a Process
-// for the given pid, regardless of whether the process exists.
 func FindProcess(pid int) (*Process, error) {
 	return findProcess(pid)
 }
--- a/src/os/exec_unix.go
+++ b/src/os/exec_unix.go
@@ -92,9 +92,15 @@ func (p *Process) release() error {
 	return nil
 }
 
-func findProcess(pid int) (p *Process, err error) {
-	// NOOP for unix.
-	return newProcess(pid, 0), nil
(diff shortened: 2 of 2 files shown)


mpldr:
I have tested it for linux/amd64 and openbsd/386 on which it works as expected.

PR #60215, opened 2023-05-16
errors: clarify that Join doesn't extend already-joined errors

### LINKED ISSUE #60209: errors: Using `errors.Join` on a `joinError` should expand the existing joined error

<pre>
$ go version
go version go1.20.4 linux/amd64
</pre>

Yes

### Description

Consider this code ([Playground](https://go.dev/play/p/oVfB_zqrjSF))

```go
var err error
err = errors.Join(err, errors.New("err1"))
err = errors.Join(err, errors.New("err2"))
err = errors.Join(err, nil)

errs := err.(interface{ Unwrap() []error }).Unwrap()
fmt.Println(len(errs), errs)

// Output:
// 1 [err1
// err2]
```

This happens because when calling `errors.Join`, it does not check if one of the errors is a `joinError` already, so it creates a new one every time.

1. After `err = errors.Join(err, errors.New("err1"))`, `err` is a `joinError` with 1 item inside (and that's expected)
2. After  `err = errors.Join(err, errors.New("err2"))`, `err` is a `joinError` in which item 1 is a `joinError` itself, and item 2 is the error.
3. After `err = errors.Join(err, nil)`, `err` is a `joinError` that contains 2 `joinError`'s (and the first one contains a `joinError` itself)

We found this bug while trying to implement a pattern where we collect all errors from a channel. We have N goroutines running in parallel, and we wanted to use this pattern to both collect errors (which could be nil) and wait for all goroutines to complete:

```go
errCh := make(chan error, 3)
errCh<-errors.New("err1")
errCh<-errors.New("err2")
errCh<-nil

var err error
for i := 0; i < 3; i++ {
	err = errors.Join(err, <-errCh)
}
```

In the case above, calling `Unwrap() []error` on the result lead to unexpected results because of errors being wrapped multiple times in `joinError` objects

ItalyPaleAle:
Updates the documentation for Join to explain that it does not append to an existing joinError (i.e. the result of calling Join), and if one of the errors is a joinError, it will be wrapped again. Fixes #60209

It also adds a test to provide an example of said behavior, as well as formalizing it in the API's contract.

### DIFF
--- a/src/errors/join.go
+++ b/src/errors/join.go
@@ -12,6 +12,11 @@ package errors
 // between each string.
 //
 // A non-nil error returned by Join implements the Unwrap() []error method.
+//
+// Calling Join on an error that was previously returned by Join wraps
+// the error again. Consequently, calling Unwrap() []error on the result
+// of Join(Join(err1, err2), err3) returns a slice with 2 items, of which
+// the first one implements Unwrap() []error too.
 func Join(errs ...error) error {
 	n := 0
 	for _, err := range errs {
--- a/src/errors/join_test.go
+++ b/src/errors/join_test.go
@@ -70,3 +70,48 @@ func TestJoinErrorMethod(t *testing.T) {
 		}
 	}
 }
+
+func TestJoinWithJoinedError(t *testing.T) {
+	err1 := errors.New("err1")
+	err2 := errors.New("err2")
+
+	var err error
+	err = errors.Join(err, err1)
+	if err == nil {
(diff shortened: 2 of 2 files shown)


ItalyPaleAle:
The goal of this PR is to explain that patterns such as this one are not supported:

```go
// Assume that errCh is populated by a separate goroutine
errCh := make(chan error, 3)
errCh<-errors.New("err1")
errCh<-errors.New("err2")
errCh<-nil

var err error
for i := 0; i < 3; i++ {
	err = errors.Join(err, <-errCh)
}
```

PR #60231, opened 2023-05-16
net/http/httptest: add support for http.ResponseController to ResponseRecorder

dunglas:
Proposal #54136 (implemented in CL 436890 which is part of Go 1.20) added the "http".ResponseController type, which allows manipulating per-request timeouts. This is especially useful for programs managing long-running HTTP connections such as Mercure.

However, testing HTTP handlers leveraging per-request timeouts is currently cumbersome (even if doable) because
"net/http/httptest".ResponseRecorder isn't compatible yet with "http".ResponseController.

This patch makes ResponseRecorder compatible with
"http".ResponseController.

All new methods are part of the contract that response types must honor to be usable with "http".ResponseController.

Deadlines are recorded but not enforced to prevent flaky tests.

Fixes #60229.


src/net/http/httptest/recorder.go
```diff
@@ -42,9 +45,18 @@ type ResponseRecorder struct {
 	// Flushed is whether the Handler called Flush.
 	Flushed bool
 
+	// ReadDeadline is the write deadline that has been set using
```
pborreli, line 48:
```suggestion
	// ReadDeadline is the read deadline that has been set using
```

PR #60472, opened 2023-05-27
fmt.doc: formatting functions will round the number according to the specified precision

Azhovan:


Azhovan:
Close the PR because factors like system architecture, etc can impact the behavior outside of the control of the  `fmt` package: 

more details on system's architecture:

On x86 processors, the default rounding mode is typically "round to nearest, ties to even."  
On some ARM processors, the default rounding mode may be "round towards zero," 
...

PR #60478, opened 2023-05-28
cmd/go/internal/modcmd/graph: add option for selected versions graph

AdamDev:
The current issue is that when using `go mod graph`, it generates packages that are not actually selected by the [Minimal version selection (MVS)](https://go.dev/ref/mod#minimal-version-selection) process. As a result, it creates a significantly large graph with packages that are not in use or downloaded.

This pull request aims to improve the graph's accuracy by cleaning it up and retaining only the packages with versions that were chosen by MVS. This issue primarily occurs in 'go.mod' files with a Go version of 1.16 and below.

To enable this feature, you can use the '-selectedVersions' flag with the 'go mod graph' command. For example:
`go mod graph -selectedVersions`


AdamDev:
**Example for my use case.**

Go mod file:
```
module example
go 1.16
require google.golang.org/grpc v1.30.0
```

**Attached are 2 generated graphs:**
Graph with the new flag of selected versions - [graph_with_selected_flag.txt](https://github.com/golang/go/files/11590318/graph_with_selected_flag.txt)
Original graph - [original_graph.txt](https://github.com/golang/go/files/11590319/original_graph.txt)

In the original graph, we can see multiple versions of `google.golang.org/grpc` (v1.19.0, v1.23.0, v1.25.1, v1.30.0)
whereas only one version was selected and downloaded 1.30.0.

And in the new graph, we can see only the selected version.

PR #60484, opened 2023-05-29
bufio.Reset: when argument io.Reader is nil, Reader initializes the internal buffer to the default size

Azhovan:
When the Reset function is called with a nil argument, The underlying reader's buffer using the defaultBufSize should be reset.


Azhovan:
close the PR, I made an error. apparently, there was a change in my master branch, caused a bug and I got confused by that.

Issue #46554, opened 2021-06-03
internal/fuzz: add missing supported GOOS to internal/fuzz/sys_posix.go [freeze exception]


dchapes:
Currently `src/internal/fuzz/sys_posix.go` only has build constraints for darwin linux.
When I add the missing `freebsd` constraint it works perfectly fine for me on freebsd/amd64 (FreeBSD 11.4).
I imagine it also works on the other BSDs as well, they are all reasonably POSIX compliant.

In future, in announcements such as [the one on the Go Blog](https://blog.golang.org/fuzz-beta), *please* include a sentence somewhere if the list of GOOS/GOARCH supported is limited! Believe it or not, there is a thriving world outside of Mac and Linux. It's a pain in the butt to go through the effort of downloading the dev.fuzz branch on a slow link to get nothing but an uninformative `panic: not implemented` and a stack trace!

dr2chase:
I don't think it was tested across all the other OSes (yet), so "supported" might be too strong a word just yet.  Good to know that it works, though.

katiehockman:
CC @golang/release
Requesting a freeze exception to support coverage instrumentation with FreeBSD. This will only affect fuzzing, not the default behavior of go test. https://go-review.googlesource.com/c/go/+/325209/ is the CL (it needs to be cherrypicked to master)

bcmills:
(This is a release-blocker in the sense that it needs a decision on the freeze exception. If the freeze exception is not granted, we can move it to the Go1.19 milestone.)

heschi:
Approved -- this is a small change to a new feature with low risk.

PR #60520, opened 2023-05-30
cmd/go: enable fuzz testing on OpenBSD

### LINKED ISSUE #60491: cmd/go: enable fuzz testing on OpenBSD
On current version of OpenBSD you cannot run fuzz tests:

<pre>
$ go test -fuzz .
-fuzz flag is not supported on openbsd/amd64
</pre>

It looks like artificial restriction (comment by @dr2chase in https://github.com/golang/go/issues/46554#issuecomment-854930140):

> I don't think it was tested across all the other OSes (yet), so "supported" might be too strong a word just yet.  Good to know that it works, though.

After extending `platform.FuzzSupported(goos, goarch string)` and build tags inside `src/internal/fuzz/` directory, I can run `all.bash` with success and use newly build executable to run fuzz tests.

I've read contribution rules, but I'm not sure what should I do next. Is it enough to create pull request with this change?

<pre>
$ go version
go version go1.20.1 openbsd/amd64
</pre>

I suppose

macie:
This change provides support for -fuzz flag on OpenBSD. According to #46554 the flag was unsupported on some OSes due to lack of proper testing.

Fixes: #60491

### DIFF
--- a/src/internal/fuzz/counters_supported.go
+++ b/src/internal/fuzz/counters_supported.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-//go:build (darwin || linux || windows || freebsd) && (amd64 || arm64)
+//go:build (darwin || linux || windows || freebsd || openbsd) && (amd64 || arm64)
 
 package fuzz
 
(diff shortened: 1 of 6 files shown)


seankhliao:
looks like the cl was merged

PR #60671, opened 2023-06-07
x509: fix typo in SetFallbackRoots documentation

edigaryev:
Add a missing space between the environment variables, see https://github.com/golang/go/issues/43958#issuecomment-1317868626 for more details.


edigaryev:
My bad: I've mistakenly thought that `x509usefallbackroots` is separate environment variable.

PR #60672, opened 2023-06-07
log/slog: fix the wrapping anti-example

Humm42:
The new logger parameter is the same as the one in the example
"wrapping".

Fixes: f67b0a73e30d ("log/slog: initial commit")


Humm42:
I humbly decline signing the CLA for this change.

PR #60677, opened 2023-06-08
database/sql: add Null[T]

methane:
Generic version of NullString, NullInt64, etc.

Fixes #60370


Nasfame:
@methane Just a thought - can't we replace all the null types implementations with the generic like 

`type NullBool Null[bool]`

methane:
> @methane Just a thought - can't we replace all the null types implementations with the generic like

We can't. `NullBool` has a `.Bool` member.

PR #60778, opened 2023-06-14
internal/trace: reduce if statement

callthingsoff:
Since "if ev.Type == EvGoSysExit" must be executed,
use if-elsif to reduce the other if statement in the case of "ev.Type == EvGoSysExit"
when assigning to ev.P.


domdom82 (approved):

domdom82:
w00ps. I approved this by accident. How can I remove my approval?

PR #60866, opened 2023-06-19
os/exec: Use the built-in function min instead of minInt

apocelipes:
The built-in function `min` has been implemented and can now be used to replace some manually written `minType` helper functions.


apocelipes:
THank you for code reviewing. I've rebased to the latest HEAD.

PR #60929, opened 2023-06-21
slices: add Concat

earthboundkid:
Fixes #56353

### DIFF
--- a/src/slices/slices.go
+++ b/src/slices/slices.go
@@ -503,7 +503,7 @@ func Concat[S ~[]E, E any](slices ...S) S {
 			panic("len out of range")
 		}
 	}
-	newslice := make(S, 0, size)
+	newslice := Grow[S](nil, size)
 	for _, s := range slices {
 		newslice = append(newslice, s...)
 	}
--- a/src/slices/slices.go
+++ b/src/slices/slices.go
@@ -493,3 +493,19 @@ func Reverse[S ~[]E, E any](s S) {
 		s[i], s[j] = s[j], s[i]
 	}
 }
+
+// Concat returns a new slice concatenating the passed in slices.
+func Concat[S ~[]E, E any](slices ...S) S {
+	size := 0
+	for _, s := range slices {
+		size += len(s)
+		if size < 0 {
+			panic("len out of range")
+		}
+	}
+	newslice := make(S, 0, size)
+	for _, s := range slices {
+		newslice = append(newslice, s...)
+	}
+	return newslice
+}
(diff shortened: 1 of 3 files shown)


src/slices/slices.go
```diff
@@ -497,3 +497,19 @@ func Reverse[S ~[]E, E any](s S) {
 		s[i], s[j] = s[j], s[i]
 	}
 }
+
+// Concat returns a new slice concatenating the passed in slices.
+func Concat[S ~[]E, E any](slices ...S) S {
```
6543, line 502:
```suggestion
func Concat[S ~[]E, E any](slices ...S) E {
```

?

6543:
my implementation uses copy instead of append
```
func MergeSlices[T any](slices ...[]T) []T {
	sl := 0
	for i := range slices {
		sl += len(slices[i])
	}
	result := make([]T, sl)
	cp := 0
	for _, s := range slices {
		if sLen := len(s); sLen != 0 {
			copy(result[cp:], s)
			cp += sLen
		}
	}
	return result
}
```

this should result in no new mem allocs if Im right

6543:
should I send new pull or work on this?

6543:
as there was no work done since ... I submitted my own impl. adjusted and simplified to meet the criteria ->  #61817

carlmjohnson:
AFAIK, the CL is just on hold because Go 1.21 isn't out quite yet. It should be reviewed and merged soon, then you can do a CL to add new tests or benchmarks.

6543:
ah that's why you did not work on it  - sorry

@carlmjohnson should I add you as co-author to my patch or do you update yours ?!?

carlmjohnson:
I don't see a reason to update to match yours without a benchmark that shows that it's faster. Write a benchmark and then open a new CL that stacks on top of mine.

6543:
well this need a patch anyway ... as your pull has this api  
`func Concat[S ~[]E, E any](slices ...S) S {}`

and  proposed is  
`func Concat[S ~[]E, E any](dst S, slices ...S) S {`

not sure what has to be done when an accepted proposal has to be adjusted ?!?

6543:
PS: I'll write some benchmarks :)

carlmjohnson:
> well this need a patch anyway ... as your pull has this api `func Concat[S ~[]E, E any](slices ...S) S {}`

Read the full discussion. This was the accepted API. The dst slice turned out to be unworkable because of the aliasing problem.

carlmjohnson:
@6543 slices.Concat has been added. Feel free to open a performance boost CL. Those don't need to go through the proposal process. You just need to demonstrate that it's faster, and it can be merged.

PR #60960, opened 2023-06-23
all: fix typos

callthingsoff:


src/runtime/race/testdata/chan_test.go
```diff
@@ -766,7 +766,7 @@ func TestNoRaceCloseHappensBeforeRead(t *testing.T) {
 
 // Test that we call the proper race detector function when c.elemsize==0.
 // See https://github.com/golang/go/issues/42598
-func TestNoRaceElemetSize0(t *testing.T) {
+func TestNoRaceElemSize0(t *testing.T) {
```
hirentimbadiya, line 769: Shouldn't be this TestNoRaceElementSize0  ??

hirentimbadiya (commented):
in chan_test.go file Shouldn't be that TestNoRaceElementSize0 ??

PR #61132, opened 2023-07-01
cmd/internal/goobj: optimize funcs - mkbuiltin and enumerateBasicTypes

callthingsoff:
Declare "pkg" as a const to reduce runtime.concat calls.
Prealloc "result" slice with cap "len(names)*2" for less allocation and faster execution.


coditamar:
@CodiumAI-Agent please review

coditamar:
@CodiumAI-Agent please explain the change. how does this change improve performance?

PR #61143, opened 2023-07-02
slices: in benchmarks remove call to rand.Seed

suyashtava:
Fixes #61142.


suyashtava:
Solving Issue: https://github.com/golang/go/issues/61142

PR #61166, opened 2023-07-05
[perf][queue] Replace %s with copy.

ttsugriy:
This avoids relatively expensive `%` operations and `copy` is much easier to optimize.

I've used https://gist.github.com/ttsugriy/45db22b2564745bedda05574c033c0b0 to get benchmark results below:
```
% go test -benchmem -run=^$ -bench ^Benchmark bench.com/b
goos: darwin
goarch: arm64
pkg: bench.com/b
BenchmarkModClone-8       903997              1123 ns/op            3200 B/op         20 allocs/op
BenchmarkCopyClone-8     1337074               890.6 ns/op          3200 B/op         20 allocs/op
PASS
ok      bench.com/b     3.266s
```

the Gerrit change.

  + No trailing period
+ No Markdown

### DIFF
--- a/src/internal/fuzz/queue.go
+++ b/src/internal/fuzz/queue.go
@@ -32,9 +32,9 @@ func (q *queue) grow() {
 	}
 	newElems := make([]any, newCap)
 	oldLen := q.len
-	for i := 0; i < oldLen; i++ {
-		newElems[i] = q.elems[(q.head+i)%oldCap]
-	}
+	copy(newElems, q.elems[q.head:])
+	wrote := oldLen - q.head
+	copy(newElems[wrote:], q.elems[:q.head])
 	q.elems = newElems
 	q.head = 0
 }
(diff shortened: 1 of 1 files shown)


shcabin:
If grow() is called when len<cap, the result will be data loss?
`	qq.enqueue(1)
	qq.enqueue(2)
	qq.dequeue()
	qq.grow()`
elem 2 will become nil

ttsugriy:
> If grow() is called when len<cap, the result will be data loss? example：
> 
> > qq := queue{}
> > qq.enqueue(1)
> > qq.enqueue(2)
> > qq.dequeue()
> > qq.grow()
> 
> elem 2 will become nil

You're correct, but currently `grow` is called only when existing storage is completely full, so `len ==cap` is a precondition.

ttsugriy:
@shcabin I've updated diff to handle the use-case you've described. Proof - https://go.dev/play/p/i9hhp89vAdP

PR #61170, opened 2023-07-05
cmd/go: chore: minor changes for performance and readability

Nasfame:
style: performance and readability


src/cmd/go/internal/work/exec.go
```diff
@@ -141,7 +141,7 @@ func (b *Builder) Do(ctx context.Context, root *Action) {
 			// TODO(matloob): Better action descriptions
 			desc := "Executing action "
 			if a.Package != nil {
-				desc += "(" + a.Mode + " " + a.Package.Desc() + ")"
```
dongzhiwei-git, line 144: "+" is better than fmt.Sprintf when dealing with fewer strings

Issue #61361, opened 2023-07-14
proposal: Go 2: Use `C.size_t` in `C.GoStringN()` and `C.GoBytes()`


bemasc:
### Proposal

Change the signature of `GoStringN(*C.char, C.int)` to `GoStringN(*C.char, C.size_t)`, and similarly for `GoBytes`.  This would enable C->Go transfer of >2GB strings on most Go systems, and avoid a cast when wrapping C APIs that follow modern usage guidelines.

This would also improve consistency with `_GoStringLen()`, which already returns `size_t`.

### Costs

All existing uses of `GoStringN()` would require a change (either removing or adding a typecast).

ianlancetaylor:
That proposed change would break the spirit of the Go 1 compatibility guarantee, so we're not going to do it.  (The Go 1 compatibility guarantee doesn't cover cgo, but, we're still not going to gratuitously break existing code without a really good reason.)  And though the issue is marked "Go 2", there isn't going to be a Go 2, and we aren't going to make a cgov2 for something like this.

So I suggest a different proposal:

```Go
// GoStringBigN is like GoStringN but for very large strings.
func GoStringBigN(*C.char, C.size_t)

// GoBytesBig is like GoBytes but for very large data.
func GoBytesBig(unsafe.Pointer, C.size_t)
```

bemasc:
That would be fine with me.  Alternatively, perhaps it would be possible to widen the argument types to accept both (e.g., `func GoStringN[T C.int | C.size_t](*C.char, T)` in generics style).

bcmills:
I think we already provide too many `cgo` helper functions, and I would prefer that we deprecate them rather than expanding them.

In particular, these functions seem redundant with the recently-added `unsafe.String`, `unsafe.Slice`, and `bytes.Clone`. Consider:
```go
func goStringN(p *C.char, n C.size_t) string {
	return string(unsafe.Slice((*byte)(unsafe.Pointer(p)), n))
}

func goBytes(p unsafe.Pointer, n C.size_t) []byte {
	return bytes.Clone(unsafe.Slice((*byte)(p), n))
}
```

rsc:
This proposal has been **[declined as infeasible](https://go.dev/s/proposal-status#declined-as-infeasible)**.
— rsc for the proposal review group

PR #61366, opened 2023-07-14
Remove "Go 2" labels from the template

bemasc:
Go is apparently no longer soliciting "Go 2" language proposals.

See discussion at https://github.com/golang/go/issues/61361#issuecomment-1636399995.

### DIFF
--- a/.github/ISSUE_TEMPLATE/11-language-change.md
+++ b/.github/ISSUE_TEMPLATE/11-language-change.md
@@ -1,8 +1,8 @@
 ---
 name: Language Change Proposals
 about: Changes to the language
-title: "proposal: Go 2: "
-labels: Proposal Go2 LanguageChange
+title: "language change proposal: "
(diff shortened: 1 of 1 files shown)


Jorropo:
I think and also I think this is fairly spread understanding that Go2 means:

> I think this would be a cool change but I don't know how to make it non breaking, please provide input on how to make it non breaking.

If you bring a huge point (like loopvar, basically pretty much just loopvar), back it up with strong evidance (for example try to run hundred of millions of lines of go in tests with your change and study the breakage, showing they were incorrect to begin with, ...), ... you would have a strong argument for a breaking proposal being added as a v2 (or using the new `go.mod` semantic update proposed with loopvar), but except than thoses extremely rare cases it's pretty much an RFC for making it non breaking (which is what was proposed in https://github.com/golang/go/issues/61361#issuecomment-1636399995).

PR #61377, opened 2023-07-16
net/http: pool transport gzip readers

### LINKED ISSUE #61353: Reuse gzip reader from pool
https://github.com/golang/go/blob/089e37a931d30d4055c7468facb602c4cfa9b537/src/net/http/transport.go#L2858
Is it possible to reuse this gzip reader from a `sync.Pool` to reduce the memory allocation cause by `flate.NewReader`

AlexanderYastrebov:
name              old time/op    new time/op    delta
GzipReaderInit-8    9.15µs ± 4%    2.38µs ± 9%  -74.04%  (p=0.000 n=10+10)

name              old alloc/op   new alloc/op   delta
GzipReaderInit-8    45.5kB ± 0%     4.3kB ± 0%  -90.47%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
GzipReaderInit-8      10.0 ± 0%       5.0 ± 0%  -50.00%  (p=0.000 n=10+10)

Allocation saving comes from absent compress/flate.(*dictDecoder).init

Fixes #61353

### DIFF
--- a/src/net/http/transport.go
+++ b/src/net/http/transport.go
@@ -11,6 +11,7 @@ package http
 
 import (
 	"bufio"
+	"compress/flate"
 	"compress/gzip"
 	"container/list"
 	"context"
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
Concurrent Read and Close does not look right and we can not call Reset while body mutex is held because it calls body Read that ends up in deadlock.

PR #61440, opened 2023-07-19
reflect: deprecate PtrTo

qiulaidongfeng:
Replace reflect.PtrTo with reflect.PointerTo.

Fixes #59599

Change-Id: I49407193e2050543ef983cd637703acc682d9f51

### DIFF
--- a/src/reflect/type.go
+++ b/src/reflect/type.go
@@ -1171,6 +1171,8 @@ var ptrMap sync.Map // map[*rtype]*ptrType
 //
 // PtrTo is the old spelling of PointerTo.
 // The two functions behave identically.
+//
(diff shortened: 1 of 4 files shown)


src/reflect/type.go
```diff
@@ -1171,6 +1171,7 @@ var ptrMap sync.Map // map[*rtype]*ptrType
 //
 // PtrTo is the old spelling of PointerTo.
 // The two functions behave identically.
+// Deprecated: Superseded by [PointerTo]
```
thaJeztah, line 1174: I think there must be an empty comment line before the Deprecated line, otherwise it may not be picked up as "deprecated"

PR #61547, opened 2023-07-24
encoding/base32: reduce overflow risk when calc encode/decode len

chanxuehong:
Same as https://go-review.googlesource.com/c/go/+/510635, reduces risk of overflow


fengzie:
老产，我是杨明峰，怎么能联系到你？给你gmail发邮件了。

PR #61601, opened 2023-07-26
archive/zip: ignore modified time if extended timestamp field exists

### LINKED ISSUE #61572: archive/zip: extra Extra fields when repacking zip file
If you use zip.(\*Writer).CreateHeader (for example using a header from a zip file you are copying), CreateHeader encodes fh.Modified into fh.Extra even if fh.Extra already contains a modified time. You get an Extra with an extra extTimeExtraID entry. If there is already a extTimeExtraID entry in Extra, we shold not add an extra extTimeExtraID.

AlexanderYastrebov:
Fixes #61572


AlexanderYastrebov:
See https://go.dev/cl/513836

Issue #54852, opened 2022-09-03
cmd/dist: no way to disable buildvcs stamping during bootstrap


williamh:
Hi,

I am the packager for Go on Gentoo linux.

We have an on-going issue with the VCS stamping which has appeared in another way.

Since go 1.17 and older are not in our tree any longer, we are able to add "-buildvcs=false" to our default build flags for go-based packages to prevent vcs stamping from breaking our package builds.

However, if go is being bootstrapped and the bootstrap binaries are newer than version 1.18, it is possible for vcs stamping to crash the build as shown below.

### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.18.5 linux/amd64

</pre>

### Does this issue reproduce with the latest release?

yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/william/.cache/go-build"
GOENV="/home/william/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/william/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/william/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/lib/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.18.5"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="x86_64-pc-linux-gnu-gcc"
CXX="x86_64-pc-linux-gnu-g++"
CGO_ENABLED="1"
GOMOD="/home/william/build/go/src/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
[...]

PR #61621, opened 2023-07-27
cmd/dist: set buildvcs=false when building go-bootstrap

### LINKED ISSUE #61620: cmd/dist: set buildvcs=false if compiling outside of a Git directory
Related: #54852 - cc @williamh @bcmills

In Buildroot we have a build failure due to not having the .git directory in the Go sources: https://gitlab.com/buildroot.org/buildroot/-/jobs/4725186525

```
Building Go toolchain1 using /builds/buildroot.org/buildroot/test-output/TestMender/host/lib/go-1.19.10.
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
go tool dist: FAILED: /builds/buildroot.org/buildroot/test-output/TestMender/host/lib/go-1.19.10/bin/go install -tags=math_big_pure_go compiler_bootstrap purego bootstrap/cmd/...: exit status 1
```

It seems that -buildvcs=false is not set in the "go install" command regardless of isGitRepo():

https://github.com/golang/go/blob/go1.20.6/src/cmd/dist/buildtool.go#L205

Applying two patches fixes the issue within Buildroot:

 - Set -buildvcs=false in the "go install" command if not `isGitDir()`: https://github.com/paralin/go/commit/1e0a65f424fe717513b2c9f7edc8a5bde643da43
 - Add an environment variable GOBOOTSTRAP_BUILDVCS to force `isGitDir()` to return false: https://github.com/paralin/go/commit/7e65e8e8b9061cffead6feb8a471a2716ebef59d

This worked fine with go1.19.x but regressed with go1.20.x.

paralin:
When building go-bootstrap as part of the make.bash process, the cmd/dist invokes the bootstrap Go compiler to build the go_bootstrap tool:

${GOROOT_BOOTSTRAP}/bin/go install -tags=math_big_pure_go compiler_bootstrap purego bootstrap/cmd/...

If there is an invalid .git directory in a parent of ${GOROOT_BOOTSTRAP}, make.bash will fail. Reproduction of the issue:

  mkdir go-issue-61620
  cd ./go-issue-61620
  wget https://go.dev/dl/go1.19.11.src.tar.gz
  mkdir go-bootstrap
  tar -xf go1.19.11.src.tar.gz -C ./go-bootstrap --strip-components=1
  cd ./go-bootstrap/src/
  bash make.bash
  cd ../../
  wget https://go.dev/dl/go1.20.6.src.tar.gz
  mkdir go
  tar -xf go1.20.6.src.tar.gz -C ./go/ --strip-components=1
  printf "gitdir: ../../does/not/exist/.git" > ./.git
  cd ./go/src/
  GOROOT_BOOTSTRAP=$(pwd)/../../go-bootstrap/ bash make.bash

The build fails with the following error:

  Building Go toolchain1 using [snip]/go-1.19.10.
  error obtaining VCS status: exit status 128
    Use -buildvcs=false to disable VCS stamping.
  go tool dist: FAILED: [snip]/go-1.19.10/bin/go install -tags=math_big_pure_go \
    compiler_bootstrap purego bootstrap/cmd/...: exit status 1

This change unconditionally sets -buildvcs=false when compiling go-bootstrap. We don't need the revision information in those binaries anyway. Setting this flag was previously not done as we were unsure if the go-bootstrap compiler would be new enough to support the buildvcs build flag. However, since Go 1.20.x, Go 1.19.x is the minimum version for go-bootstrap, and supports -buildvcs=false. We can now set -buildvcs=false without worrying about compatibility.

Related: #54852
Fixes: #61620

### DIFF
--- a/src/cmd/dist/buildtool.go
+++ b/src/cmd/dist/buildtool.go
@@ -244,6 +244,10 @@ func bootstrapBuildTools() {
 	cmd := []string{
 		pathf("%s/bin/go", goroot_bootstrap),
 		"install",
+		// In case we are bootstrapping from a source tarball instead of a VCS checkout,
+		// ignore any VCS roots found in parent directories of the bootstrap module.
+		// See https://go.dev/issue/54852 and https://go.dev/issue/61620.
+		"-buildvcs=false",
 		"-tags=math_big_pure_go compiler_bootstrap purego",
 	}
 	if vflag > 0 {
--- a/src/cmd/go/internal/load/pkg.go
+++ b/src/cmd/go/internal/load/pkg.go
@@ -2510,12 +2510,6 @@ func (p *Package) setBuildInfo(ctx context.Context, f *modfetch.Fetcher, autoVCS
 	}
 
 	if wantVCS && p.Module != nil && p.Module.Version == "" && !p.Standard {
-		if p.Module.Path == "bootstrap" && cfg.GOROOT == os.Getenv("GOROOT_BOOTSTRAP") {
-			// During bootstrapping, the bootstrap toolchain is built in module
-			// "bootstrap" (instead of "std"), with GOROOT set to GOROOT_BOOTSTRAP
-			// (so the bootstrap toolchain packages don't even appear to be in GOROOT).
-			goto omitVCS
-		}
 		repoDir, vcsCmd, err = vcs.FromDir(base.Cwd(), "")
 		if err != nil && !errors.Is(err, os.ErrNotExist) {
 			setVCSError(err)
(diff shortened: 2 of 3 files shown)


paralin:
@seankhliao this is still applicable, can we reopen?

seankhliao:
yes, sorry I looked at the wrong linked issue

paralin:
hi @seankhliao this is still needed and we have to copy this patch into every version in buildroot.org source tree, it would be great if it could be merged. thanks!

PR #61623, opened 2023-07-28
debug/elf: uncompress .zdebug sections in Open

mauri870:
The method (*Section).Open automatically decompresses section contents when they are marked with SHF_COMPRESSED. This is the recommended approach in the ELF standard. Some tools use the older mechanism of prefixing the sections with a z as in .zdebug_str for example, this case was unhandled.

The new implementation will look for SHF_COMPRESSED as well as sniffing for ZLIB magic bytes and for sections starting with .z.

Fixes #58254


mauri870:
Closing since this was handled by https://go.dev/cl/480895

PR #61695, opened 2023-08-01
net: use the extended RCode from EDNS(0) OPT resources

mateusz834:
For a while now we support EDNS, but the current
implementation only sends the OPT resource and doesn't
do anything with the response OPT resource.

For reference the miekg/dns updates the RCode in the
header when there is a OPT resource:
https://github.com/miekg/dns/blob/48f38ebef989eedc6b57f1869ae849ccc8f5fe29/msg.go#L868-L872


rolandshoemaker:
@gopherbot please open backport issues, this is a PUBLIC track security issue.

PR #61971, opened 2023-08-12
syscall: update windows zsyscalls to use SyscallN

mauri870:
This is a follow up for CL 518995 that regenerates the
windows zsyscalls to use SyscallN instead of the various
SyscallX functions.

Fixes #57914


mauri870:
Marking as abandoned since there is a solution where we don't need to generate the syscalls again, it only uses SyscallN for syscalls with more than 15 params.

PR #62499, opened 2023-09-07
net/http: add partitioned attribute to cookie type

islishude:
Fixes #62490

### DIFF
--- a/src/net/http/cookie.go
+++ b/src/net/http/cookie.go
@@ -319,8 +319,8 @@ func (c *Cookie) Valid() error {
 		}
 	}
 	if c.Partitioned {
-		if !c.Secure || c.Path != "/" {
-			return errors.New("http: partitioned cookies must be set with Secure and Path=/")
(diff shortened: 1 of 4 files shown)


src/net/http/cookie.go
```diff
@@ -267,6 +274,11 @@ func (c *Cookie) Valid() error {
 			return errors.New("http: invalid Cookie.Domain")
 		}
 	}
+	if c.Partitioned {
+		if !c.Secure || c.Path != "/" {
+			return errors.New("http: partitioned cookies must be set with Secure and Path=/")
```
Haswf, line 279: From my understanding, partitioned cookie's path doesn't have to be `/`. Only __Host prefixed cookie must be both secure and path=/.

> Partitioned cookies must be set with Secure.
It is recommended to use the __Host prefix when setting partitioned cookies to make them bound to the hostname (and not the registrable domain).
https://developers.google.com/privacy-sandbox/3pcd/chips

Issue #62257, opened 2023-08-24
cmd/compile: special-case MethodByName(string literal) to keep the DCE enabled.


dominiquelefevre:
Normally, a call to `MethodByName()` disables the DCE because the linker assumes that any method can be accessed this way. This pessimises the code generation for `k8s.io/apimachinery` which needs `MethodByName()` to verify whether or not a struct implements `DeepCopyInto()`. It cannot cast a struct to `interface { DeepCopyInto() Foo }` because the return type may vary. Instead, it does the following:
```
      if m := reflect.ValueOf(obj).MethodByName("DeepCopyInto"); ... {
```
In this case there is no need to disable the DCE altogether. It suffices to add a relocation to keep methods named `DeepCopyInto()`.

thanm:
Thanks for sending this.  I touched up the reviewer set for your CLs and ran the trybots.

PR #62681, opened 2023-09-16
cmd/link: add debug log to identify deadcode blockers

jquirke:
In deadcode removal in the linker, any use of 
reflect.Type.Method/MethodByName result in an increased amount of 
incorrectly assumed reachable code  (specifically methods), due to the
 intractability of knowing what methods might be used.

In https://github.com/golang/go/issues/62257 there was some compiler assisted improvement added to special
case MethodByName and Method with constant arguments, and avoid marking
those as AttrReflectMethod.

At Uber this optimisation has resulted in smaller binary sizes when
the the usage of reflect.Type.Method/ByName had constant arguments.

However, there are still a number of less trivial cases which we would
like to easily identify in our codebase, and logging would be helpful
here.


jquirke:
Closing as won't do: @cherrymui pointed out there is already a log that is useful for detecting this.

```
I'll note that if you use -ldflags=-dumpdep linker flag, it will print reachability edges and symbols marked as reflect method are noted with "<ReflectMethod>". Does that help? Would the logging still be needed?

```

PR #63231, opened 2023-09-26
x/sys: fix trimmed socket opt string in unix/syscall_linux

### LINKED ISSUE #63217: x/sys: trimmed string socket opt on linux
# Bug description

Func `GetsockoptString` from x/sys/unix package always removes the last byte from the buffer, which in some cases leads to a truncated option value. 

From `unix` [man page](https://man7.org/linux/man-pages/man7/unix.7.html):

```
              The security context string may include a terminating null
              character in the returned length, but is not guaranteed to
              do so: a security context "foo" might be represented as
              either {'f','o','o'} of length 3 or {'f','o','o','\0'} of
              length 4, which are considered to be interchangeable.  The
              string is printable, does not contain non-terminating null
              characters, and is in an unspecified encoding (in
              particular, it is not guaranteed to be ASCII or UTF-8).
```

But the implementation ([this patch](https://go-review.googlesource.com/c/sys/+/83275/4/unix/syscall_linux.go#813)) always cuts the last byte without checking it for the null character.

<pre>
$ go version
go version go1.21.1
</pre>

Yes.

I'm using intel mac, but building for arm linux (GOARCH=arm64 GOOS=linux)
$ go env
go env
GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/Users/kostik/Library/Caches/go-build'
GOENV='/Users/kostik/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/kostik/go/pkg/mod'
GOOS='darwin'
GOPATH='/Users/kostik/go'
GOROOT='/usr/local/Cellar/go/1.21.1/libexec'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/Cellar/go/1.21.1/libexec/pkg/tool/darwin_amd64'
GOVCS=''
GOVERSION='go1.21.1'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='cc'
CXX='c++'
CGO_ENABLED='1'
GOMOD='/dev/null'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/3p/1hh1wbcx7hjgl_79qxfdbqtr0000gq/T/go-build2506492106=/tmp/go-build -gno-record-gcc-switches -fno-common'
</pre></details>

### What did you do?

The service code is running on the linux with **apparmor enabled**. It just reads the app armor label from the Unix socket connection by calling:
```go
var getSocketSecurityLabel = func(rc syscall.RawConn) (string, error) {
	var label string
	var readLabelError error
	err := rc.Control(func(fd uintptr) {
		// Read the apparmor label from the socket option SO_PEERSEC
		label, readLabelError = unix.GetsockoptString(int(fd), syscall.SOL_SOCKET, unix.SO_PEERSEC)
	})

	if err != nil {
		return "", err
	}
	if readLabelError != nil {
		return "", readLabelError
	}

	return label, nil
}
```

* For unconfined application the expected output is `unconfined`
* For snap running in development mode the expected output is `appname (complain)`
* For a labeled app the expected output is full app name

### What did you see instead?

* For unconfined application the real output is `unconfine`
* For snap running in development mode the real output is `appname (complain`
* For a labeled app the real output miss the last char of the name

KSDaemon:
Take null-byte into account and trim only if needed. Fixes #63217

### DIFF
--- a/src/cmd/vendor/golang.org/x/sys/unix/syscall_linux.go
+++ b/src/cmd/vendor/golang.org/x/sys/unix/syscall_linux.go
@@ -1310,7 +1310,11 @@ func GetsockoptString(fd, level, opt int) (string, error) {
 			return "", err
 		}
 	}
-	return string(buf[:vallen-1]), nil
+	if buf[vallen-1] == 0 {
+		return string(buf[:vallen-1]), nil
+	} else {
+		return string(buf[:vallen]), nil
+	}
 }
 
 func GetsockoptTpacketStats(fd, level, opt int) (*TpacketStats, error) {


KSDaemon:
Hi there! If you can point me and advice how to write a tests for this - I would love to do. But it seems to be hard as it involves mocking low-level syscall things...

Issue #61725, opened 2023-08-02
cmd/compile: intrinsify strings.Compare


dsnet:
A [comment by @rsc in `strings.Compare` says](https://github.com/golang/go/blob/ed8cbaf6ac58120c7171dadaaa091f031712c6d2/src/strings/compare.go#L14-L20):
> This function does NOT call the runtime cmpstring function,
> because we do not want to provide any performance justification for
> using strings.Compare. Basically no one should use strings.Compare.
> As the comment above says, it is here only for symmetry with package bytes.
> If performance is important, the compiler should be changed to recognize
> the pattern so that all code doing three-way comparisons, not just code
> using strings.Compare, can benefit.

I'm not sure this comment has aged well with the recent change to `slices.SortFunc`,
where it now takes in an `func(T, T) int` instead of `func(T, T) bool`.
This means that the utility of `strings.Compare` suddenly shot up.

We should either intrinsify `strings.Compare` as `runtime.cmpstring` OR (even better) recognize this pattern.
That way, `cmp.Compare` on `string` types will benefit as well.

go101:
https://github.com/golang/go/issues/50167, https://github.com/golang/go/issues/58142, https://go101.org/blog/2022-10-01-three-way-string-comparison.html

dr2chase:
@golang/compiler

yuryu:
Can we bring back this removed optimization? I'm hitting the same issue. `slices.SortFunc` needs a three-way comparison and there's no good way to do it right now.

https://go-review.googlesource.com/c/go/+/3012

ncruces:
Another vote for something that recognizes `cmp.Compare` and/or sequences of `cmp.Less`.

They seem to have been written in such a way that the optimizer can throw away all the `NaN` handling, yay! 🎉

But `cmp.Compare` gets particularly bad for strings: it doesn't even prioritize `==`, like `strings.Compare` does.

Or should we also document `cmp.Compare` to say no one should call it? 🫤

ncruces:
Good. Yet, I'm a bit sad that this means giving up on `Ordered` having a decent implementation that's any good for strings.

It's a bit stunning to, for years, take the stance that you don't really need three-way compares, then build an entire 2nd search/sort library around three-way compares, and then fail to optimize the handful `Ordered` types for it, in effect pushing non-generic callbacks over operators for performance.

There _are_ cases where three-way compares are _not_ something to avoid, and leaving `Ordered` out of those is… sad.

[...]

Issue #31187, opened 2019-04-01
cmd/compile: Optimize three-way string comparisons


ashi009:
As shown in `strings.Compare`'s [source code](https://golang.org/src/strings/compare.go#L13), three-way string comparisons can be optimized.

A few use cases are:

- sorted slice merging
- sorted slice intersection
- sorted slice intersection detection

For the last use case my [benchmark](https://play.golang.org/p/VoMTfd39MA3) showed a over 10% improvement by using `bytes.Compare` instaed of `strings.Compare`.

agnivade:
@rsc

randall77:
Currently we break string comparison ops into runtime calls during walk.
If we delayed that lowering until later in SSA, we could maybe detect the use of both `s1 == s2` and `s1 < s2` and use a single call in that case.

josharian:
Perhaps the implementation should be moved to internal/bytealg.

josharian:
This would be another possible use for an rtcall SSA pass: #24926

But it’d be better indeed to leave these unlowered until much later. Which would require making it reasonable to manipulate and insert calls in SSA. Which is yet another reason to move parameter and result handling into ssa ops instead of SP writes and reads. (I’m pretty sure I’ve written this paragraph a dozen times. I should just go do it.)

dsymonds:
This would also be nice so that `strings.Compare` can have that rsc comment removed entirely, since it can be clearer to use it instead of the builtin operators in some cases, such as building a higher level comparison func.

```go
type key struct {
  f1, f2, f3 string
}

func (k key) Compare(o key) int {
  if c := strings.Compare(k.f1, o.f1); c != 0 {
    return c
  }
  if c := strings.Compare(k.f2, o.f2); c != 0 {
    return c
  }
  if c := strings.Compare(k.f3, o.f3); c != 0 {
    return c
  }
  return 0
}
```

go101:
It is done for https://github.com/golang/go/issues/61725

ianlancetaylor:
Technically this issue isn't done, but since people can now call `strings.Compare`, close enough.

PR #63360, opened 2023-10-03
cmp: optimize Compare

ncruces:
Compare provides three-way comparison for Ordered types, which certain
algorithms benefit from.

Float comparisons require NaN handling that should be optimized away for
all other types.
String comparisons benefit from using != which short-circuits on
different length strings.
Ideally the compiler would recognize the pattern and specialize on
strings, but this tries to improve the status quo.

goos: darwin
goarch: amd64
pkg: cmp
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
                   │  base.txt   │              patch.txt              │
                   │   sec/op    │   sec/op     vs base                │
Compare_int-12       1.638n ± 1%   1.635n ± 2%        ~ (p=0.698 n=10)
Compare_float64-12   2.397n ± 1%   2.206n ± 0%   -7.99% (p=0.000 n=10)
Compare_strings-12   8.054n ± 1%   6.705n ± 2%  -16.75% (p=0.000 n=10)
geomean              3.162n        2.892n        -8.56%

Updates #31187
Updates #61725

### DIFF
--- a/src/cmp/cmp.go
+++ b/src/cmp/cmp.go
@@ -38,18 +38,19 @@ func Less[T Ordered](x, y T) bool {
 // For floating-point types, a NaN is considered less than any non-NaN,
 // a NaN is considered equal to a NaN, and -0.0 is equal to 0.0.
 func Compare[T Ordered](x, y T) int {
-	xNaN := isNaN(x)
-	yNaN := isNaN(y)
-	if xNaN && yNaN {
+	if x == y {
 		return 0
 	}
-	if xNaN || x < y {
+	if x < y {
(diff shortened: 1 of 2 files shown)


src/cmp/cmp.go
xiaotushaoxia, line 1: may I know why "if x < y" placed above  “if isNaN(x)”， "if x < y" before, I need know what is the result of x<y if x or y is NaN.

ncruces:
https://go-review.googlesource.com/c/go/+/578835

PR #63386, opened 2023-10-05
sync: use atomic.Uint32 in Once

mstmdev:


uhthomas:
Sorry to bring up an old change, but could you help me understand why atomic.Uint32 would be used instead of atomic.Bool? Is there a difference? I noticed in the commit history that atomic.Bool was used, but then reverted due to struct field ordering.

Thanks in advance for your help!

@mstmdev @ianlancetaylor

ianlancetaylor:
@uhthomas Note that we use Gerrit for code review. Almost nobody sees comments on GitHub pull requests. Please see  https://go.dev/cl/532895 for the review, and comment there if appropriate. Or ask on a forum: see https://go.dev/wiki/Questions. Thanks.

PR #63519, opened 2023-10-12
encoding/xml: use reflect.Value.IsZero

apocelipes:
IsZero does the same thing, using this rather than writing it again.


andig:
@apocelipes do you plan to open the same PR for encoding/json?

apocelipes:
> @apocelipes do you plan to open the same PR for encoding/json?

Not yet. You can take it if you want. :)

Issue #63824, opened 2023-10-30
internal/zstd: Bugs and crashes 


klauspost:
### What version of Go are you using (`go version`)?

Go `tip`, self compiled.

### Does this issue reproduce with the latest release?

Yes.

### What did you do?

Run some fuzz tests against the internal zstd package.

### What did you expect to see?

No output.

#### A) Crash

With the following input, the internal/zstd decompressor can fail. To reproduce:

Add src/internal/zstd/testdata/fuzz/FuzzReader/e4f4dab13d374c2d with the following content:

go test fuzz v1
[]byte("(\xb5/\xfd00\xec\x00\x00&@\x05\x05A7002\x02\x00\x02\x00\x02\x0000000000000000")

This produces the following crash:

```
--- FAIL: FuzzReader (10.70s)
    --- FAIL: FuzzReader (0.00s)
        testing.go:1504: panic: runtime error: index out of range [2] with length 2
            goroutine 145795 [running]:
            runtime/debug.Stack()
                C:/go/src/runtime/debug/stack.go:24 +0x9b
            testing.tRunner.func1()
                C:/go/src/testing/testing.go:1504 +0x1ee
            panic({0x6952c0?, 0xc000130618?})
                C:/go/src/runtime/panic.go:914 +0x21f
            [github.com/klauspost/compress/z2.(*Reader).readLiteralsFourStreams(0xc001cb1e00](http://github.com/klauspost/compress/z2.(*Reader).readLiteralsFourStreams(0xc001cb1e00), {0xc000502d20, 0x1d, 0x20}, 0x9, 0xf, 0x2, {0x0, 0x0, 0x0})
                e:/gopath/src/[github.com/klauspost/compress/z2/literals.go:324](http://github.com/klauspost/compress/z2/literals.go:324) +0x1079
            [github.com/klauspost/compress/z2.(*Reader).readHuffLiterals(0xc001cb1e00](http://github.com/klauspost/compress/z2.(*Reader).readHuffLiterals(0xc001cb1e00), {0xc000502d20, 0x1d, 0x20}, 0x1, 0x26, {0x0, 0x0, 0x0})
                e:/gopath/src/[github.com/klauspost/compress/z2/literals.go:168](http://github.com/klauspost/compress/z2/literals.go:168) +0x11e7
[...]

PR #64056, opened 2023-11-10
internal/zstd: fix seek offset bounds check in skipFrame

aimuz:
This change enhances the zstd Reader's skipFrame function to validate
the new offset when skipping frames in a seekable stream, preventing
invalid offsets that could occur previously.

A set of "bad" test strings has been added to fuzz_test.go to extend
the robustness checks against potential decompression panics.

Additionally, a new test named TestReaderBad is introduced in 
zstd_test.go to verify proper error handling with corrupted input
strings.

The BenchmarkLarge function has also been refactored for clarity,
removing unnecessary timer stops and resets.

Updates #63824


aimuz:
(cc @klauspost)

PR #64155, opened 2023-11-15
internal/zstd: enforce block size limit according to RFC 8878

aimuz:
This change updates the zstd reader implementation to check for block 
size limits as specified by the zstd specification RFC 3.1.1.2.4. 
Now, it ensures that both decompressed and compressed sizes of any block 
do not exceed the frame's Block_Maximum_Size.

The previous logic only checked the buffer and literal buffer combined 
size against a hard-coded maximum which did not account for varying frame 
sizes or match lengths. This has been corrected to dynamically calculate 
the size based on actual literals and matches processed, ensuring 
compliance with the standard.

Updates #63824


aimuz:
wait #64056

PR #64326, opened 2023-11-22
case(syslog): added a case

flc1125:


flc1125:
https://github.com/go-packagist/go-kratos-components/pull/33

flc1125:
https://github.com/golang/go/issues/64298

PR #64441, opened 2023-11-29
net: fixes to dnsReadConfig in dnsconfig_windows.go

qdm12:
- Only search DNS servers for network interfaces with at least one gateway
- Clarify comment on deprecated site local anycast fec0/10 DNS IPv6 addresses
- Minor maintenance: skip not "up" interfaces earlier in outer loop

### DIFF
--- a/src/internal/syscall/windows/syscall_windows.go
+++ b/src/internal/syscall/windows/syscall_windows.go
@@ -116,26 +116,30 @@ type IpAdapterGatewayAddress struct {
 }
 
 type IpAdapterAddresses struct {
-	Length                uint32
-	IfIndex               uint32
-	Next                  *IpAdapterAddresses
-	AdapterName           *byte
-	FirstUnicastAddress   *IpAdapterUnicastAddress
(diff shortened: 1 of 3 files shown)


src/internal/syscall/windows/syscall_windows.go
```diff
@@ -98,6 +101,20 @@ type IpAdapterPrefix struct {
...
 }
 
+type IpAdapterWinsServerAddress struct {
+	Length   uint32
+	Reserved uint32
+	Next     *IpAdapterWinsServerAddress
+	Address  SocketAddress
+}
+
+type IpAdapterGatewayAddress struct {
```
qdm12, line 111: Let me know if I should add a documentation link for it, for example: https://learn.microsoft.com/en-us/windows/win32/api/iptypes/ns-iptypes-ip_adapter_gateway_address_lh
I did not to conform with existing structures not having such documentation link.

That would had helped me to understand the structures there, although I managed without initially.

PR #64728, opened 2023-12-15
database/sql: wrap errors with %w in driverArgsConnLocked

### LINKED ISSUE #64707: database/sql: use %w verb for valuer errors to enable errors.Is/As
### Go version

go version go1.21.4 darwin/arm64

```shell
GO111MODULE=''
GOARCH='arm64'
GOBIN=''
GOCACHE='/Users/gofer/Library/Caches/go-build'
GOENV='/Users/gofer/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='arm64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/gofer/go/pkg/mod'
GOOS='darwin'
GOPATH='/Users/gofer/go'
GOROOT='/opt/homebrew/Cellar/go/1.21.4/libexec'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/opt/homebrew/Cellar/go/1.21.4/libexec/pkg/tool/darwin_arm64'
GOVCS=''
GOVERSION='go1.21.4'
GCCGO='gccgo'
AR='ar'
CC='cc'
CXX='c++'
CGO_ENABLED='1'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/cf/314tnd5d22129h1z7srmcklr0000gn/T/go-build479648600=/tmp/go-build -gno-record-gcc-switches -fno-common'
```

### What did you do?

```go
package main

import (
	"context"
	"database/sql"
	"database/sql/driver"
	"errors"
)

type Valuer struct {
	field []byte
}

func (v Valuer) Value() (driver.Value, error) {
	if len(v.field) == 0 {
		// return an error that we hope to be able to handle later
		return nil, errEmpty
	}
	return v, nil
}

// errEmpty is returned when an empty value is found
var errEmpty = errors.New("empty value")

func main() {
	db, _ := sql.Open("pgx", dsn)
	_, err := db.ExecContext(
		context.Background(),
		`INSERT INTO some_table (some_bytea_column) VALUES ($1)`,
		&Valuer{nil},
	)
	// errors.Is doesn't work b/c %v verb was used for error instead of %w
	if !errors.Is(err, errEmpty) {
		panic(err)
	}
}
```

`errors.Is` should be able to detect sentinel errors returned by valuer.  error will need to be properly wrapped by fmt.Errorf with `%w` verb instead of `%v`, similar to [scanner](https://github.com/golang/go/blob/master/src/database/sql/sql.go#L3357)

### What did you see instead?

[valuer errors are wrapped by fmt.Errorf with %v verb](https://github.com/golang/go/blob/master/src/database/sql/convert.go#L195)

aimuz:
Use fmt.Errorf %w verb to wrap errors in driverArgsConnLocked,
which allows for easier unwrapping and checking of error types.

Add tests in sql_test.go to ensure that Stmt.Exec and Stmt.Query
correctly wrap underlying Valuer errors, adhering to the new change.

Fixes #64707.

### DIFF
--- a/src/database/sql/convert.go
+++ b/src/database/sql/convert.go
@@ -192,7 +192,7 @@ func driverArgsConnLocked(ci driver.Conn, ds *driverStmt, args []any) ([]driver.
 			}
 			goto nextCheck
 		default:
-			return nil, fmt.Errorf("sql: converting argument %s type: %v", describeNamedValue(nv), err)
+			return nil, fmt.Errorf("sql: converting argument %s type: %w", describeNamedValue(nv), err)
 		}
 	}
 
--- a/src/database/sql/driver/types.go
+++ b/src/database/sql/driver/types.go
@@ -34,6 +34,10 @@ type ValueConverter interface {
 
 // Valuer is the interface providing the Value method.
 //
+// Errors returned by the [Value] method are wrapped by the database/sql package.
+// This allows callers to use [errors.Is] for precise error handling after operations
+// like [database/sql.Query], [database/sql.Exec], or [database/sql.QueryRow].
+//
 // Types implementing Valuer interface are able to convert
 // themselves to a driver [Value].
 type Valuer interface {
--- a/src/database/sql/sql_test.go
+++ b/src/database/sql/sql_test.go
@@ -4398,6 +4398,49 @@ func TestRowsScanProperlyWrapsErrors(t *testing.T) {
 	}
 }
 
+type alwaysErrValuer struct{}
+
+// errEmpty is returned when an empty value is found
+var errEmpty = errors.New("empty value")
+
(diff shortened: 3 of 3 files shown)


joshgarrett-hellofresh (approved):
🥳

PR #65018, opened 2024-01-08
runtime: crash asap and extend total sleep time for slow machine in test

zzkcode:
Running with few threads usually does not need 500ms to crash, so let it
crash as soon as possible. While the test may caused more time on slow
machine, try to expand the sleep time in test.

Updates #64752

### DIFF
--- a/src/runtime/signal_unix.go
+++ b/src/runtime/signal_unix.go
@@ -599,9 +599,6 @@ func adjustSignalStack(sig uint32, mp *m, gsigStack *gsignalStack) bool {
 // GOTRACEBACK=crash when a signal is received.
 var crashing atomic.Int32
 
-// declared as global for dynamic change in TestGdbCoreCrashThreadBacktrace, see issue 64752.
-var totalSleepTimeUs = 5 * 1000 * 1000
-
 // testSigtrap and testSigusr1 are used by the runtime tests. If
 // non-nil, it is called on SIGTRAP/SIGUSR1. If it returns true, the
 // normal behavior on this signal is suppressed.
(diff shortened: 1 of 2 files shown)


src/runtime/signal_unix.go
```diff
@@ -769,19 +772,35 @@ func sighandler(sig uint32, info *siginfo, ctxt unsafe.Pointer, gp *g) {
...
 		}
 		if isCrashThread {
-			i := 0
-			for (crashing.Load() < mcount()-int32(extraMLength.Load())) && i < 10 {
-				i++
-				usleep(500 * 1000)
+			// Sleep for short intervals so that we can crash quickly after all ms have received SIGQUIT.
+			// Reset the timer whenever we see more ms received SIGQUIT
+			// to make it have enough time to crash (see issue #64752).
+			timeout := watchdogTimeoutMicros
```
ankit-gautam23, line 784: So many variable assignments I can see, just suggesting a small piece of code below

```
for timeout := watchdogTimeoutMicros; timeout > 0 && crashing.Load() < mcount()-int32(extraMLength.Load()); timeout -= crashSleepMicros {
    usleep(crashSleepMicros)
    if crashing.Load() > maxCrashing {
        maxCrashing = crashing.Load()
        timeout = watchdogTimeoutMicros
    }
}

```

zzkcode: Thanks, this pr has pending for Go 1.22 release before since the test failure may caused by another [issue](https://github.com/golang/go/issues/64752#issuecomment-1885242770). So this may not able to fix the issue, see [gerrit](https://go-review.googlesource.com/c/go/+/554615) for more details.

PR #65038, opened 2024-01-09
runtime: optimize permission changes with mprotect

ioworker0:
On Linux, both mprotect() and mmap() acquire the mmap_lock (in writer mode), 
posing scalability challenges.

The mmap_lock (formerly called mmap_sem) is a reader/writer lock that controls 
access to a process's address space; before making changes there (mapping in a 
new range, for example), the kernel must acquire that lock. 

Page-fault handling must also acquire mmap_lock (in reader mode) to ensure that 
the address space doesn't change in surprising ways while a fault is being resolved.

A process can have a large address space and many threads running (and incurring 
page faults) concurrently, turning mmap_lock into a significant bottleneck.

While both mmap() and mprotect() are protected by the mmap_lock, the shorter 
duration of mprotect system call, due to their simpler nature, results in a reduced 
locking time for the mmap_lock.

### DIFF
--- a/src/internal/runtime/syscall/defs_linux_386.go
+++ b/src/internal/runtime/syscall/defs_linux_386.go
@@ -6,6 +6,7 @@ package syscall
 
 const (
 	SYS_FCNTL         = 55
+	SYS_MPROTECT      = 125
 	SYS_EPOLL_CTL     = 255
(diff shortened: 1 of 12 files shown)


ioworker0:
@gopherbot

ioworker0:
@gopherbot ping

Issue #61530, opened 2023-07-22
net/http: ServeFile 2X slower in Go 1.21


r-hang:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.21rc3 linux/amd64
</pre>

### Does this issue reproduce with the latest release?
Yes. Slower in go1.21.

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
GOARCH='amd64'
GOBIN='/home/user/gocode/bin'
GOCACHE='/home/user/.cache/go-build'
GOENV='/home/user/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/user/gocode/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/user/gocode'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/home/user/gocode/src/github.com/golang/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/home/user/gocode/src/github.com/golang/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.21rc3'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='gcc'
CXX='g++'
CGO_ENABLED='1'
GOMOD=''
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -ffile-prefix-map=/tmp/go-build3856449379=/tmp/go-build -gno-record-gcc-switches'
</pre></details>

### What did you do?

[...]

Issue #16474, opened 2016-07-22
io: CopyBuffer should not perform io.ReaderFrom and io.WriterTo check


dsnet:
Currently, io.CopyBuffer uses WriteTo if the src supports it and ReadFrom if the dst supports it, presumably to avoid an allocation.

The problem is that there exist certain io.ReaderFrom that are not efficient 100% of the time. For example, consider the following implementation of ReadFrom:
```go
func (mw *myWriter) ReadFrom(src io.Reader) (int64, error) {
	if r, ok := ww.w.(io.ReaderFrom); ok {
		return r.ReadFrom(src)
	}
	return io.Copy(struct{ io.Writer }{mw}, src)
}
```
This is unfortunate as myWriter satisfies the io.ReaderFrom interface, but it's implementation is no better than io.Copy in some circumstances.

In the case of io.CopyBuffer, I would argue that the user's expectation is that the buffer passed in is *always* used. It has the detriment of an extra copy, but I believe the cost of that is much lower than the allocation that is already saved by the user providing the buffer. It also avoids any shenanigans that occur because of inefficient implementations of io.ReaderFrom and io.WriterTo.

Alternative is to do anonymous struct value hack:
```go
io.CopyBuffer(struct{ io.Writer }{dst}, struct{ io.Reader }{src}, buf)
```
But, this hack and not intuitive to most people.

Filed on behalf of @jkji

nhooyr:
Why would someone implement `io.ReaderFrom` or `io.WriterTo` if they are inefficient? Seems really backwards to me.

I see what you mean with the buf. Why pass a buffer if `io.CopyBuffer` will just end up calling `io.ReaderFrom`?

jkji:
ReaderFrom does not have to be horribly inefficient, but it can be _inefficient enough_ for the CopyBuffer caller to care. In my case, this was severely reducing the file transfer rate but it might be totally fine for some other application. Someone probably added it for a different use case, without realizing the side effects on CopyBuffer.
Conceptually, it seems that ReaderFrom should have a buffer argument and ignore it if it is not needed or nil. Efficient buffer-less ReaderFrom implementation sounds like a rare beast to me.

nhooyr:
Well we can't really change this now. Perhaps in Go 2?

dsnet:
[...]

PR #65106, opened 2024-01-15
net/http: use Copy in ServeContent if CopyN not needed

costela:
This small PR allows optimizations made in io.Copy (like the use of
io.WriterTo) to be used in one possible path of http.ServeContent
(in case of a non-Range request).
This, in turn, allows us to skip the buffer allocation in io.Copy.

This is a retry of CL446276, reverted to address #61530 (whose root
cause has since been addressed)

### DIFF
--- a/src/net/http/fs.go
+++ b/src/net/http/fs.go
@@ -374,7 +374,7 @@ func serveContent(w ResponseWriter, r *Request, name string, modtime time.Time, 
 	}
 	w.WriteHeader(code)
 
-	if r.Method != "HEAD" {
+	if r.Method != MethodHead {
 		if sendSize == size {
 			// use Copy in the non-range case to make use of WriterTo if available
 			io.Copy(w, sendContent)
--- a/src/net/http/fs.go
+++ b/src/net/http/fs.go
@@ -375,7 +375,12 @@ func serveContent(w ResponseWriter, r *Request, name string, modtime time.Time, 
 	w.WriteHeader(code)
 
 	if r.Method != "HEAD" {
-		io.CopyN(w, sendContent, sendSize)
+		if sendSize == size {
+			// use Copy in the non-range case to make use of WriterTo if available
+			io.Copy(w, sendContent)
+		} else {
+			io.CopyN(w, sendContent, sendSize)
+		}
 	}
 }
 
--- a/src/net/http/fs_test.go
+++ b/src/net/http/fs_test.go
@@ -1143,6 +1143,22 @@ func testServeContent(t *testing.T, mode testMode) {
 			wantStatus:  412,
 			wantLastMod: htmlModTime.UTC().Format(TimeFormat),
 		},
+		"uses_writeTo_if_available_and_non-range": {
+			content:          &panicOnNonWriterTo{seekWriterTo: strings.NewReader("foobar")},
+			serveContentType: "text/plain; charset=utf-8",
+			wantContentType:  "text/plain; charset=utf-8",
+			wantStatus:       StatusOK,
+		},
+		"do_not_use_writeTo_for_range_requests": {
+			content:          &panicOnWriterTo{ReadSeeker: strings.NewReader("foobar")},
+			serveContentType: "text/plain; charset=utf-8",
+			reqHeader: map[string]string{
+				"Range": "bytes=0-4",
+			},
+			wantContentType:  "text/plain; charset=utf-8",
+			wantContentRange: "bytes 0-4/6",
+			wantStatus:       StatusPartialContent,
+		},
 	}
 	for testName, tt := range tests {
 		var content io.ReadSeeker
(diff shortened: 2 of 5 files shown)


costela:
although the [excelent benchmark](https://github.com/golang/go/issues/61530#issuecomment-1646994568) by @AlexanderYastrebov seemed to indicate the underlying issue was solved, the [higher-level benchmark](https://github.com/golang/go/issues/61530#issuecomment-1646701847) by @mauri870 still shows a measurable slowdown (though not quite on the same level as originally observed; tested on current `master`):
```
goos: linux
goarch: amd64
pkg: go61530
cpu: 12th Gen Intel(R) Core(TM) i7-1250U
             │  old.bench  │             new.bench              │
             │   sec/op    │   sec/op     vs base               │
FileServe-12   17.27µ ± 6%   18.12µ ± 2%  +4.94% (p=0.007 n=10)

             │  old.bench   │               new.bench               │
             │     B/op     │     B/op      vs base                 │
FileServe-12   19.97Ki ± 0%   42.73Ki ± 0%  +114.01% (p=0.000 n=10)

             │ old.bench  │             new.bench              │
             │ allocs/op  │ allocs/op   vs base                │
FileServe-12   25.00 ± 0%   28.00 ± 0%  +12.00% (p=0.000 n=10)
```

Since the original [syntetic benchmarks](https://github.com/golang/go/pull/56480#issuecomment-1295451380) remain promising, this looks like another negative interaction with `os.File`.

I'll draft this PR while I investigate further. Sorry for the noise.

costela:
TL;DR: benchmarks were skewed; this PR should cause no performance regressions :tada: :crossed_fingers: 

Longer:

The benchmark results above are skewed by the using `httptest.ResponseRecorder`, which causes a fallback to `os.genericWriteTo` for not being one of the supported targets for a `WriteTo` from an `os.File`. 

Old (`CopyN`)
image

New (`Copy`)
image

The overhead is then caused by missing the [optimization in `copyBuffer` for `LimitReader`](https://github.com/golang/go/blob/7abeefd2b1a03932891e581f1f90656ffebebce4/src/io/io.go#L419). It pretty much disappears as soon as the the response size is bigger than the standard buffer size of 32k:
```
goos: linux
goarch: amd64
pkg: go61530
cpu: 12th Gen Intel(R) Core(TM) i7-1250U
             │  old.bench  │             new.bench              │
             │   sec/op    │   sec/op     vs base               │
FileServe-12   1.425m ± 2%   1.453m ± 5%  +1.95% (p=0.005 n=10)

             │  old.bench   │           new.bench            │
             │     B/op     │     B/op      vs base          │
FileServe-12   75.58Ki ± 3%   76.43Ki ± 3%  ~ (p=0.280 n=10)

             │ old.bench  │             new.bench              │
             │ allocs/op  │ allocs/op   vs base                │
FileServe-12   16.00 ± 0%   19.00 ± 0%  +18.75% (p=0.000 n=10)
```

Adding `ReadFrom` to the `http.ResponseRecorder` also negates the negative effects (I'd say this is an acceptable comparison with `http.response`, which also implements `ReaderFrom`):
```
goos: linux
goarch: amd64
pkg: go61530
cpu: 12th Gen Intel(R) Core(TM) i7-1250U
             │  old.bench  │           new.bench           │
             │   sec/op    │   sec/op     vs base          │
FileServe-12   1.018m ± 7%   1.032m ± 3%  ~ (p=0.165 n=10)

             │  old.bench   │           new.bench            │
             │     B/op     │     B/op      vs base          │
FileServe-12   32.45Ki ± 3%   31.10Ki ± 7%  ~ (p=0.165 n=10)

             │ old.bench  │             new.bench              │
             │ allocs/op  │ allocs/op   vs base                │
FileServe-12   15.00 ± 0%   18.00 ± 0%  +20.00% (p=0.000 n=10)
```

This means this PR shouldn't harm performance for the `io.File` case, and should enable `WriteTo` from sources that support if (e.g. [x/exp/mmap](https://github.com/golang/go/issues/56422#issue-1422973099), hopefully :crossed_fingers:)

costela:
AFAICT the PR should actually _improve_ performance for the realistic `io.File` case because `ReadFrom` in `http.response` will have access to the `io.File` (instead of `io.LimitReader`)? :thinking:

costela:
This still causes a regression in the `io.File` case:
image

1. `io.Copy` tries to use `WriteTo` from the `io.File` (not available if using `io.CopyN`), which fails because `http.response` doesn't implement `syscall.Conn`, so we use `genericWriteTo`
2. that calls a **second** round of `io.Copy`, now with a wrapped `io.File`, so we skip `WriteTo` and go straight to `http.response.ReadFrom`.
3. that then cannot use `sendFile`, because we got a wrapped file as source, so we're back to `genericReadFrom`
4. finally, a **third** round of `io.Copy`, which just falls back to buffered copy.

costela:
With the latest commit the regression is solved. We now see a working `ReadFrom`+`sendFile in the `io.File` case:
image

However, since there are currently multiple proposals open around the intersection of `WriterTo` and `io.Copy` (e.g. #16474, #58331, #56742), it may be necessary to split the last commit into a proposal.

Alternatively, we could also add an internal interface that is only implemented by `io.File`, so that all type checks can be done against it. This would ensure this change has no side-effects outside of the stdlib. In its current form, the change has an unlikely impact, but could still theoretically accidentally match some interface in the wild.

PR #65107, opened 2024-01-15
cmd/internal/bio: resolve todo to query vm.max_map_count on linux for mmapLimit

exapsy:
Resolves todo comment for linux reading the vm.max_map_count from the system

### DIFF
--- a/src/cmd/internal/bio/buf_mmap.go
+++ b/src/cmd/internal/bio/buf_mmap.go
@@ -7,7 +7,10 @@
 package bio
 
 import (
+	"os"
 	"runtime"
+	"strconv"
+	"strings"
 	"sync/atomic"
 	"syscall"
 )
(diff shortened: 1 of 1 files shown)


src/cmd/internal/bio/buf_mmap.go
```diff
@@ -20,18 +23,32 @@ import (
...
 
+func getLinuxMaxMapCount() int32 {
+	data, err := os.ReadFile("/proc/sys/vm/max_map_count")
+	if err != nil {
+		return 1<<31 - 1
+	}
+
+	maxMapCount, err := strconv.Atoi(strings.TrimSpace(string(data)))
+	if err != nil {
+		return 1<<31 - 1
```
exapsy, line 46: I would leave this under debate whether it should be as previously, 30000, or max.

PR #65233, opened 2024-01-23
net: document LookupTXT behavior with multiple strings per record

emersion:
CL 8966 ("net: allow a dns TXT record to contain more than
one <character-string>") concatenates strings in a TXT record. Document
that behavior, since it's not obvious whether the strings will be
concatenated or accumulated in the returned slice.


emersion:
@ianlancetaylor @neild any chance to get this merged?

PR #65360, opened 2024-01-30
errors: unwrap joinError to avoid deeply-nested joins

sethvargo:
This PR is a quick proposal to unwrap `joinError` (and only `joinError`)
when calling `errors.Join`. While I don't think we should unwrap
everything, this is an internal structure that has only internally
semantic meaning. Therefore, I think it makes sense to special-case
`joinError` and unwrapped the contained errors on `joinError`.

Pros:
- More closely mirrors how other error joining packages in the ecosystem
  behave.

- Even though `Unwrap() []error` isn't a public API, this will ensure
  that the method behaves as end-users expect for `errors.Join`,
  returning the list of joined errors.

Cons:
- Introduces type assertion and could result in a performance
  regression, since we're not pre-computing the slice length. This could
  be addressed if there's interest in pursuing this. I also think there's
  a reasonable argument to be made about the benefits realized from this
  change when printing the error.

- Users could be depending on the broken behavior, but that seems
  unlikely since `joinError` is an internal API.

I could not find any corresponding issues, but "errors" and "unwrap" are
not particularly unique keywords, so apologies if I missed them.

### DIFF
--- a/src/errors/join.go
+++ b/src/errors/join.go
@@ -17,24 +17,33 @@ import (
 //
 // A non-nil error returned by Join implements the Unwrap() []error method.
 func Join(errs ...error) error {
-	n := 0
-	for _, err := range errs {
-		if err != nil {
-			n++
+	allErrs := make([]error, 0, len(errs))
+	for _, e := range errs {
+		// Ignore nil errors.
+		if e == nil {
+			continue
 		}
+
+		// Specifically handle nested join errors from the standard library. This
+		// avoids deeply-nesting values which can be unexpected when unwrapping
+		// errors.
+		joinErr, ok := e.(*joinError)
+		if !ok {
+			allErrs = append(allErrs, e)
+			continue
+		}
+
+		allErrs = append(allErrs, joinErr.errs...)
 	}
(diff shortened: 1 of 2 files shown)


src/errors/join.go
```diff
@@ -17,24 +17,33 @@ import (
 //
 // A non-nil error returned by Join implements the Unwrap() []error method.
 func Join(errs ...error) error {
-	n := 0
-	for _, err := range errs {
-		if err != nil {
-			n++
+	allErrs := make([]error, 0, len(errs))
```
sethvargo, line 20: This might be one of those situations where optimizing for the common use case makes sense. Given "most" error joining is of the form `merr = errors.Join(merr, err)`, we can optimize for the single-entry error:

```go
switch len(errs) {
  case 0:
    return nil
  case 1:
    return errs[0]
  case 2:
    if errs[0] == nil {
      return errs[1]
    } else if errs[1] == nil {
      return errs[0]
    } else {
      return &joinError{
        errs: errs, 
      }
    }
  default:
    // as below
}
```

Issue #65355, opened 2024-01-29
runtime: move internals to `internal/runtime`


prattmic:
Right now package runtime appears to serve two purposes:

* It contains most of the implementation of the Go runtime.
* And, it exposes the world-public API for various runtime-related functionality (`runtime.GOMAXPROCS`, etc).

At first glance, this seems perfectly normal and reasonable, however there is actually a third purpose:

* Expose `std`-internal APIs that should not be world-public. e.g., `runtime.semacquire` for `sync.Mutex` or `runtime.netpollopen` for `os` and `net`.

These APIs should not be public to the whole world as they are implementation details that may change. However, with the current layout there is no good way to expose these internal APIs, so we must use `//go:linkname` between packages, and in some cases copy entire type definitions between packages (#64549).

If we move most of the runtime implementation to `internal/runtime`, then it could expose a richer API surface useful to other packages in `std`, but nothing outside of `std` would be able to access them due to the rules of `internal`. Package `runtime` would remain to provide the same API, but it would be much smaller and simply import `internal/runtime` for most functionality.

In moving, there are also opportunities for cleanup along the lines of #51087. e.g., perhaps the timer implementation is in `internal/runtime/timer`, etc. Though this move isn't really a prerequisite for that; we could do that in `runtime/internal/timer` as well.

This change would technically be solely an implementation detail, though in practice there are likely symbols that folks depend on anyway, either via linkname (https://go.dev/cl/548235) or implicitly by looking for in profiles, stack traces, etc.

cc @golang/runtime

mauri870:
I really like the idea of having a `internal/runtime` package. The `runtime` code always felt too "convoluted"  for me. It grew a lot with time to the point it is now a sum of numerous assembly files, internal code for the GC and auxiliary runtime stuff, race, asan, msan code, and buried in the middle of all this the public runtime apis.  I believe having an internal package dedicated to these bits would hopefully clean `runtime` enough that it will be easier to navigate and more inlined with the other standard library packages.

PR #65427, opened 2024-02-01
Clean up duplicate function  IssystemGoroutine

Jun10ng:
For [#65355](https://go.dev/issue/65355)

### DIFF
--- a/src/internal/trace/goroutines.go
+++ b/src/internal/trace/goroutines.go
@@ -5,8 +5,8 @@
 package trace
 
 import (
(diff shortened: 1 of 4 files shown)


src/internal/trace/goroutines.go
```diff
@@ -5,8 +5,8 @@
 package trace
 
 import (
+	"internal/runtime/goroutine"
```
Jun10ng, line 8: I am not think "goroutine" is a good package name, any advance?

PR #65692, opened 2024-02-13
syscall: enable netlink support on FreeBSD

igalic:
- enable netlink support on FreeBSD
- rebuild syscalls for Tier 1 architectures (arm64 & amd64) and 386


igalic:
n.b.: I wasn't able to rebuild syscalls for riscv64 and arm;
Hopefully someone else can provide those; or I'll be able to rebuild my VMs… to actually work.
In the meantime, I wanted to open this up for feedback.

PR #66186, opened 2024-03-08
strings: add strings helpers for prefixes and suffixes

testwill:
This PR relates to proposal: https://github.com/golang/go/issues/66191


darth-raijin (commented):
Use another word instead of indicates

PR #66231, opened 2024-03-10
slices: add slices.Unique

antonsacred:
I found this function useful as in other languages:
1. array_unique in php
2. unique() in rust


ayanmw:
I prefer slices.Unique() , not  slices.Sort and slices.Compact ... 

slices.Compact is useless than slices.Unique()  just I think.

PR #66349, opened 2024-03-16
Fix comment typo on slog package

andreleoni:
I'm studying the slog library for a presentation and realized this comment might be wrong.

just a suggested correction.

It will be an honor to be able to collaborate with the project, even if it is just a drop of water. 💧


seankhliao:
https://en.wikipedia.org/wiki/If_and_only_if

andreleoni:
thanks for reply @seankhliao !!

PR #66648, opened 2024-04-02
syscall: implement Mkfifo on illumos

jperkin:


jperkin:
This fixes the build of newer versions of gitea, which are currently failing on illumos with the following error:

```
# github.com/sassoftware/go-rpmutils/fileutil
../.gopath/pkg/mod/github.com/sassoftware/go-rpmutils@v0.2.0/fileutil/fileutil_unix.go:37:17: undefined: syscall.Mkfifo
```

With this patch applied to the pkgsrc build in https://github.com/NetBSD/pkgsrc/commit/4d5c71236dcd79888e58d34050337bdf907396a8 the build succeeds.

PR #66684, opened 2024-04-04
syscall/js: allocate arg slices on stack for small numbers of args

### LINKED ISSUE #39740: syscall/js: Increase performance of Call, Invoke, and New by not allowing new slices to escape onto the heap

unitoftime:
The existing implementation causes unnecessary heap allocations for
javascript syscalls: Call, Invoke, and New. The new change seeks to
hint the Go compiler to allocate arg slices with length <=16 to the
stack.

Original Work: CL 367045
- Calling a JavaScript function with 16 arguments or fewer will not
induce two additional heap allocations, at least with the current Go
compiler.
- Using syscall/js features with slices and strings of
statically-known length will not cause them to be escaped to the heap,
at least with the current Go compiler.
- The reduction in allocations has the additional benefit that the
garbage collector runs less often, blocking WebAssembly's one and only
thread less often.

Fixes #39740

### DIFF
--- a/src/syscall/js/js.go
+++ b/src/syscall/js/js.go
@@ -210,7 +210,13 @@ func ValueOf(x any) Value {
 	}
 }
 
+// stringVal copies string x to Javascript and returns a ref.
+//
+// (noescape): This is safe because no references are maintained to the
+//             Go string x after the syscall returns.
+//
 //go:wasmimport gojs syscall/js.stringVal
+//go:noescape
 func stringVal(x string) ref
 
 // Type represents the JavaScript type of a Value.
@@ -294,7 +300,13 @@ func (v Value) Get(p string) Value {
 	return r
 }
 
+// valueGet returns a ref to JavaScript property p of ref v.
+//
+// (noescape): This is safe because no references are maintained to the
+//             Go string p after the syscall returns.
+//
 //go:wasmimport gojs syscall/js.valueGet
+//go:noescape
 func valueGet(v ref, p string) ref
 
 // Set sets the JavaScript property p of value v to ValueOf(x).
@@ -309,7 +321,13 @@ func (v Value) Set(p string, x any) {
 	runtime.KeepAlive(xv)
 }
 
+// valueSet sets property p of ref v to ref x.
+//
+// (noescape): This is safe because no references are maintained to the
+//             Go string p after the syscall returns.
+//
 //go:wasmimport gojs syscall/js.valueSet
+//go:noescape
 func valueSet(v ref, p string, x ref)
 
 // Delete deletes the JavaScript property p of value v.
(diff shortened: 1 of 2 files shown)


unitoftime:
Putting this here for posterity (Also linked in code review: https://go-review.googlesource.com/c/go/+/576575/comments/16d89df8_6a869458 )

Hi, Thanks for reviewing this change.

I just wanted to give some additional context about this change and where it came from. As I did not author all of it. The original research and work was done by finnbear (https://github.com/finnbear): https://github.com/golang/go/pull/49799 - I'm hoping to help get his changes through the review process. Notes from his original PR (this is from Finnbear):

- Calling a JavaScript function with 16 arguments or fewer will not induce two additional heap allocations, at least with the current Go compiler
- Using syscall/js features with slices and strings of statically-known length will not cause them to be escaped to the heap, at least with the current Go compiler
- The reduction in allocations has the additional benefit that the garbage collector runs less often, blocking WebAssembly's one and only thread less often
Other Notes (The rest is from me):

---

1. Tested locally on a WASM game I am building to verify that the heap allocations have all been removed in the expected areas
2. Ran what appears to be the full test suite of the go repo: ./all.bash

This is my first PR to golang, so please let me know if I've done anything incorrectly, or if you need anything else! Thanks!

PR #66757, opened 2024-04-10
syscall: implement O_DIRECT on illumos

jperkin:


jperkin:
This fixes the build of minio, which currently fails due to:

```
# github.com/ncw/directio
../.gopath/pkg/mod/github.com/ncw/directio@v1.0.5/direct_io_unix.go:22:35: undefined: syscall.O_DIRECT
```

PR #66801, opened 2024-04-12
crypto/rand: make Read not escape the byte slice

### LINKED ISSUE #66779: crypto/rand: Read argument escapes to heap
Currently, Read is implemented as `io.ReadFull(Reader, b)` which escape analysis can't track, so b escapes to the heap.

This is somewhat annoying, since it causes allocations in paths that could otherwise have none.

It should be possible to rewrite the package so that it doesn't escape on major platforms.

mateusz834:
│ /tmp/before │             /tmp/after             │
              │   sec/op    │   sec/op     vs base               │
ReadAllocs-12   3.050µ ± 0%   2.912µ ± 0%  -4.52% (p=0.000 n=10)

              │ /tmp/before  │             /tmp/after              │
              │     B/s      │     B/s       vs base               │
ReadAllocs-12   320.2Mi ± 0%   335.3Mi ± 0%  +4.73% (p=0.000 n=10)

              │ /tmp/before  │              /tmp/after               │
              │     B/op     │     B/op      vs base                 │
ReadAllocs-12   1.000Ki ± 0%   0.000Ki ± 0%  -100.00% (p=0.000 n=10)

              │ /tmp/before │             /tmp/after              │
              │  allocs/op  │ allocs/op   vs base                 │
ReadAllocs-12    1.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)


mateusz834:
Closing in favour of [CL 602498](https://go.dev/cl/602498)

PR #66805, opened 2024-04-13
os/file: return a fs.ReadFileFS instead of a fs.FS for os.DirFS

kilianc:
`func DirFS(dir string) fs.FS` return a `fs.FS` but implements `ReadFile`. returning `fs.ReadFileFS` instead will allow `ReadFile` to be called on the returned instance.

### DIFF
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -680,7 +680,7 @@ func (f *File) SyscallConn() (syscall.RawConn, error) {
 //
 // The result implements [io/fs.StatFS], [io/fs.ReadFileFS] and
 // [io/fs.ReadDirFS].
-func DirFS(dir string) fs.ReadDirFile {
+func DirFS(dir string) fs.ReadFileFS {
 	return dirFS(dir)
 }
 
--- a/src/os/file.go
+++ b/src/os/file.go
@@ -680,7 +680,7 @@ func (f *File) SyscallConn() (syscall.RawConn, error) {
 //
 // The result implements [io/fs.StatFS], [io/fs.ReadFileFS] and
 // [io/fs.ReadDirFS].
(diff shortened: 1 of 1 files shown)


src/os/file.go
```diff
@@ -680,7 +680,7 @@ func (f *File) SyscallConn() (syscall.RawConn, error) {
 //
 // The result implements [io/fs.StatFS], [io/fs.ReadFileFS] and
 // [io/fs.ReadDirFS].
-func DirFS(dir string) fs.FS {
+func DirFS(dir string) fs.ReadDirFS {
```
kilianc, line 683:
```suggestion
func DirFS(dir string) fs.ReadFileFS {
```

src/os/file.go
```diff
@@ -680,7 +680,7 @@ func (f *File) SyscallConn() (syscall.RawConn, error) {
 //
 // The result implements [io/fs.StatFS], [io/fs.ReadFileFS] and
 // [io/fs.ReadDirFS].
-func DirFS(dir string) fs.FS {
+func DirFS(dir string) fs.ReadFileFS {
```
kilianc, line 683:
```suggestion
func DirFS(dir string) fs.ReadDirFile {
```

src/os/file.go
```diff
@@ -680,7 +680,7 @@ func (f *File) SyscallConn() (syscall.RawConn, error) {
 //
 // The result implements [io/fs.StatFS], [io/fs.ReadFileFS] and
 // [io/fs.ReadDirFS].
-func DirFS(dir string) fs.FS {
+func DirFS(dir string) fs.ReadDirFile {
```
kilianc, line 683:
```suggestion
func DirFS(dir string) fs.ReadFileFS {
```

PR #67157, opened 2024-05-03
crypto/x509: include OID in duplicate extension error message

### LINKED ISSUE #66880: crypto/x509: ParseCertificate and ParseCertificateRequest duplicate extensions errors should include OID of the affected extension
I recently ran into the issue that generated X.509 certificates very rejected by Go due to containing duplicate extensions. The exact extension was not reported in the error message, which made the debugging process rather complicated - it required manual changes in the standard library. The current error message reads: `x509: certificate contains duplicate extensions`.

Therefore, I would like to propose to change the error message to include the detected duplicate OID. The error message could then look like: `x509: certificate contains duplicate extension {oid}`. Feel free to suggest a different error message format.

I'm also happy to contribute the required changes in a separate PR if it is decided to move forward.

Techassi:
Include the OID in the error message when parsing X.509
certificates. This should ease fixing such issues, because
users can clearly identify the duplicate extension via the
reported error. Previously, this wasn't possible and
required either manually adjusting the standard library or
inspecting the certificate with various debugging tools.

Fixes #66880

### DIFF
--- a/src/crypto/x509/parser.go
+++ b/src/crypto/x509/parser.go
@@ -964,7 +964,7 @@ func parseCertificate(der []byte) (*Certificate, error) {
 					}
 					oidStr := ext.Id.String()
 					if seenExts[oidStr] {
-						return nil, errors.New("x509: certificate contains duplicate extensions")
+						return nil, errors.New("x509: certificate contains duplicate extension %s", oidStr)
 					}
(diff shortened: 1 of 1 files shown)


NickLarsenNZ (approved):
Thanks

PR #67254, opened 2024-05-08
runtime: don't fetch c-archive/c-shared os.Args on musl/uclinux

1800alex:
This works around the segfault seen when compiling a shared archive and linking it on a musl/uclinux system.

The issue is that the go runtime is doing some magic to extract the program arguments even though the go side is compiled as a library.

By including my changes the glibc behavior remains unchanged, and for musl/uclinux systems we simply do not get any arguments if we are linked as a shared library.

I've attempted to get these changes included in the past and have been manually patching my go releases for alpine.

I believe the issue raised previously was that in theory we could parse the arguments from /proc but honestly I don't have time to mess with that and don't see the need for a library to be aware of argc/argv unless someone here thinks otherwise.


1800alex:
I created an example sandbox repository to demonstrate the issue before/after the patch has been applied https://github.com/1800alex/go-cshared-example

PR #67369, opened 2024-05-14
runtime: add GoID to StackRecord produced by GoroutineProfile()

vaind:
Making this change allows mapping stack traces to goroutines across 
subsequent runs of `GoroutineProfile()`. This is necessary in order to 
use said function as a simple continuous in-process profiler (with sample 
timestamps). The other option, involving `runtime.Stack()` has worse 
performance (in the runtime package) and also produces stack as a string 
which then needs to be parsed back (although, at least it has the goroutine 
ID)

Closes #59663


vaind:
#59663 proposal has been declined so closing this.

PR #67422, opened 2024-05-16
enconding/json: add ValidPos() function

dveeden:
Add json.ValidPos() to allow getting the position of where the validation failed
in addition to a boolean to indicate the failure.

Closes #67423

### DIFF
--- a/src/encoding/json/scanner.go
+++ b/src/encoding/json/scanner.go
@@ -25,6 +25,12 @@ func Valid(data []byte) bool {
 	return checkValid(data, scan) == nil
 }
 
+func ValidPos(data []byte) (bool, int64) {
+	scan := newScanner()
+	defer freeScanner(scan)
+	return checkValid(data, scan) == nil, scan.bytes
+}
+
 // checkValid verifies that data is valid JSON-encoded data.
 // scan is passed in for use by checkValid to avoid an allocation.
 // checkValid returns nil or a SyntaxError.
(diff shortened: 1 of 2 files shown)


dveeden:
We use `json.Valid()` in https://github.com/pingcap/tidb to implement the [`JSON_VALID()`](https://docs.pingcap.com/tidb/dev/json-functions/) SQL function and would like to return the position on which the validation failed to the user.

https://github.com/pingcap/tidb/blob/0875abde25467897e4e81ae6c18c5e8792ca28a9/pkg/expression/builtin_json.go#L1019

Example usage:
```go
package main

import (
	"encoding/json"
	"fmt"
	"strings"
)

func main() {
	inputs := []string{
		`{}`,
		`{"foo": "bar"}`,
		`{"foo: "bar"}`,
		`{"foo": "bar", "bar": foo}`,
		`{"foo": "bar", "bar", foo}`,
	}
	for _, input := range inputs {
		v, pos := json.ValidPos([]byte(input))
		fmt.Printf("'%s': valid: %#v, pos: %d\n", input, v, pos)
		if !v {
			fmt.Printf(strings.Repeat(" ", int(pos)) + "^\n")
		}
	}
}
```

Output:
```
go run main.go 
'{}': valid: true, pos: 2
'{"foo": "bar"}': valid: true, pos: 14
'{"foo: "bar"}': valid: false, pos: 9
         ^
'{"foo": "bar", "bar": foo}': valid: false, pos: 24
                        ^
'{"foo": "bar", "bar", foo}': valid: false, pos: 21
                     ^
```

This would make it easier to see why validation failed, especially for larger JSON documents.

PR #67501, opened 2024-05-18
fmt: replace sort with slices to reduce dependency tree

aimuz:
This CL replaces the usage of sort with slices in fmt package, reducing
the dependency tree. The fmt package previously depended on sort, which
in turn depended on slices and other packages. This change simplifies
the dependencies.

### DIFF
--- a/src/fmt/errors.go
+++ b/src/fmt/errors.go
@@ -6,7 +6,7 @@ package fmt
 
 import (
 	"errors"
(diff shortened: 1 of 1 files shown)


ridhoramawijaya (approved):

aimuz:
close for [67510](https://github.com/golang/go/issues/67510)

PR #67539, opened 2024-05-21
maps: clarify iteration order and insertion behavior

### LINKED ISSUE #67537: maps: Clarify iteration order of Keys & Values and that Insert overwrites existing keys
These functions were added in 2b0f2f8169bed01e3bf72d49863cd2852bdf7c7e with minimal documentation. 

```go
// Keys returns an iterator over keys in m.
func Keys[Map ~map[K]V, K comparable, V any](m Map) iter.Seq[K] {
	...
}

// Values returns an iterator over values in m.
func Values[Map ~map[K]V, K comparable, V any](m Map) iter.Seq[V] {
	...
}

// Insert adds the key-value pairs from seq to m.
func Insert[Map ~map[K]V, K comparable, V any](m Map, seq iter.Seq2[K, V]) {
	...
}
```

The doc comments should be updated to specify the iteration order of Keys and Values, and the behaviour of Insert when duplicate keys are inserted, per discussion on #61900 and in particular [rsc's comment](https://github.com/golang/go/issues/61900#issuecomment-1944349828):

> Happy to update docs to say that iteration order is undefined (different each time) and Insert does overwrite keys.

aimuz:
Fixes #67537.

### DIFF
--- a/src/maps/iter.go
+++ b/src/maps/iter.go
@@ -7,6 +7,8 @@ package maps
 import "iter"
 
 // All returns an iterator over key-value pairs from m.
+// The iteration order is not specified and is not guaranteed
+// to be the same from one call to the next.
 func All[Map ~map[K]V, K comparable, V any](m Map) iter.Seq2[K, V] {
 	return func(yield func(K, V) bool) {
 		for k, v := range m {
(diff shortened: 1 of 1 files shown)


magical:
@aimuz Thank you!

PR #67888, opened 2024-06-07
time: add to format documentation about dangers of using RFC1123{,Z} for parsing

### LINKED ISSUE #67887: time: RFC1123{,Z} format strings appear to be very slightly wrong
### Go version

go version go1.22.1 darwin/arm64

```shell
GO111MODULE=''
GOARCH='arm64'
GOBIN=''
GOCACHE='/Users/dburkart/Library/Caches/go-build'
GOENV='/Users/dburkart/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='arm64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/dburkart/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='darwin'
GOPATH='/Users/dburkart/go'
GOPRIVATE=''
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/darwin_arm64'
GOVCS=''
GOVERSION='go1.22.1'
GCCGO='gccgo'
AR='ar'
CC='clang'
CXX='clang++'
CGO_ENABLED='1'
GOMOD='/dev/null'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/1m/2p4knn216k755wpkh1vlz4_00000gn/T/go-build1317344447=/tmp/go-build -gno-record-gcc-switches -fno-common'
```

### What did you do?

We have been using `time.RFC1123Z` to parse the `Date` header of emails.

We found that sometimes this would work, and sometimes it would not. I tracked this down to "the time of the month". Specifically, if you try to parse an email date from days 1 - 9 of the month, `time.Parse()` with `time.RFC1123Z` throws an error, but on days 10 - 3x, it parses just fine.

Specifically, it seems that this part of [RFC 1123](https://www.rfc-editor.org/rfc/rfc1123#page-55) indicates that the day can ~~only be 2 digits if the first is a '1'~~ one or two digits:

`date = 1*2DIGIT month 2*4DIGIT`

This means that a Date string (that I pulled from one of my emails) like this doesn't parse:

`Fri, 7 Jun 2024 13:04:42 -0400`

But this will:

`Fri, 31 May 2024 09:57:48 +0000`

### What did you see happen?

Most of the time this works, but sometimes it does not.

It should always work, since RFC 1123 defines the date / time expected in an email header.

dburkart:
When using time.RFC1123Z to parse the date header value out of an email,
an error is returned for dates that occur in the first 9 days of a
month. This is because the format strings for RFC 1123 defined in the
time package indicate that the day should be prefixed with a leading 0.

Reading the spec, the line that talks about it seems to indicate that
days can be either 1 or 2 digits:

  `date = 1*2DIGIT month 2*4DIGIT`

So a date header with a day like `7` with no leading zero should be
accepted.

Fixes #67887

### DIFF
--- a/src/time/format.go
+++ b/src/time/format.go
@@ -32,6 +32,10 @@ import (
 // only to local times. Applying them to UTC times will use "UTC" as the
 // time zone abbreviation, while strictly speaking those RFCs require the
 // use of "GMT" in that case.
+// When using the [RFC1123] or [RFC1123Z] formats for parsing, note that these
+// formats define a leading zero for the day-in-month portion, which is not
+// strictly allowed by RFC 1123. This will result in an error when parsing
+// date strings that occur in the first 9 days of a given month.
 // In general [RFC1123Z] should be used instead of [RFC1123] for servers
 // that insist on that format, and [RFC3339] should be preferred for new protocols.
 // [RFC3339], [RFC822], [RFC822Z], [RFC1123], and [RFC1123Z] are useful for formatting;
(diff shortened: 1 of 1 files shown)


dburkart:
Note: I am not sure if the best approach is to fix the existing format strings, or define new (more conformant) ones... since conceivably someone could be depending on the leading 0 defined in the current strings.

PR #68143, opened 2024-06-23
net/http: make the http server con write buffer size configurable

oddmario:
This change adds a configurable WriteBufferSize option to the Go HTTP server to be used (if set) instead of the hard coded 4 KB

Fixes #68149


neild:
Sorry, I managed to copy a repo name from the wrong window and accidentally overwrote the PR briefly. I believe I've put things back the way they were. Again, sorry!

PR #68239, opened 2024-06-28
sys/cpu: add Int8 matrix multiplication instructions CPU feature flag for ARM64

hmartinez82:
References:
- https://github.com/torvalds/linux/blob/5bbd9b249880dba032bffa002dd9cd12cd5af09c/arch/arm64/include/uapi/asm/hwcap.h#L75C9-L75C31
- https://developer.arm.com/documentation/ddi0601/2024-03/AArch64-Registers/ID-AA64ISAR1-EL1--AArch64-Instruction-Set-Attribute-Register-1


hmartinez82:
Pull request created against https://go.googlesource.com/sys instead

PR #68337, opened 2024-07-08
cmd/go: add file names for cyclic import error

### LINKED ISSUE #66078: cmd/go: print file name when showing an import cycle
We've all hit the common:

```
package foo
	imports whatever1
	imports whatever2
	imports whatever3
	imports bar: import cycle not allowed
```

It would great if it showed the actual file name of the imports, so that when there are multiple files in a package, it's easier to narrow down which one is the offender.

Eg:

```
package foo: blah.go
	imports whatever1: file1.go
	imports whatever2: something.go
	imports whatever3: thing.go
	imports bar: import cycle not allowed
```

Of course a more symmetrical approach might look like:

```
import cycle not allowed
imports foo: blah.go
imports whatever1: file1.go
imports whatever2: something.go
imports whatever3: thing.go
imports package foo
```

But whatever the formatting the code idea is that we make it easier for the user to know which file was part of the loop.

What about the situation of more than one file offending? Well this is a topological sort, so we already only guarantee that we print a single loop, even when there are more than one paths.

Of note, we _don't_ want a full absolute path printed, we already know where things should be based on package name, we do care about the relative file path in that package.

Thanks!

xin-tsla:
The PR is to add more details for the error, so that it would be easier to troubleshoot the cyclic imports error.

The change for the error looks like the following:

package cyclic-import-example
        imports cyclic-import-example/packageA from /Users/personal/cyclic-import-example/main.go:4:5
        imports cyclic-import-example/packageB from /Users/personal/cyclic-import-example/packageA/a.go:5:2
        imports cyclic-import-example/packageA from /Users/personal/cyclic-import-example/packageB/bb.go:5:2: import cycle not allowed

Fixes #66078

### DIFF
--- a/src/cmd/go/alldocs.go
+++ b/src/cmd/go/alldocs.go
@@ -905,12 +905,8 @@
 //
 //	type PackageError struct {
 //	    ImportStack        []string // shortest path from package named on command line to this one
-//	    ImportStackWithPos []string // shortest path from package named on command line to this one with position
 //	    Pos                string   // position of error (if present, file:line:col)
 //	    Err                string   // the error itself
-//	    IsImportCycle      bool     // the error is an import cycle
-//	    Hard               bool     // whether the error is soft or hard; soft errors are ignored in some places
-//	    alwaysPrintStack   bool     // whether to always print the ImportStack
 //	}
 //
 // The module information is a Module struct, defined in the discussion
--- a/src/cmd/go/alldocs.go
+++ b/src/cmd/go/alldocs.go
@@ -904,9 +904,13 @@
 // The error information, if any, is
 //
 //	type PackageError struct {
-//	    ImportStack   []string // shortest path from package named on command line to this one
-//	    Pos           string   // position of error (if present, file:line:col)
-//	    Err           string   // the error itself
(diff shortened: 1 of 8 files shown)


xin-tsla:
Link issue here:
https://github.com/golang/go/issues/66078

xin-tsla:
Hi @michaelmatloob, wondering are you still seeing markdown in the CL?
image

qiulaidongfeng:
@xin-tsla please change https://github.com/golang/go/pull/68337#issue-2394612339 , Fixes [#66078](https://go.dev/issue/66078) change to Fixes #66078
Note that on github, for #66078, with or without markdown, the same thing is displayed, but editing it reveals a difference.

xin-tsla:
> @xin-tsla please change [#68337 (comment)](https://github.com/golang/go/pull/68337#issue-2394612339) , Fixes [#66078](https://go.dev/issue/66078) change to Fixes #66078 Note that on github, for #66078, with or without markdown, the same thing is displayed, but editing it reveals a difference.

@qiulaidongfeng , ah, got it. Thank you! Yes, it is different. Without edit mode, they look same.

PR #68519, opened 2024-07-18
text/template: fix doc spacing

willfaught:


willfaught:
Copying the original PR description here:

---

This doc syntax:

```md
- The name of a field of the data, which must be a struct, preceded
  by a period, such as
		.Field
```

was producing this formatting in the terminal and pkg.go.dev:

```
The name of a field of the data, which must be a struct, preceded by a period, such as .Field The result is the value of the field.
```

This change makes the formatting more aligned with the intent, and doesn't have bad punctuation:

```
The name of a field of the data, which must be a struct, preceded by a period, such as

.Field

The result is the value of the field.
```

There doesn't appear to be a way to have pre-formatted content inside a list in Go doc, sadly.

PR #68524, opened 2024-07-19
strconv: document that Unquote("''") returns an empty string

### LINKED ISSUE #64280: strconv: Document that Unquote("''") doesn't return error

<pre>
$ go version
go version go1.21.4 darwin/arm64
</pre>

The documentation for `strconv.Unquote` currently reads

> package strconv // import "strconv"
> 
> func Unquote(s string) (string, error)
>     Unquote interprets s as a single-quoted, double-quoted, or backquoted
>     Go string literal, returning the string value that s quotes. (If s is
>     single-quoted, it would be a Go character literal; Unquote returns the
>     corresponding one-character string.)

However, `strconv.Unquote("''")` equals the empty string, whereas from the documentation one would assume this to return an invalid-syntax error. Perhaps a reasonable addition to the documentation would be

> If s is an empty pair of single quotes, Unquote returns the empty string.

yossev:
Fixes #64280

### DIFF
--- a/src/strconv/quote.go
+++ b/src/strconv/quote.go
@@ -378,7 +378,7 @@ func QuotedPrefix(s string) (string, error) {
 // or backquoted Go string literal, returning the string value
 // that s quotes.  (If s is single-quoted, it would be a Go
 // character literal; Unquote returns the corresponding
-// one-character string, For '' Unquote returns the empty string.)
+// one-character string; For '' Unquote returns the empty string.)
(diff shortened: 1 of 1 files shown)


Jorropo:
If anyone got an email about the previous ai powered `pip` installed ai powered issue resolver.
It was created 8 hours ago, a bot is spamming instructions on how to use it on random unrelated repos.

This is very likely to be malware disguised as a python CLI. Use with caution.

PR #68586, opened 2024-07-25
misc/wasm: remove unused argument

Zxilly:
https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/TextEncoder 
TextEncoder receives no argument and always works under utf-8

previous https://github.com/golang/go/pull/67778


wild-fruits:
- [x] > yhhj@benburkert []()~~~~

PR #68964, opened 2024-08-20
src: fix typos

Sajjon:
Fix typos in ~30 files

### DIFF
--- a/src/cmd/internal/obj/loong64/obj.go
+++ b/src/cmd/internal/obj/loong64/obj.go
@@ -411,7 +411,7 @@ func preprocess(ctxt *obj.Link, cursym *obj.LSym, newprog obj.ProgAlloc) {
 				// The NOP is needed to give the jumps somewhere to land.
 				// It is a liblink NOP, not a hardware NOP: it encodes to 0 instruction bytes.
 				//
-				// We don't generate this for leafs because that means the wrapped
+				// We don't generate this for leaves because that means the wrapped
 				// function was inlined into the wrapper.
 
 				q = obj.Appendp(q, newprog)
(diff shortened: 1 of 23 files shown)


src/cmd/internal/obj/loong64/obj.go
```diff
@@ -411,7 +411,7 @@ func preprocess(ctxt *obj.Link, cursym *obj.LSym, newprog obj.ProgAlloc) {
 				// The NOP is needed to give the jumps somewhere to land.
 				// It is a liblink NOP, not a hardware NOP: it encodes to 0 instruction bytes.
 				//
-				// We don't generate this for leafs because that means the wrapped
+				// We don't generate this for leaves because that means the wrapped
```
Sajjon, line 414: Understood if you prefer to revert this. I prefer to use correct English at all times, however, some people might find `leafs` more suitable in code.

src/net/http/fs.go
```diff
@@ -264,7 +264,7 @@ var errNoOverlap = errors.New("invalid range: failed to overlap")
 
 // if name is empty, filename is unknown. (used for mime type, before sniffing)
 // if modtime.IsZero(), modtime is unknown.
-// content must be seeked to the beginning of the file.
+// content must be sought to the beginning of the file.
```
Sajjon, line 267: Some people might find (incorrect form) `"seeked"` more easy to read in code, understood if you prefer me to revert this.

src/runtime/map_noswiss.go
```diff
@@ -358,7 +358,7 @@ func makemap(t *maptype, hint int, h *hmap) *hmap {
 // 1<<b is the minimum number of buckets to allocate.
 // dirtyalloc should either be nil or a bucket array previously
 // allocated by makeBucketArray with the same t and b parameters.
-// If dirtyalloc is nil a new backing array will be alloced and
```
Sajjon, line 361: or is `alloced` to be read as: "alloc-ed", to have used `(m)alloc` on something? Still, `alloc` does indeed allocate memory, so `allocated` is correct, no?

src/cmd/internal/obj/loong64/obj.go
Zxilly: Please use Gerrit for comments and code review, but not GitHub.

timothy-king: https://go.dev/doc/contribute has contribution instructions.

PR #69021, opened 2024-08-22
flag: fix small error in docs

WofWca:
The flag type is string and not int.


WofWca:
Oh. Yeah. I misunderstood. This is embarrassing.

PR #69440, opened 2024-09-13
net/http/httptest: ensure that a superfluous WriteHeader call panics in httptest

dmathieu:
In net/http, using WriteHeader() twice creates a log entry, since the second one is superfluous:
https://cs.opensource.google/go/go/+/refs/tags/go1.23.1:src/net/http/server.go;l=1201

In net/http/httptest though, while a check is made, there is no way to detect the superfluous usage.
That makes superfluous calls reasy to reach production runtimes.
To help prevent that, this change proposes rejecting any superfluous WriteHeader calls in the httptest package.


dmathieu:
I considered logging first, but it appears nothing uses logs in that package (yet), and we do already have a `panic()` with invalid HTTP statuses.
I'm happy to change this to something less intrusive though.

PR #69856, opened 2024-10-12
timer: fix Stop link in Timer.Reset doc comment

pellared:


pellared:
Noticed when reading https://pkg.go.dev/time#Timer.Reset

PS. I also noticed improper links in https://pkg.go.dev/time#hdr-Monotonic_Clocks, but I would rather create a separate PR for it.

PR #69914, opened 2024-10-16
time: correct time.AppendText's error message

apocelipes:
"time.AppendText" returns error messages that start with the prefix
"time.MarshalText: " which seems confusion.

Now correct the message prefix to "time.AppendText: " and add a test
to prevent regression.


andrewmed-prop:
Just wondering why fn is called append, while it actually prepends text....

andrewmed-prop:
... and why it has 'Text' in name while it accepts and returns bytes, not strings...

PR #69943, opened 2024-10-20
net/http: initialize Value with File length in cloneMultipartForm

1911860538:
Improve the initialization of the Value map in cloneMultipartForm by
utilizing the length of the File map to optimize memory allocation.


cekrem (approved):
LGTM!

Issue #29310, opened 2018-12-17
encoding/json: omitempty for arrays is confusing


skriptble:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.11.2 darwin/amd64
</pre>

### Does this issue reproduce with the latest release?
Yes.

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary></details>

```bash
$ go env
GOARCH="amd64"
GOOS="darwin"
```

### What did you do?
https://play.golang.org/p/NlRBFTZHFYd
```go
func main() {
	type Foo struct {
		ID [16]byte `json:",omitempty"`
	}

	var f Foo
	j, _ := json.Marshal(f)
	fmt.Println(string(j))
}
```

### What did you expect to see?
```json
{}
```

### What did you see instead?
```json
{"ID":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}
```

This is caused by `isEmptyValue` inside `encoding/json` checking to see if the length of the array is 0. For people actually using an array (and not a slice), it is unlikely that they would ever use an array of zero length. It is likely, however, that they might want to omit the result if all of the values in the array are empty. For example, if a user is using `[16]byte` to represent a UUID, they might want to omit encoding it if it is all zeroes. This was brought up in #11939, specifically [this comment](https://github.com/golang/go/issues/11939#issuecomment-429184373). It would be nice if we could do this generically for arrays with values we can determine are empty.

crhntr:
An array that has a bunch of 'zero values' is still is not empty. The list of zeros means something. I would expect the array to be printed as is shown in the "what did you see instead" (for example a zero matrix). I played around with this a little and think this is a real issue.

My first attempt would be to implement the json.Marshaler interface for some type encapsulating [16]byte.
```go
type UUID [16]int

func (uuid UUID) MarshalJSON() ([]byte, error) {
	for _, num := range uuid {
		if num != 0 {
			return json.Marshal([10]int(uuid))
		}
	}
	return json.Marshal(/* some zero value */) // <--- this is the real problem in my opinion.
}

```
Since it seams MarshalJSON is called after the value is established to not be zero, no matter what you put return from MarshalJSON, the value returned is not considered to be an empty value.

I think a json.ErrValueIsZero would be useful in this situation. So you can override the zero-ness of a type.

[...]

PR #69984, opened 2024-10-22
encoding/json: clarify omitempty option for {array,slice,map,string}

callthingsoff:
This CL is inspired by:
https://github.com/golang/go/issues/29310#issuecomment-758768325

When I read omitempty option in encoding/xml package, I find it's
a bit different than encoding/json package.

I think it's more precise to say:
"any array, slice, map, or string of length zero."

Update #29310


Hadihoseinnejad1992 (changes requested):
Hadihoseinnejad92

PR #70266, opened 2024-11-09
go/printer: don't strip parens around *ast.Ellipsis in return field list

mateusz834:
Before this change:

func A() (...A)

was formatted into:

func A() ...A

which causes a parse error.

With this change it is going to be formatted as (no change):

func A() (...A)


djalil-code:
- [ ] ****

PR #70368, opened 2024-11-15
os/signal: check int type of signal using reflection

### LINKED ISSUE #70369: x/sys: windows can not handle ctrl+c (SIGINT) well
### Go version

go version go1.22.3 windows/amd64

```shell
set GO111MODULE=
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\qian\AppData\Local\go-build
set GOENV=C:\Users\qian\AppData\Roaming\go\env
set GOEXE=.exe
set GOEXPERIMENT=
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOINSECURE=
set GOMODCACHE=C:\Users\qian\go\pkg\mod
set GONOPROXY=
set GONOSUMDB=
set GOOS=windows
set GOPATH=C:\Users\qian\go
set GOPRIVATE=
set GOROOT=D:\Environments\go
set GOTMPDIR=
set GOTOOLCHAIN=auto
set GOTOOLDIR=D:\Environments\go\pkg\tool\windows_amd64
set GOVCS=
set GOVERSION=go1.22.3
set GCCGO=gccgo
set GOAMD64=v1
set AR=ar
set CC=gcc
set CXX=g++
set CGO_ENABLED=0
set GOMOD=NUL
set GOWORK=
set CGO_CFLAGS=-O2 -g
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-O2 -g
set CGO_FFLAGS=-O2 -g
set CGO_LDFLAGS=-O2 -g
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -fno-caret-diagnostics -Qunused-arguments -Wl,--no-gc-sections -fmessage-length=0 -ffile-prefix-map=C:\Users\qian\AppData\Local\Temp\go-build3528663502=/tmp/go-build -gno-record-gcc-switches
```

### What did you do?

When developing a server on Windows, we want to capture the SIGINT signal for extension purposes.
https://go.dev/play/p/wiNlThXDUPT

### What did you see happen?

signal.Notify(c chan<- os.Signal, sig ...os.Signal) cannot correctly capture windows.Signal.

signal.Notify can correctly handle windows.Signal. 
This means that when pressing Ctrl+C, the logic can be correctly triggered.

WineChord:
The current implementation of the signum function only checks if os.Signal is 
of type syscall.Signal, then converts it to int. However, golang.org/x/sys 
defines its own Signal type for Windows at 
https://cs.opensource.google/go/x/sys/+/refs/tags/v0.27.0:windows/syscall_windows.go;l=1472

This causes the signum function and the higher-level signal.Notify function 
to fail to recognize signals like windows.SIGINT defined in golang.org/x/sys.

One possible approach is to specifically handle windows.Signal in signum, 
but it's unreasonable to make go/src reference golang.org/x/sys.

Another approach is to make windows.Signal directly type alias to 
syscall.Signal in golang.org/x/sys, but windows.Signal is fundamentally 
different from syscall.Signal (the signals map is different), which is also 
not feasible.

Therefore, this PR attempts to use reflection to obtain the possible int 
type, to handle this special case for Windows.

Fixes #70369

### DIFF
--- a/src/os/signal/signal_unix.go
+++ b/src/os/signal/signal_unix.go
@@ -8,6 +8,7 @@ package signal
 
 import (
 	"os"
+	"reflect"
 	"syscall"
 )
 
@@ -41,7 +42,24 @@ func signum(sig os.Signal) int {
 		}
 		return i
 	default:
-		return -1
+		// Use reflection to determine if sig has an underlying integer type.
+		// In some systems like Windows, os.Signal may have implementations
+		// with underlying integer types that are not directly accessible,
+		// as they might be defined in external packages like golang.org/x/sys.
+		// Since we cannot import those platform-specific signal types,
+		// reflection allows us to handle them in a generic way.
+		kind := reflect.TypeOf(sig).Kind()
+		switch kind {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			// Extract the integer value from sig and validate it.
+			i := int(reflect.ValueOf(sig).Int())
+			if i < 0 || i >= numSig {
+				return -1
+			}
+			return i
+		default:
+			return -1
+		}
 	}
 }
 
--- a/src/os/signal/signal_unix.go
+++ b/src/os/signal/signal_unix.go
@@ -48,11 +48,11 @@ func signum(sig os.Signal) int {
 		// as they might be defined in external packages like golang.org/x/sys.
 		// Since we cannot import those platform-specific signal types,
 		// reflection allows us to handle them in a generic way.
-		kind := reflect.TypeOf(sig).Kind()
-		switch kind {
+		v := reflect.ValueOf(sig)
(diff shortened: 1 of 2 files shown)


src/os/signal/signal_unix.go
```diff
@@ -41,7 +42,24 @@ func signum(sig os.Signal) int {
 		}
 		return i
 	default:
-		return -1
+		// Use reflection to determine if sig has an underlying integer type.
+		// In some systems like Windows, os.Signal may have implementations
+		// with underlying integer types that are not directly accessible,
+		// as they might be defined in external packages like golang.org/x/sys.
+		// Since we cannot import those platform-specific signal types,
```
sandyskies, line 49: cannot -> can't

PR #70388, opened 2024-11-16
fmt: correct fmt doc for hexadecimal notation

codeimmortal:
the Gerrit change.


codeimmortal:
This is wrong pull request

PR #70400, opened 2024-11-17
debug/dwarf: fail on reading incorrect offset due to skipped compilation unit

### LINKED ISSUE #3861: dwarf/godwarf: Can't adjust rnglist offset when seeking past current CU
This issue is the root cause of https://github.com/parca-dev/parca/issues/5291.

### What's the issue

Calling `godwarf.LoadTree` with certain offsets in certain binaries can result in incorrect DWARF entries being read.

### How to reproduce

First, download and decompress ClickHouse binary:

```
curl -LO https://s3.amazonaws.com/clickhouse-builds/24.10/2e054d555719f509b96e0cfea4d5f5819c69a61f/package_release/clickhouse
chmod +x clickhouse
./clickhouse --version
```

Then, compile this:

```go
import (
	"debug/dwarf"
	"debug/elf"
	"fmt"

	"github.com/go-delve/delve/pkg/dwarf/godwarf"
)

func main() {
  path := "clickhouse"
  f, err := elf.Open(path)
  if err != nil {
    print(err)
    return
  }

  d, err := f.DWARF()
  if err != nil {
    print(err)
    return
  }

  // DWARF info reader
  reader := d.Reader()

   _, err = godwarf.LoadTree(0x139d6483, d, 0)
   if err != nil {
    fmt.Printf("errored: %v\n", err)
  }
```

You will get the following output:
```
errored: invalid rnglist offset 2858131463 (max 636512)
```

If, instead, you change the offset passed to `LoadTree` to `0x139d02d2` (the start of the compilation unit), the example works properly. Also, by print-debugging, I have checked that the offset that is read for the problematic entry matches the one outputted by `llvm-dwarfdump`.

### Why does this happen

According to [this comment](https://github.com/golang/go/blob/1340662ab4c56f946337518bda94f861478d2661/src/debug/dwarf/entry.go#L745) in `debug/dwarf`, `Reader.Next()` sets the current unit when it passes by a compilation unit DWARF entry, and then uses the offset of the CU to adjust the rnglist offset when encountering one. If `Seek()` is called directly to an offset that is after the compilation unit DWARF entry, that information is not set and the offset is incorrect.

### How to solve this

Probably, by finding the CU for an offset in the first call and then reusing the reader across recursive calls.

danipozo:
Don't return incorrect offsets silently on Reader.Next(). Modify Seek()
method to set the CU before reading entries to prevent this failure. Would
prevent and serve to fix go-delve/delve#3861 and parca-dev/parca#5291.

### DIFF
--- a/src/debug/dwarf/entry.go
+++ b/src/debug/dwarf/entry.go
@@ -541,6 +541,8 @@ func (b *buf) entry(cu *Entry, atab abbrevTable, ubase Offset, vers int) *Entry 
 			} else if a.tag == TagCompileUnit {
 				delay = append(delay, delayed{i, off, formAddrx})
 				break
+			} else {
+				b.error("Can't adjust offset: compilation unit skipped")
 			}
 
 			var err error
@@ -692,6 +694,8 @@ func (b *buf) entry(cu *Entry, atab abbrevTable, ubase Offset, vers int) *Entry 
 			} else if a.tag == TagCompileUnit {
 				delay = append(delay, delayed{i, off, formStrx})
 				break
+			} else {
+				b.error("Can't adjust offset: compilation unit skipped")
 			}
 
 			val = resolveStrx(uint64(strBase), off)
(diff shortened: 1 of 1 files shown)


aarzilli:
Shouldn't this be what Seek always does?

danipozo:
Closing as it was fixed in https://go-review.googlesource.com/c/go/+/655976

PR #70520, opened 2024-11-22
text/template: handle UnsafePointer in isTrue

lxl-renren:


lxl-renren:
add test cases. the original code run test will failed on reflect.UnsafePointer non nil value

PR #70642, opened 2024-12-02
crypto/x509: support certificates with X25519 public keys

EdSchouten:
Even though X25519 can't be used for signing, it may be desirable to use
a CA structure to sign X25519 keys that are used for ECDH purposes. This
change adds the ability to properly extract X25519 public keys from
certificates that contain such keys. It also makes it possible to create
such certificates through CreateCertificate().

### DIFF
--- a/src/crypto/x509/x509.go
+++ b/src/crypto/x509/x509.go
@@ -267,6 +267,7 @@ const (
 	RSA
 	DSA // Only supported for parsing.
 	ECDSA
+	X25519
 	Ed25519
 )
 
@@ -274,6 +275,7 @@ var publicKeyAlgoName = [...]string{
 	RSA:     "RSA",
 	DSA:     "DSA",
 	ECDSA:   "ECDSA",
+	X25519:  "X25519",
 	Ed25519: "Ed25519",
 }
 
@@ -505,6 +507,8 @@ func getPublicKeyAlgorithmFromOID(oid asn1.ObjectIdentifier) PublicKeyAlgorithm 
 		return DSA
 	case oid.Equal(oidPublicKeyECDSA):
 		return ECDSA
+	case oid.Equal(oidPublicKeyX25519):
+		return X25519
 	case oid.Equal(oidPublicKeyEd25519):
 		return Ed25519
 	}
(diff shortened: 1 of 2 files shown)


ssinjinx:
## Code Review: X25519 Certificate Support

**PR:** #70642 — support certificates with X25519 public keys
**Files:** `src/crypto/x509/x509.go` (+7/-3), `src/crypto/x509/x509_test.go` (+61/-0)
**Reviewed by:** Jeramey Schaffner — AI-assisted code review

---

### ✅ Summary

Clean, well-scoped addition. The parser and marshal paths already supported X25519 keys; this PR closes the loop by exposing the `X25519` public key algorithm constant and recognizing the X25519 OID in `getPublicKeyAlgorithmFromOID()`. The test uses a real IETF test certificate, which is great for correctness.

---

### 🔴 Bug: Incorrect error variable in test (src/crypto/x509/x509_test.go:1166)

```go
_, ok := cert.PublicKey.(*ecdh.PublicKey)
if !ok {
    t.Fatalf("Parsed key was not an X25519 key: %s", err)
}
```

When the type assertion fails (`ok == false`), `err` is `nil` — it was discarded from `pem.Decode` with `_`, and `der == nil` was already handled above. The error message would print `%!s(<nil>)` which is misleading.

**Fix:**
```go
if !ok {
    t.Fatalf("Parsed key was not an X25519 key, got: %T", cert.PublicKey)
}
```

---

### ⚠️ Suggestions

**1. Verify key bytes in test (nice to have)**

The test checks the algorithm enum and the Go type, but doesn't verify the actual key bytes match the certificate. Consider adding:
```go
pub := cert.PublicKey.(*ecdh.PublicKey)
expected := []byte{0x85, 0x20, 0xf0, 0x09, /* ... remaining bytes from cert ... */}
if !bytes.Equal(pub.Bytes(), expected) {
    t.Fatalf("Parsed key bytes don't match certificate")
}
```
This catches cases where the parser returns the wrong key data with the right type — rare but worth the 3 extra lines for a crypto test.

**2. Test CreateCertificate with X25519 keys**

The PR description says "It also makes it possible to create such certificates through CreateCertificate()." While `marshalPublicKey` already handles `*ecdh.PublicKey`, there's no test covering the full `CreateCertificate` → `ParseCertificate` round-trip with an X25519 key. A round-trip test would catch regressions in either direction.

**3. Test ParsePKIXPublicKey with X25519**

Not blocking — `ParseCertificate` exercises the same `parsePublicKey` path internally. But a direct `ParsePKIXPublicKey` test on the SubjectPublicKeyInfo from this certificate would be belt-and-suspenders.

---

### ✅ Looks Good

- **Minimal diff** — exactly the lines needed, no scope creep
- **Correct enum placement** — `X25519` between `ECDSA` and `Ed25519` follows the const ordering
- **Real certificate in test** — using an IETF test demo cert instead of a hand-rolled one is the right call for crypto tests
- **Parser already existed** — `parsePublicKey` in `parser.go:283` already maps `oidPublicKeyX25519` → `ecdh.X25519().NewPublicKey(der)` with proper parameter validation per RFC 8410
- **Marshal path already existed** — `marshalPublicKey` handles `*ecdh.PublicKey` with X25519 curve detection via `oidFromECDHCurve()`
- **RFC compliance** — parameters MUST be absent per RFC 8410 Section 3, and the parser enforces this at parser.go:286-287

---

### Verdict: **Approve with minor fix**

The one-line test bug should be fixed before merge (incorrect error message in a test is still a bug), but the feature itself is solid. The suggestions are enhancements, not blockers.

---

*Reviewed by Jeramey Schaffner — [AI-assisted PR review service available](https://github.com/ssinjinx)*

PR #70841, opened 2024-12-14
cmd/internal/dwarf: fix missing decl_file tag for abstract subprograms

piob-io:
Abstract subprograms are declared at specific file:line just as regular
subprograms. Line annotation was already added, but as todo notes,
not the file.


piob-io:
This won't work, see Than's comment on https://go-review.googlesource.com/c/go/+/636096
TLDR: Abstract subprograms can currently be emitted more than once, from different packages, and every time they must contain exact bytes. But file index reference will depend on interned file list for specific package.

PR #70933, opened 2024-12-19
mime/multipart: add field Reader.MaxMIMEHeaderSize

### LINKED ISSUE #26339: mime/multipart: Allow limiting maximum ammount of part's header data
This is feature request, to allow user of `mime/multipart` package's `Reader` to specify limit of header data loaded into memory when calling `NewPart()` function.
`Reader.NewPart()` calls `net/textproto` package's `Reader.ReadMIMEHeader()` function which without any limitation loads all header data into memory.
`mime/multipart`'s `Reader` is also used in `net/http` package to parse `multipart/form-data` POST requests, which means that functions like `Request.ParseMultipartForm`, `Request.FormValue`, `Request.FormFile`, `Request.MultipartReader` are affected by this.
I believe this can be used as remote denial of service attack vector.
Possible solution for this would be limiting HTTP request body size (done by default in some servers like nginx), but this is unsuitable for cases where server needs to accept POST requests with big files (for example, file hosting service).
Similar case like this is `net/http` package's `Server`'s `MaxHeaderBytes` field, which limits ammount of header data, but not body.

### LINKED ISSUE #73087: mime/multipart: add field Reader.MaxMIMEHeaderSize
### Proposal Details

Please provide a way to limit the maximum size of the header when using `http.ParseMultipartForm`.

Currently package `mime/multipart` has a hardcoded constant [maxMIMEHeaderSize](https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/mime/multipart/multipart.go;l=346) equal to 10 megabytes. This means that an app having multipart HTTP endpoints should allocate at least 10 megabytes of RAM per request, which is quite high. Most headers are much smaller and allowing 10 megabytes for the header is not useful, just exposes an app to memory exhaustion by many parallel slow requests with large MIME headers.

I propose to add a public field **MaxMIMEHeaderSize** to multipart.Reader.
Methods `NextPart()` and `NextRawPart()` should use the value of that field if it is not 0, otherwise they should use the constant.

To use this new field, one would call request.MultipartReader() method in `http` package, set the field to the desired value (e.g. 1024) and then parse the form using `reader.NextPart()` method without the risk that multipart headers eat 10 megabytes of RAM.

starius:
If the field is set, it is used instead of maxMIMEHeaderSize
constant, allowing to further constraint memory usage when
parsing multipart streams.

Fixes #68889
Fixes #73087
Fixes #26339

### DIFF
--- a/src/mime/multipart/multipart.go
+++ b/src/mime/multipart/multipart.go
@@ -331,6 +331,11 @@ func (p *Part) Close() error {
 // Reader's underlying parser consumes its input as needed. Seeking
 // isn't supported.
 type Reader struct {
+	// MaxMIMEHeaderSize is the maximum size of a MIME header we will parse,
+	// including header keys, values, and map overhead. If not set, then the
+	// default value of 10 << 20 is used.
+	MaxMIMEHeaderSize int64
+
 	bufReader *bufio.Reader
 	tempDir   string // used in tests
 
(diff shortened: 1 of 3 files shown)


sdavids:
This constant should be mentioned in the doc of `ParseMultipartForm`.

starius:
> This constant should be mentioned in the doc of `ParseMultipartForm`.

I updated the PR. Added notes to methods request.MultipartReader and request.ParseMultipartForm.

PR #71233, opened 2025-01-11
feat(cert_pool): add AppendCertsFromFile func

MohsenKasraeifar:
This request adds a function that appends certificates from the file.


seankhliao:
see https://go.dev/s/proposal

PR #71475, opened 2025-01-29
encoding/json: improve decoder alloc count

oiweiwei:
the allocs per op is twice as more for json.Decoder compared to json.Unmarshal to interface{}, while
the reason for excessive allocation is SyntaxError read-ahead generation (that is subsequently
discarded) leaving the memory imprint.

### DIFF
--- a/src/encoding/json/decode_test.go
+++ b/src/encoding/json/decode_test.go
@@ -467,21 +467,21 @@ var unmarshalTests = []struct {
 	{CaseName: Name(""), in: `{"alphabet": "xyz"}`, ptr: new(U), err: fmt.Errorf("json: unknown field \"alphabet\""), disallowUnknownFields: true},
 
 	// syntax errors
-	{CaseName: Name(""), in: `{"X": "foo", "Y"}`, err: &SyntaxError{"invalid character '}' after object key", 17}},
-	{CaseName: Name(""), in: `[1, 2, 3+]`, err: &SyntaxError{"invalid character '+' after array element", 9}},
-	{CaseName: Name(""), in: `{"X":12x}`, err: &SyntaxError{"invalid character 'x' after object key:value pair", 8}, useNumber: true},
+	{CaseName: Name(""), in: `{"X": "foo", "Y"}`, err: &SyntaxError{invalidChar: '}', invalidCharContext: "after object key", Offset: 17}},
+	{CaseName: Name(""), in: `[1, 2, 3+]`, err: &SyntaxError{invalidChar: '+', invalidCharContext: "after array element", Offset: 9}},
+	{CaseName: Name(""), in: `{"X":12x}`, err: &SyntaxError{invalidChar: 'x', invalidCharContext: "after object key:value pair", Offset: 8}, useNumber: true},
 	{CaseName: Name(""), in: `[2, 3`, err: &SyntaxError{msg: "unexpected end of JSON input", Offset: 5}},
-	{CaseName: Name(""), in: `{"F3": -}`, ptr: new(V), err: &SyntaxError{msg: "invalid character '}' in numeric literal", Offset: 9}},
+	{CaseName: Name(""), in: `{"F3": -}`, ptr: new(V), err: &SyntaxError{invalidChar: '}', invalidCharContext: "in numeric literal", Offset: 9}},
 
 	// raw value errors
-	{CaseName: Name(""), in: "\x01 42", err: &SyntaxError{"invalid character '\\x01' looking for beginning of value", 1}},
-	{CaseName: Name(""), in: " 42 \x01", err: &SyntaxError{"invalid character '\\x01' after top-level value", 5}},
-	{CaseName: Name(""), in: "\x01 true", err: &SyntaxError{"invalid character '\\x01' looking for beginning of value", 1}},
-	{CaseName: Name(""), in: " false \x01", err: &SyntaxError{"invalid character '\\x01' after top-level value", 8}},
-	{CaseName: Name(""), in: "\x01 1.2", err: &SyntaxError{"invalid character '\\x01' looking for beginning of value", 1}},
-	{CaseName: Name(""), in: " 3.4 \x01", err: &SyntaxError{"invalid character '\\x01' after top-level value", 6}},
-	{CaseName: Name(""), in: "\x01 \"string\"", err: &SyntaxError{"invalid character '\\x01' looking for beginning of value", 1}},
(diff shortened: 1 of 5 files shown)


oiweiwei:
# Summary

when working with the json.Decoder implementation, I've noticed that the difference between using decoder and json.Unmarshal to `any` type seems to be not in Decoder favor. The allocs per op as per benchmarking bellow is twice as more for Decoder based implementation, while the reason for excessive allocation is SyntaxError read-ahead generation (that is subsequently discarded) leaving the memory imprint.

# Solution

Delay the memory-affecting operations until they are really needed.

# Details

## Code to reproduce

 I've written simple (key, value) enumerator as following:

```go
package jsonkv

import (
	"fmt"
	"io"

	// "github.com/golang/go/src/encoding/json"
	"encoding/json"
)

type scope []bool

func (s *scope) array() bool {
	return len(*s) > 0 && (*s)[len(*s)-1]
}

func (s *scope) consume(tok json.Token) bool {
	switch tok := tok.(type) {
	case json.Delim:
		switch tok {
		case '{':
			(*s) = append(*s, false)
		case '[':
			(*s) = append(*s, true)
		case '}', ']':
			(*s) = (*s)[:len(*s)-1]
		}
	default:
		return false
	}
	return true
}

type KeyValuer struct {
	scope scope
	dec   *json.Decoder
	key   string
}

func (kv *KeyValuer) flush() string {
	key := kv.key
	kv.key = ""
	return key
}

func NewKeyValuer(r io.Reader) *KeyValuer {
	return &KeyValuer{dec: json.NewDecoder(r), scope: make(scope, 0)}
}

func (kv *KeyValuer) Reset(r io.Reader) {
	kv.dec = json.NewDecoder(r)
	if kv.scope != nil {
		kv.scope = kv.scope[:0]
	}
	kv.key = ""
}

// Range reads key and value pairs.
func (kv *KeyValuer) Range(f func(string, any) bool) error {
	for {
		key, val, err := kv.next()
		if err == io.EOF {
			break
		}
		if err != nil {
			return err
		}
		if !f(key, val) {
			break
		}
	}
	return nil
}

// First reads first occurence of the key.
func (kv *KeyValuer) First(key string) (any, error) {
	var val any
	err := kv.Range(func(k string, v any) bool {
		if k == key {
			val = v
			return false
		}
		return true
	})
	return val, err
}

// Next reads next key and value.
func (kv *KeyValuer) Next() (string, any, error) {
	return kv.next()
}

// next reads next key and value.
func (kv *KeyValuer) next() (string, any, error) {

	for {

		// read next token.
		tok, err := kv.dec.Token()
		if err != nil {
			return "", nil, err
		}

		// if delimiter is consumed or current scope is array, skip.
		if kv.scope.consume(tok) || kv.scope.array() {
			kv.flush() // clear key.
			continue
		}

		// if key is empty, set key.
		if kv.key == "" {
			if tok, ok := (any)(tok).(string); !ok {
				return "", nil, fmt.Errorf("unexpected token: %v", tok)
			} else {
				kv.key = tok
			}
			continue
		}

		// unwrap json.Number to string.
		if num, ok := tok.(json.Number); ok {
			return kv.flush(), string(num), nil
		}

		// return key and value.
		return kv.flush(), tok, nil
	}
}
```

## Benchmark

```go
package jsonkv

import (
	"encoding/json"
	"fmt"
	"strings"
	"testing"
)

var j = `[{"obj": {"name": "test1", "age": 20, "arr": ["a", "b", {"a": "c"}]}}, {"obj": {"name": "test2", "age": 30}}]`

func TestKeyValuerFirst(t *testing.T) {

	kv := NewKeyValuer(strings.NewReader(j))
	val, err := kv.First("a")
	if err != nil {
		t.Fatalf("error: %v", err)
	}
	if val != "c" {
		t.Fatalf("expected c, got %v", val)
	}
}

func BenchmarkKeyValuerFirst(b *testing.B) {
	r := strings.NewReader("")
	kv := NewKeyValuer(r)
	for i := 0; i < b.N; i++ {
		r.Reset(j)
		kv.Reset(r)
		if _, err := kv.First("a"); err != nil {
			b.Fatalf("error: %v", err)
		}
	}
}

func JSONUnmarshal(key string, b []byte) (any, error) {

	var v any

	err := json.Unmarshal(b, &v)
	if err != nil {
		return nil, err
	}

	varr, _ := v.([]any)
	for _, v := range varr {
		obj, ok := v.(map[string]any)
		if !ok {
			continue
		}
		objMap, ok := obj["obj"].(map[string]any)
		if !ok {
			continue
		}
		objArr, ok := objMap["arr"].([]any)
		if !ok {
			continue
		}
		for _, v := range objArr {
			obj, ok := v.(map[string]any)
			if !ok {
				continue
			}
			val, ok := obj[key]
			if !ok {
				continue
			}
			return val, nil
		}
	}

	return nil, fmt.Errorf("not found")
}

func BenchmarkUnmarshalJSON(b *testing.B) {

	bj := []byte(j)
	for i := 0; i < b.N; i++ {
		_, err := JSONUnmarshal("a", bj)
		if err != nil {
			b.Fatalf("error: %v", err)
		}
	}
}
```

## Benchmark Results

benchmark results where quite surprising for me:

```
BenchmarkKeyValuerFirst-4   	  142459	      7541 ns/op	    2176 B/op	      82 allocs/op
BenchmarkUnmarshalJSON-4    	  236854	      5666 ns/op	    2320 B/op	      41 allocs/op
```

profile revealed:

```
Showing top 10 nodes out of 114
      flat  flat%   sum%        cum   cum%
     440ms 15.33% 15.33%      960ms 33.45%  runtime.mallocgc
     240ms  8.36% 23.69%      560ms 19.51%  runtime.concatstrings
```

when looking further into graph, it seems like the culprit code responsible for frequent call of concatstrings is error handling in decoder:

image

it turned out that error is emitted for every delimiter like ":", "," and so on. so, we need to delay the execution of error until the actual Error is being called.

## Benchmark(-etts) After Fix

Applying these changes gives a signinficant benefit (x2) in allocations for decoder:

```
BenchmarkKeyValuerFirst-4   	  196369	      5413 ns/op	    1776 B/op	      42 allocs/op
BenchmarkUnmarshalJSON-4    	  215534	      5745 ns/op	    2320 B/op	      41 allocs/op
```

same shows profiler:

```
Showing top 10 nodes out of 133
      flat  flat%   sum%        cum   cum%
     600ms 18.07% 18.07%     1300ms 39.16%  runtime.mallocgc
     190ms  5.72% 23.80%      810ms 24.40%  github.com/golang/go/src/encoding/json.(*Decoder).readValue
```

PR #71618, opened 2025-02-07
image/jpeg: improve decoder performance by ~12%

romshark:
improve JPEG decoder performance by about 10-14% on most
architectures by unrolling unzig and shift-clamp loops.
The trade-off is ~16KiB additional binary size.


romshark:
The benchmark results and experimentation playground is here
https://github.com/romshark/jpegbench

EliCDavis:
A potential improvement if the dst array is 0'd out

```go
	writeDst := func(index int) {
		c := (*b)[index]
		if c < -127 {
			return
		}

		c += 128
		if c > 255 {
			c = 255
		}
		dst[(index/8)*stride+(index%8)] = uint8(c)
	}
```

wxiaoguang:
It seems to be caused by inline optimization https://godbolt.org/z/fP741vfoo

The `writeDst` is inline-ed and compiler generates a lot of duplicate code including "panic" calls, that's why it bloats 16KiB ....


No idea why the `writeDst` is faster than the `loop` version, is it true for all CPU architecture? Maybe such case should be handled by compiler optimization and/or assemble code?

Issue #43126, opened 2020-12-10
encoding/json: UnmarshalTypeError.Error message seems insensible


dsnet:
[Consider this snippet](https://play.golang.org/p/lKh1SHb-9VT):
```go
type GoStruct1 struct {
	AlphaField GoStruct2 `json:"alpha_field"`
}
type GoStruct2 struct {
	BravoField []GoStruct3 `json:"bravo_field"`
}
type GoStruct3 struct {
	StringField string `json:"string_field"`
}

func main() {
	fmt.Println(json.Unmarshal([]byte(`{"alpha_field":{"bravo_field":[{"string_field": 3.14159}]}}`), new(GoStruct1)))
}
```

On Go1.15, this prints:
```
json: cannot unmarshal number into Go struct field GoStruct3.alpha_field.bravo_field.string_field of type string
```
Notice that it combines two different namespaces where it is:
```
{{GoTypeName}}.{{JSONPath}}
```
and that the message says that it is a "Go struct field" when it isn't.
Also, it uses `GoStruct3`, when it makes little sense.
At minimum, it should at least used `GoStruct1`.

\cc @mvdan

dsnet:
Related considerations:
* Should the JSON path include array indexes?
* Should the JSON path disambiguate the path if dots appear in the object name? [Example](https://play.golang.org/p/ZkMvETBcFHa).

Strawman proposal: the JSON path should be a valid expression in JavaScript to address the specific value from the root value.
* For objects, we use `.foo` if the object name is a valid JavaScript identifier, otherwise we use `["fizz.buzz"]` if it isn't.
* For arrays, we use `[5]`.

mvdan:
I agree with the strawman proposal. In that proposal, what would you use in place of GoStruct3?

Also, is there a "json path" spec we can simply adhere to? or do we simply try to get as close to a JS expression as we reasonably can?

cc @cuonglm who worked on the original feature. I did think about proposing to make it more consistent with slices and such, but at the end of the day I thought the addition was already a big step forward.

dsnet:
> Also, is there a "json path" spec we can simply adhere to? 

[...]

PR #71638, opened 2025-02-10
encoding/json: add context during decoding when disallowUnknownFields is set

machine424:
With most efforts redirected towards encoding/json v2. I don't think
refactoring/cleaning changes in #43126 will ever be addressed in v1.

I don't think we should consider it as a requirement for this patch.

This could have been added as part of https://github.com/golang/go/commit/2596a0c075aeddec571cd658f748ac7a712a2b69

encoding/json v2 is moving in the same direction by adding context by
default as shown in https://github.com/go-json-experiment/json/blob/4e0381018ad68adc9c0cb7896d109c994429654e/errors.go#L364-L373

But we don't need to wait for v2 to become stable to improve UX
and save users some time, and help them keep their hair!

Fixes #58649

the Gerrit change.

  + No trailing period
+ No Markdown


machine424:
`encoding/json` is frozen for such changes.
See discussion here https://go-review.googlesource.com/c/go/+/648055#message-c41900eb8f058832f1e30fa679a6c8a35dd50159

PR #71723, opened 2025-02-13
debug/buildinfo: add OpenVEX statement for test binary

LaurentGoderre:
Adds a OpenVex for the buildinfo test binary to prevent false positive from vulnerability scanners


LaurentGoderre:
This was meant to address the 4 critical 43 high and nearly 20 medium CVEs here:

https://hub.docker.com/layers/library/golang/1.24/images/sha256-b95002399f27188790f0a2b598bee84ae7bfbf1043fb60921da3b81928e303ba

seankhliao:
Go should not need to litter its codebase with files for external scanners.

prattmic:
Please open a general issue to discuss what we should do about this. I tend to agree that we shouldn't need files for various third party tools in our source tree, but I'm not very familiar with the area, so normal issue can include a wider audience for discussion.

PR #72047, opened 2025-03-01
cmd/go: clear GOROOT for func ldShared when -trimpath is used

hongxu-jia:
In commit [cmd: remove support for GOROOT_FINAL][1], it clear GOROOT for func ld when -trimpath is used. [2]

This commit do the same thing for func ldShared, otherwise run go_bootstrap with -trimpath does not remove build path directory from the generated share libarary

  $ ./make.bash --target-only --no-banner std
  $ go_bootstrap install -linkshared -buildmode=shared -trimpath std
  $ grep "$(pwd)" ./pkg/linux_amd64_dynlink/libstd.so
  Binary file ./pkg/linux_amd64_dynlink/libstd.so matches

[1] https://github.com/golang/go/commit/507d1b22f4b58ac68841582d0c2c0ab6b20e5a98
[2] https://github.com/golang/go/commit/507d1b22f4b58ac68841582d0c2c0ab6b20e5a98#diff-cab5921f94f2667bb0bc1b935d2d46b4c03541b4351b33438ab7290b94dea212R669

the Gerrit change.

  + No trailing period
+ No Markdown

### DIFF
--- a/src/cmd/go/internal/work/gc.go
+++ b/src/cmd/go/internal/work/gc.go
@@ -728,7 +728,14 @@ func (gcToolchain) ldShared(b *Builder, root *Action, toplevelactions []*Action,
 	// the output file path is recorded in the .gnu.version_d section.
 	dir, targetPath := filepath.Split(targetPath)
 
-	return b.Shell(root).run(dir, targetPath, nil, cfg.BuildToolexec, base.Tool("link"), "-o", targetPath, "-importcfg", importcfg, ldflags)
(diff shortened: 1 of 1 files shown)


hongxu-jia:
Submitted https://go-review.googlesource.com/c/go/+/653895 for review

hongxu-jia:
Duplicated with https://go-review.googlesource.com/c/go/+/653895, close this one

PR #73005, opened 2025-03-22
slices: add PushBack, PopBack, PopFront, Front, and Back functions

rishabhkhemka:
This change modifies Go slices package to support common slice operations by adding PushBack, PopBack, PopFront, Front, and Back functions. These functions introduce stack and queue behavior to slices, thereby improving code readability and package consistency.

### DIFF
--- a/src/slices/slices.go
+++ b/src/slices/slices.go
@@ -517,3 +517,44 @@ func Repeat[S ~[]E, E any](x S, count int) S {
 	}
 	return newslice
 }
+
+// PushBack adds an element to the end of a slice and returns the modified slice.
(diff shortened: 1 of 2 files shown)


rishabhkhemka:
@robpike @griesemer @ianlancetaylor
is it too much to ask for a push pop helper in slices? 
it's another thing if there are plans to introduce fundamental data structures in the container package

ever since 1.18 it's going a lot sideways when it comes to readability
google is so much about ds/algo while hiring..yet go fails to provide a decent container/algo package like other prominent languages
for a language with the best std library, this is a travesty
with utmost respect and love for the language.. please look into it

PR #73066, opened 2025-03-27
log/slog: log and logAttrs initialize ctx at top

wangzhione:
In extreme cases (e.g., ctx = nil), it is recommended to initialize the
context only once at the entry point before using log and logAttrs.


April-lies:
There's this same code in Enabled, so why bring it to the top instead of just deleting it?
```
func (l *Logger) Enabled(ctx context.Context, level Level) bool {
	if ctx == nil {
		ctx = context.Background()
	}
	return l.Handler().Enabled(ctx, level)
}
```

PR #73313, opened 2025-04-10
runtime: use built-in min function

mrclmr:


mrclmr:
I'm not sure if `maxGs` is a correct name for the function parameter.

PR #73477, opened 2025-04-22
maps: handle copy with nil maps

afiune:
By default, range will cover the case where src is nil, but if dst is nil, then Copy will throw a panic like:

panic: assignment to entry in nil map


kishorpau (approved):

Jorropo:
This has been [closed on gerrit](https://go-review.googlesource.com/c/go/+/667395), but for some reason the bot hasn't updated the status here.

PR #73889, opened 2025-05-27
wrong initialization cycle error fix 3

runsys:
wrong initialization cycle error fix 3


seankhliao:
As we've noted before, this is not considered a bug.
spamming PRs will not change that.

PR #74124, opened 2025-06-12
json/decoder: Allow empty strings for parsing numbers

KitiHey:
Empty strings in a json are considered like nonexistent. But when decoding to, for example, a struct with an int or float field and tries to decode an empty string, instead of marking it as 0 (which is nonexistent but for numbers), it would raise an error.
The code I want to merge tries to solve that. Not sure if this works though, but I would be glad if considering this change.


Analistarjx (commented):
Requisitos de mudanças

PR #74332, opened 2025-06-22
os: trivial comment fix

calmh:
"Geese" here looks like an autocorrect-o of "oses", I think writing it out
makes more sense.


nord-erik:
it's a pun on switch runtime.GOOS

PR #74333, opened 2025-06-23
sort: clarify Less doc

### LINKED ISSUE #73420: sort: clarify Slice comparator requirements
The doc for `sort.Slice` [states](https://github.com/golang/go/blob/7b263895f7dbe81ddd7c0fc399e6a9ae6fe2f5bf/src/sort/slice.go#L19):

```go
// The less function must satisfy the same requirements as
// the Interface type's Less method.
```

This is referring to the less function interface [here](https://github.com/golang/go/blob/7b263895f7dbe81ddd7c0fc399e6a9ae6fe2f5bf/src/sort/sort.go#L29):

```go
// Less must describe a transitive ordering:
//  - if both Less(i, j) and Less(j, k) are true, then Less(i, k) must be true as well.
//  - if both Less(i, j) and Less(j, k) are false, then Less(i, k) must be false as well.
```

This describes only transitive ordering. In contrast, the underlying algorithm, pdqsort, [requires](https://github.com/orlp/pdqsort/issues/28) strict weak ordering.

This is an important distinction. If a comparator is, for example:

 - reflexive
 - transitive

Then it will be conforming to the documented requirements yet will still  cause crashes under nebulous implementation defined circumstances. e.g. the number of elements to sort is greater than 33, and elements happen to be in a specific order.

To fix:

Change the documented requirement of `sort.Slice` to what is [stated](https://github.com/golang/go/blob/7b263895f7dbe81ddd7c0fc399e6a9ae6fe2f5bf/src/slices/sort.go#L30) in `sort.SortFunc`.
```go
// SortFunc sorts the slice x in ascending order as determined by the cmp
// function. This sort is not guaranteed to be stable.
// cmp(a, b) should return a negative number when a < b, a positive number when
// a > b and zero when a == b or a and b are incomparable in the sense of
// a strict weak ordering.
//
// SortFunc requires that cmp is a strict weak ordering.
// See https://en.wikipedia.org/wiki/Weak_ordering#Strict_weak_orderings.
// The function should return 0 for incomparable items.
```

jagprog5:
clarifies the requirements for Less

Fixes https://github.com/golang/go/issues/73420

### DIFF
--- a/src/sort/sort.go
+++ b/src/sort/sort.go
@@ -26,13 +26,15 @@ type Interface interface {
 	// Sort may place equal elements in any order in the final result,
 	// while Stable preserves the original input order of equal elements.
 	//
-	// Less must describe a transitive ordering:
+	// Less must describe a [Strict Weak Ordering]. For example:
 	//  - if both Less(i, j) and Less(j, k) are true, then Less(i, k) must be true as well.
 	//  - if both Less(i, j) and Less(j, k) are false, then Less(i, k) must be false as well.
 	//
 	// Note that floating-point comparison (the < operator on float32 or float64 values)
-	// is not a transitive ordering when not-a-number (NaN) values are involved.
+	// is not a strict weak ordering when not-a-number (NaN) values are involved.
 	// See Float64Slice.Less for a correct implementation for floating-point values.
+	//
+	// [Strict Weak Ordering]: https://en.wikipedia.org/wiki/Weak_ordering#Strict_weak_orderings
 	Less(i, j int) bool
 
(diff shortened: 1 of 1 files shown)


jagprog5:
Original PR: https://github.com/golang/go/pull/73421

I accidentally pushed a shallow history. This automatically killed the previous PR and abandoned the [Gerrit](https://go-review.googlesource.com/c/go/+/666355#message-16ca65ff5aa3073364d5b53ba570277945c6d0ef). I won't make that mistake again :). I have addressed the feedback given by the reviewer.

PR #74372, opened 2025-06-25
sort: add `SearchWithError`

zhiqiangxu:
This pr adds a `SearchWithError`, which is like `Search` but allows f to return an error, and exit earlier on error.


manju085ruckus:
lgtm

Issue #74473, opened 2025-07-04
cmd/compile: loopbce cares about the order of successors in the loop header


Jorropo:
https://github.com/golang/go/blob/6c3b5a2798c83d583cb37dba9f39c47300d19f1f/src/cmd/compile/internal/ssa/loopbce.go#L175-L182

This code here assumes the the first successor leads to the loop and that second successor break the loop.
it looks like ssagen always generate the `for` keyword this way but if you write a loop using `if ... { break }` inside `for { ... }` this is usually inverted.

PR #74479, opened 2025-07-05
fix loopbce bug

Raezil:
fixing https://github.com/golang/go/issues/74473


Jorropo:
#74473 has been fixed

PR #74740, opened 2025-07-24
crypto/internal/boring: make dst allocation for AES-GCM more efficient

juergw:
When the Seal or the Open function for AES-GCM are called with a dst of
insufficient size, the implementation currently calls the append function
in a loop until dst has enough capacity. This results in many unnecessary
allocations and copy operations.

For example, in this call with a a plaintext of size 1MB:

aeadCipher.Seal(nil, nonce, plaintext, associatedData)

we will currently make 33 allocations.

After this change, we only make one allocation.


juergw:
I think this will not be merged, so I close it.

PR #75048, opened 2025-08-16
cmd/link,runtime: fix c-shared dlopen on non-glibc systems (#13492)

### LINKED ISSUE #13492: runtime: c-shared builds fail with musllibc

### LINKED ISSUE #15336: x/net/http2: handle CONNECT request
As h2 spec : http://httpwg.org/specs/rfc7231.html#CONNECT

when we receive a CONNECT request, we should open a tcp connection to the target server identified in the :authority..and response to sender an 2XX, then pipe client connection and target server connection..

But in current h2: we can't response sender with and only with 2xx code, `WriteHeader(code int)` alway take content-length and other headers in and it causes some problems( http://httpwg.org/specs/rfc7231.html#CONNECT) and let client don't send any more data

Except that, h2-server always send 2xx and some headers to client after `http.Handler` finish its execution..and how can we forbidden this? because in 'proxy' situation, data will be exchanged between client and proxied server, proxy should not generate a 'response' by itself...(except previous response 2xx code to client)

Is there a recommend way to send client only one `HeaderFrame` with `:status: 200` and forbidden 'after handle done response(just like http1.x `w.conn.hijacked()`)'?

### LINKED ISSUE #54805: runtime: c-shared builds fail to load with dlopen with musllibc: `initial-exec TLS resolves to dynamic definition`
Splitting out from #13492:

@ansiwen's [comment](https://github.com/golang/go/issues/13492#issuecomment-1002224721):

> For what it's worth, compiled with -buildmode=c-shared I get:
> 
> ```
> sc_dlopen failed: Error relocating ./foobar.so: (null): initial-exec TLS resolves to dynamic definition in ./foobar.so
> ```
> 
> with
> 
> ```
> $ go version
> go version go1.17.5 linux/amd64
> $ cat /etc/alpine-release 
> 3.15.0
> $ apk version musl
> Installed:                                Available:
> musl-1.2.2-r7                           = 1.2.2-r7 
> ```

I am still seeing a similar issue today with Go 1.19.

jgowdy:
This PR fixes long-standing compatibility issues preventing Go c-shared
  and c-archive libraries from working correctly on non-glibc systems,
  particularly Alpine Linux (musl libc) and other alternative libc
  implementations.

  Problem

  Go's c-shared/c-archive build modes make glibc-specific assumptions that
  cause failures on other systems:

  1. SIGSEGV on dlopen(): Go assumes DT_INIT_ARRAY functions receive
     (argc, argv, envp) arguments, but this is glibc-specific behavior not
     guaranteed by the ELF specification. On musl and other systems, these
     pointers are null or garbage, causing immediate segfaults when Go
     tries to dereference argv in runtime.sysargs().

  2. TLS incompatibility: Go uses Initial Exec TLS model which requires
     static TLS allocation. This prevents dlopen() from working on musl and
     other dynamic loaders that don't reserve extra TLS space. The error
     "initial-exec TLS resolves to dynamic definition" makes Go shared
     libraries impossible to load dynamically.

  Solution

  This PR implements three key fixes:

  1. argc/argv null safety: Add null checks in runtime initialization
     to handle systems where DT_INIT_ARRAY functions don't receive arguments.
     The code now gracefully handles null argv instead of crashing.

  2. TLS General Dynamic support: Implement TLS GD model across all
     architectures (x86, ARM64, ARM, PowerPC64, s390x, RISC-V, LoongArch64,
     MIPS) with a new -tls flag for explicit control. The GD model uses
     __tls_get_addr() for dynamic TLS resolution, enabling dlopen()
     compatibility.

  3. Standards compliance: All changes follow established standards:
     - ELF gABI for DT_INIT_ARRAY behavior
     - ELF TLS specification for General Dynamic model

  Implementation

  The implementation maintains full backward compatibility with existing
  glibc systems while enabling support for other libc implementations. The
  -tls flag allows users to control TLS model selection (auto/LE/IE/GD)
  with sensible defaults that automatically select GD for c-shared/c-archive
  builds on Unix platforms.

  Testing

  - Verified dlopen() works on Alpine Linux (musl) with the fixes
  - Confirmed no regressions on Ubuntu/Debian (glibc) systems
  - All existing tests pass
  - Cross-compilation tested for linux/amd64, linux/arm64, linux/386,
    freebsd/amd64, openbsd/amd64

  Impact

  This enables Go shared libraries to work correctly on Alpine Linux and
  other non-glibc systems, addressing a major compatibility gap that has
  affected container deployments and embedded systems for years.

  Fixes #13492
  Fixes #54805

  Co-authored-by: Alexander Musman <alexander.musman@gmail.com>


jgowdy-godaddy:
@richfelker 👍

teta2k:
Hi @jgowdy, do you mind having another look at this pr? I tried it on both alpine and ubuntu and it didn't work for me (SIGSEGV when loading the test library from test-cshared.sh)

jgowdy-godaddy:
I'm digging into this again, sorry for the delay!

PR #75074, opened 2025-08-18
cmd/link: remove hard dependency on ld.gold on aarch64

### LINKED ISSUE #22040: cmd/link: stop requiring gold on arm64 when GNU ld is fixed
Any, more specifically 1.8, 1.9, master, most probably all versions supporting AArch64/arm64 target

Yes

linux/arm64

### What did you do?

GC on linux/arm64 assumes gold linker availability see `src/cmd/link/internal/ld/lib.go:1182`
When gold is not available gcc fails due to it missing, when it tries to invoke it. All in shared_test.
When gold flag is remove by removing `sys.ARM64`  from codition on `src/cmd/link/internal/ld/lib.go:1173`, build/test fail with 

```
--- FAIL: TestTrivialExecutable (0.29s)
	shared_test.go:66: executing go install -installsuffix=5577006791947779410 -linkshared trivial failed exit status 2:
		# trivial
		/root/go/pkg/tool/linux_arm64/link: running gcc failed: exit status 1
		/usr/bin/ld: /tmp/go-link-151650837/go.o(.data.rel.ro+0x2c0): unresolvable R_AARCH64_ABS64 relocation against symbol `go.link.abihash.libruntime,sync-atomic.so'
		/usr/bin/ld: final link failed: Nonrepresentable section on output
		collect2: error: ld returned 1 exit status
```
And so one for most of the cases shared_test.

All tests pass. GC working with default binutils ld.

### What did you see instead?

GC failing to build some binaries with binutils ld.

solskogen:
Fixes #22040

### DIFF
--- a/src/cmd/link/internal/ld/lib.go
+++ b/src/cmd/link/internal/ld/lib.go
@@ -1675,27 +1675,8 @@ func (ctxt *Link) hostlink() {
 			altLinker = "lld"
 		}
 
-		if ctxt.Arch.InFamily(sys.ARM64) && buildcfg.GOOS == "linux" {
-			// On ARM64, the GNU linker will fail with
-			// -znocopyreloc if it thinks a COPY relocation is
-			// required. Switch to gold.
-			// https://sourceware.org/bugzilla/show_bug.cgi?id=19962
-			// https://go.dev/issue/22040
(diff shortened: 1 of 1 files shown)


solskogen:
duplicate of 391115

PR #75551, opened 2025-09-20
cmd/compile/internal/ssa: remove awkward formatting

willfaught:
Use formatting consistent with rest of the sentence.

### DIFF
--- a/src/cmd/compile/internal/ssa/README.md
+++ b/src/cmd/compile/internal/ssa/README.md
@@ -58,7 +58,7 @@ value of 1. One more example:
 
 	v17 (361) = Store <mem> {int} v16 v14 v8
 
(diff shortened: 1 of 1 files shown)


willfaught:
@Jorropo I don't see a way to reply to you in Gerrit, so I will here:

>Store refers to the SSA Pseudo-Instruction, not the word store.

I'm aware. The "int" in that same sentence, referred to in the description, doesn't use the code syntax, and the "`foo`ed" formatting is awkward in my opinion, hence the change.

willfaught:
@Jorropo Forgot to mention that it also doesn't render well. It looks more like some non-ASCII Unicode chars were accidentally used than something done intentionally:

Screenshot 2025-09-19 at 6 13 00 PM

This is what got me to look at the source in the first place. It just looks wrong.

Issue #75350, opened 2025-09-10
proposal: cmd/compile, cmd/asm: add support for Zicond extension on riscv64


lxq015:
### Background
The RISC-V Zicond extension provides conditional zero instructions (czero.eqz and czero.nez), which are useful for implementing conditional moves and other conditional operations without branching. This can lead to performance improvements by avoiding branch mispredictions.

Currently, the Go compiler does not generate these instructions for riscv64 targets, even when the underlying hardware supports them. This represents a missed optimization opportunity.
### Proposed Solution
Add comprehensive support for the Zicond extension to the Go toolchain for riscv64, including assembler, compiler, and disassembler support. All features are conditionally enabled when buildcfg.GORISCV64 >= 23.
### Implementation Details
1. Assembler Support (cmd/internal/obj/riscv):
- Add ACZEROEQZ and ACZERONEZ to instruction enum in cpu.go
- Add instruction encoding in inst.go:
  - CZEROEQZ: opcode=0x33, funct3=0x5, funct7=0x7
  - CZERONEZ: opcode=0x33, funct3=0x7, funct7=0x7
- Add instruction data in obj.go with rIIIEncoding format
- Add instruction names in anames.go
- Add instruction mapping in obj.go for proper register assignment

2. Compiler Support (cmd/compile):
Based on the RISC-V Zicond specification, implement comprehensive SSA lowering rules:

RISCV64latelower.rules
```
 // Zicond rules support 
(Select <typ.UInt64> cond trueval falseval)  =>  (ADD (CZEROEQZ <typ.UInt64> falseval cond) (CZERONEZ <typ.UInt64> trueval cond))

// Conditional arithmetic operations with Zicond 
(CondSelect <t> (ADD x y) x (SEQZ z)) && buildcfg.GORISCV64 >= 23 => (ADD x (CZERONEZ <t> y z))
(CondSelect <t> (ADD x y) x (SNEZ z)) && buildcfg.GORISCV64 >= 23 => (ADD x (CZEROEQZ <t> y z))

// Conditional subtract operations with Zicond
(CondSelect <t> (SUB x y) x (SEQZ z)) && buildcfg.GORISCV64 >= 23 => (SUB x (CZERONEZ <t> y z))
(CondSelect <t> (SUB x y) x (SNEZ z)) && buildcfg.GORISCV64 >= 23 => (SUB x (CZEROEQZ <t> y z))

// Conditional bitwise-or operations with Zicond
[...]

PR #75577, opened 2025-09-24
cmd/compile/internal/ssa: add codegen for Zicond extension on riscv64

lxq015:
Zicond is a mandatory extension in rva23u64. This patch converts 
certain branches to CondSelect and optimizes them to Zicond 
instructions on RISC-V in appropriate cases, along with additional 
optimization rules.

Zicond can provide performance benefits on unpredictable branches by
avoiding branch misprediction penalties. However, on simple predictable
branches, zicond uses 4 instructions vs 2 for traditional branches,
which can cause performance regressions.

To avoid regressions, we keep CondSelect globally disabled for riscv64 
and only enable it for the ConstantTimeSelect intrinsic,
which has been shown to benefit from zicond:

goos: linux
goarch: riscv64
pkg: crypto/subtle
CPU: SG2044
                      │   nozicond.txt   │     zicond.txt     │
                      │     sec/op       │   sec/op   vs base │
ConstantTimeSelect-44   2.325n ± 4%   1.750n ± 2%  -24.69% (p=0.000 n=10)

Future work can explore enabling zicond for other cases that can benefit
from zicond.

Follow-up to CL 631595
Updates #75350

Co-authored-by: wangpengcheng.pp@bytedance.com
mengzhuo1203@gmail.com


Ahmadmizo:
75577

Issue #69536, opened 2024-09-19
crypto: obtain a FIPS 140-3 validation


FiloSottile:
## Background

FIPS 140 is a set of U.S. Government requirements for cryptographic modules. A number of companies must comply with them, for example as part of a broader FedRAMP compliance posture. (If that's not you, you can ignore this. Run!)

Current solutions for Go program compliance are based on cgo, and replace some of the crypto packages internals with FIPS 140 validated non-memory safe modules. These solutions come with varying levels of support (for example the Go+BoringCrypto solution is not officially supported and its compliance profile is left to the user to assess), introduce memory unsafe code, sometimes delay Go version updates, can have performance issues, affect the developer experience (for example inhibiting cross-compilation), and their compliance profile is debatable. As Go is adopted more and more in regulated settings, this is going to affect Go's adoption and developer experience.

## The Go FIPS module

We plan to pursue a FIPS 140-3 validation for the NIST approved components of the Go standard library. The resulting module will be distributed as part of the standard library under the same license as the rest of the Go project, and will be transparently used by the relevant standard library packages with no API changes (wherever possible).

Users will be able to select the module to use at build time, for example choosing between a certified version, a version in the In Process list, or the latest unvalidated update. Moreover, we'll provide some mechanism for applications to disable the use of non-approved algorithms and modes at runtime.

## Further planning details

The goal is shipping the module as part of Go 1.24, assuming our validation strategy is successful. This is the first time as far as we know that a Go library (or any non-Java memory safe library) is validated.

Unless completely unavoidable, we'll not compromise on security to achieve compliance. For example, we will inject random bytes from the kernel as additional input per SP 800-90Ar1, Section 8.7.2, every time we use the mandatory DRBG, and we'll use a dedicated DRBG for ECDSA to implement a "hedged" nonce generation equivalent to what crypto/ecdsa does now (safer than both NIST options of fully random and deterministic). Also, we'll try to add minimal complexity to regular non-FIPS builds.

NIST approved packages will be prioritized in being moved to the standard library (#65269) to get validated along the rest.

We'll test at least on Linux on amd64 and arm64. Further details will be available later in the process. (If you have specific requirements, please inquire about becoming a sponsor, see below.)

We aim to deprecate and hopefully remove Go+BoringCrypto once the module lands.

After the initial validation, we plan to revalidate at least every year, and every time a CVE affects the module with no standard library-side mitigation.

[...]

PR #75658, opened 2025-09-29
crypto/rand: add syscall number fallbacks in seccomp_linux

dougm:
seccomp_linux.go is only used for Go's own tests (commit 644628536f24), but the package is included when building programs with CGO_ENABLED=1.

When using an older glibc (e.g. 2.17), SYS_getrandom and SYS_seccomp are not defined, causing cgo compilation to fail.

Note that there are already several types and defines copied from linux headers into seccomp_linux.go to support compilation on older systems.

Updates #69536


dougm:
cc @FiloSottile

PR #75754, opened 2025-10-04
errors: add Match and IsAny

maranqz:
Fixes https://github.com/golang/go/issues/70961


jub0bs:
@maranqz Perhaps you should close this PR, at least until (if ever) the proposal you link to gets approved.

PR #76211, opened 2025-11-07
cmd/compile: support Zba extensions in riscv64 compiler

MouseSplinter:
Add compiler support for Zba entensions, which are mandatory in the rva22u64 profile. These can be used to accelerate address computation.


MouseSplinter:
This is a duplicate of https://go-review.googlesource.com/c/go/+/580276/2, @lxq015 and @newborn22 are the co-authors and @4a6f656c is the reviewer. For the reason that Google blocks access from Chinese IPs and my company data security, I want to upstream on github. It seems that I need to process the blocked merging.

lxq015:
I'm facing the same IP blocking issue, so I can only submit via GitHub as well. We've tried various approaches but haven't been able to resolve this problem.

However, the current failure in your submission is because you need to sign the CLA first.

MouseSplinter:
> I'm facing the same IP blocking issue, so I can only submit via GitHub as well. We've tried various approaches but haven't been able to resolve this problem.
> 
> However, the current failure in your submission is because you need to sign the CLA first.

Yes, this because the commit account didn't sign the CLA but upstream account has signed. Thank you.

PR #76658, opened 2025-12-02
net: shuffle resolved IP addresses for load balancing

### LINKED ISSUE #31698: net: DNS load balancing during bursts of traffic
Right now inflight DNS lookups are [grouped together](https://github.com/golang/go/blob/master/src/net/lookup.go#L267). 

Since DNS resolution picks the [first available IP](https://github.com/golang/go/blob/master/src/net/dial.go#L417), this has the side effect of resolving the same IP when there are bursts in traffic. 

This was surprising to me since I expected load to be distributed amongst the resolved hosts, especially during these bursts (assuming equal RFC 6724 weighting).

Load distribution via DNS seems like a common use case/expectation - it would be nice to be able to enable randomized DNS resolution. Being able to disable DNS query grouping or customize the way the group key is generated would also work.

AnonyScorpio:
Resolved addresses are now shuffled before sorting per RFC 6724,
so concurrent callers do not always receive the same order.
This helps distribute network connections across multiple addresses.

The address slice is now always cloned before returning,
allowing the caller to mutate it safely even when results are shared
by concurrent singleflight callers.

Fixes #31698

### DIFF
--- a/src/net/dnsclient.go
+++ b/src/net/dnsclient.go
@@ -27,6 +27,18 @@ func randIntn(n int) int {
 	return randInt() % n
 }
 
+// shuffle the same as rand.Shuffle in package math/rand/v2
+// for reduce dependency usage.
+func shuffle(n int, swap func(i, j int)) {
+	if n < 0 {
+		panic("invalid argument to Shuffle")
+	}
+	for i := n - 1; i > 0; i-- {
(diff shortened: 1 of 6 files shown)


AnonyScorpio:
@ianlancetaylor here

PR #76667, opened 2025-12-03
syscall: make LoadConnectEx not IPv4-specific on Windows

### LINKED ISSUE #29759: httptest on Windows is IPv4-specific

go version go1.11.1 windows/amd64

Yes

windows/amd64

### What did you do?

Run a test leveraging net/http/httptest on a host with IPv6 but without IPv4.

To reproduce:

1. As administrator, run <pre>netsh interface ipv4 uninstall</pre>
1. Reboot
1. Execute an httptest as in https://play.golang.org/p/xlMUEEhgj5s

Test would pass, since the client and server code were both valid, and pass when IPv4 is available.

### What did you see instead?

2019/01/15 17:26:59 Get http://[::1]:49679: dial tcp [::1]:49679: failed to find ConnectEx: An address incompatible with the requested protocol was used.

linsite:
Calls to Socket in LoadConnectEx always fail with the error
"An address incompatible with the requested protocol was used"
when IPv4 is disabled on Windows. We can work around this by
detecting that specific error and retrying with an IPv6 socket.

Fixes #29759

### DIFF
--- a/src/syscall/syscall_windows.go
+++ b/src/syscall/syscall_windows.go
@@ -14,6 +14,7 @@ import (
 	"internal/oserror"
 	"internal/race"
 	"internal/strconv"
+	strings "internal/stringslite"
 	"sync"
 	"unsafe"
 )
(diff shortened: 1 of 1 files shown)


linsite:
@gopherbot  abandon the change in Gerrit.

PR #76863, opened 2025-12-17
runtime: floor cgroup CPU limits in adjustCgroupGOMAXPROCS

marcosdotps:
adjustCgroupGOMAXPROCS derives GOMAXPROCS from the cgroup CPU quota. When the quota is fractional, the current computation rounds up, which can set GOMAXPROCS above the available CPU quota.

Round fractional quotas down instead. Keep the existing minimum of 2 for very small quotas to preserve a baseline level of parallelism.

This reduces contention and CPU throttling in production deployments using fractional CPU quotas.

Proposal: #76835 

Fixes #76835


marcosdotps:
## Evidences

### Before change
before

### After change
after

PR #77247, opened 2026-01-21
cmd/compile: allow generic function values in composite literals

next-n:
Allow generic function values to be used in composite literals.

This fixes a case where assigning a generic function value inside a
composite literal was incorrectly rejected.

This change is mirrored in go/types via go generate go/types.

Fixes [#77245](https://go.dev/issue/77245)


Oguidan:
Hi @next-n please can you give me more details, I am new and will like to know

Issue #78513, opened 2026-04-02
runtime: crash in WASM plugins without open file descriptors due to `poll_oneoff` call


agirbal:
Summary: `runtime: netpoll_wasip1 throws "poll_oneoff failed" in WASM runtimes that return ENOTSUP, crashing plugins without open file descriptors`

### What version of Go are you using?

go1.26.1 (reproducible from go1.24+ where standard Go wasip1 support was introduced)

### What operating system and processor architecture are you using?

`GOOS=wasip1 GOARCH=wasm`

### What did you do?

Built a Go proxy-WASM HTTP filter using `GOOS=wasip1 -buildmode=c-shared` and deployed it in Envoy via the `envoy.filters.http.wasm` filter. The plugin:

- Has no open file descriptors
- Makes no network calls
- Uses no `time.Sleep` or channels
- Does allocate ~200–400 bytes of heap per request (Go struct + JSON marshal)

After some hours of traffic (enough to trigger a GC cycle), Envoy logs:

```
fatal error: poll_oneoff failed

runtime stack:
runtime.throw2({0x...})
    src/runtime/panic.go:1023
runtime.netpoll(0x0)
    src/runtime/netpoll_wasip1.go:216
runtime.startTheWorldWithSema(...)
    src/runtime/proc.go:1753
runtime.gcStart(...)
    src/runtime/mgc.go:...
```

Envoy returns ENOTSUP (errno 52) for `poll_oneoff`. The Go runtime treats any non-EINTR error as fatal.

### What did you expect to see?

The Go runtime should not crash. For WASM plugins with no open FDs, `poll_oneoff` is called only as a scheduler/timer sleep mechanism — not for real I/O. Returning 0 events is the correct and harmless answer.

### What did you see instead?

Fatal crash, WASM VM permanently dead, all subsequent requests on that Envoy worker return 503.

### Root cause

The trigger is `os` package initialisation. Any real-world plugin uses the `proxywasm` SDK, which imports `fmt`, which imports `os`. The `os` package initialises `os.Stdout`/`os.Stderr` as pollable file descriptors, which calls:

```
poll.FD.Init → serverInit.Do(runtime_pollServerInit) → netpollGenericInit()
```

This sets `netpollinited = true` at process startup, before any request is served.

From that point, every GC cycle calls `netpoll(0)` from `startTheWorldWithSema`:

```go
// src/runtime/proc.go
if netpollinited() {
    list, delta := netpoll(0) // non-blocking poll
    ...
}
```

[...]

PR #78522, opened 2026-04-03
runtime: handle ENOTSUP/ENOSYS from poll_oneoff in netpoll_wasip1

ssam18:
Some WASM host runtimes export poll_oneoff but return ENOTSUP or ENOSYS rather than implementing it, because they're event-driven and blocking the WASM thread would deadlock their worker. This caused a fatal crash during GC cycles in any wasip1 plugin that transitively imports fmt or os — which is nearly everything.

The fix treats ENOTSUP (58) and ENOSYS (52) the same way as returning zero events, which is always safe when there are no real FD subscriptions and harmless otherwise since the scheduler will re-poll on the next GC cycle. This mirrors what netpoll_fake.go already does for the js/wasm target.

Fix for the issue #78513

### DIFF
--- a/src/runtime/netpoll_wasip1.go
+++ b/src/runtime/netpoll_wasip1.go
@@ -33,7 +33,11 @@ import "unsafe"
 // Since poll_oneoff is similar to poll(2), the implementation here was derived
 // from netpoll_aix.go.
 
-const _EINTR = 27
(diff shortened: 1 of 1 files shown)


ssam18:
can someone review this PR?

PR #78631, opened 2026-04-10
runtime: fix should not compare uint64 with zero

cuiweixie:


afurm:
The uint64 underflow fix is correct — comparing before subtracting prevents the wraparound to a huge number. The logic is cleaner now too.

One small thing: the original code had `a.gWaiting < 0` as the condition, which was the bug (unsigned subtraction wraps). The new code fixes it, but consider adding a comment explaining *why* the comparison must be done in this order, since to a casual reader `gTotal < gRunning+gRunnable+gNonGo` might look like an arbitrary ordering. A brief comment about the uint64 semantics would make the intent clearer for future maintainers.

Issue #26492, opened 2018-07-19
[proposal] go build: add -static flag


kolyshkin:
This is a proposal to add `-static` flag to `go build`.

Producing a static build with go already requires a non-trivial amount of flags passed to `go build`, which on Linux currently amounts to something like:

`-ldflags '-extldflags "-fno-PIC -static"' -buildmode pie -tags 'osusergo netgo static_build'` [1]

...and this magic string keeps growing.

It would be nice encapsulate this sacred knowledge internally, exposing it via a new `-static` flag for `go build` and friends?

In addition to the above benefit of hiding the complexity, `-static` can also bring us another few nice things, such as:
* automatic `static_build` build tag third-party software can rely upon. Currently it seems like a de-facto standard, used a lot, but it has to be explicitly defined like in [1] above.
* providing `--static` flag to `pkg-config` invocations (those initiated by `// #cgo pkg-config: lib` lines in the source code). It will solve another issue (linking against a proper set of libraries for both static and dynamic link cases) for which a somewhat verbose workaround is currently required (for example, see [ploop_link_static.go](https://github.com/kolyshkin/goploop/blob/master/ploop_link_static.go) and [ploop_link_dynamic.go](https://github.com/kolyshkin/goploop/blob/master/ploop_link_dynamic.go)).

### See also
* https://github.com/golang/go/issues/12058 (original proposal for adding conditional `--static` to `pkg-config`)
* https://github.com/golang/go/issues/24787 (initial version of this very proposal)

rsc:
This seems OK but the tag should just be "static" not "static_build" (it's a build tag already). And we should make os/user and net do the right thing - whatever that is - instead of having to hard-code their tags too. 

Note that the assumption is that -static does not disable cgo; it just makes the cgo uses statically linked. FWIW, I don't know how well statically linked cgo works, but apparently well enough.

It sounds like maybe the os/user and net non-cgo restrictions only apply on Linux (more precisely, GNU/Linux, since this is a glibc restriction).

rsc:
If anyone wants to work on this for Go 1.12, help is appreciated.

amenzhinsky:
I get the idea of combining netgo and usergo tags that looks a bit magical, but what's the reason of using pie builmode since if forces the toolchain to use an external linker (gcc, clang) and many build systems, like alpine docker images, come without one.

kolyshkin:
> This seems OK but the tag should just be "static" not "static_build" (it's a build tag already).

[...]

PR #78659, opened 2026-04-12
cmd/go: add -static flag for simplified static builds

SAY-5:
When `-static` is passed, the build produces a statically linked binary without having to remember `CGO_ENABLED=0` or the `-extldflags` incantation.

If CGO is explicitly enabled (`CGO_ENABLED=1`), it keeps cgo and passes `-extldflags=-static` to the linker. Otherwise it just disables cgo for a pure Go static binary.

Updates #26492


SAY-5:
I have read and agree to the Google Contributor License Agreement at https://cla.developers.google.com/. My GitHub account is SAY-5.

PR #78738, opened 2026-04-14
fmt: clarify %q verb is not valid for integers

Jah-yee:
Good day,

As of Go 1.26, go vet reports an error when %q is used with integers:

This commit adds notes to the documentation to clarify that %q should
not be used with integers.

Thank you for your work on this project. I hope this small fix is helpful. Please let me know if there's anything to adjust.

Warmly, RoomWithOutRoof

### DIFF
--- a/src/fmt/doc.go
+++ b/src/fmt/doc.go
@@ -44,6 +44,8 @@ Integer:
 	%o	base 8
 	%O	base 8 with 0o prefix
 	%q	a single-quoted character literal safely escaped with Go syntax.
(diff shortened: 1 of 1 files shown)


afurm:
The %q note mentions %c as an alternative for the character representation, but %c handles Unicode code points while %q is specifically for character literals with Go syntax escaping. For negative integers or large unsigned integers, %c would produce unexpected output. Is it worth also mentioning that %d or %v is the safest general-purpose choice for integers?

PR #78761, opened 2026-04-15
reflect: use MapMaxElemBytes instead of MapMaxKeyBytes

cuiweixie:


src/reflect/map.go
```diff
@@ -97,7 +97,7 @@ func MapOf(key, elem Type) Type {
 	if ktyp.Size_ > abi.MapMaxKeyBytes {
 		mt.Flags |= abi.MapIndirectKey
 	}
-	if etyp.Size_ > abi.MapMaxKeyBytes {
+	if etyp.Size_ > abi.MapMaxElemBytes {
```
afurm, line 100: The `MapIndirectElem` flag controls whether element pointers are stored indirectly (for large elements). Using `MapMaxElemBytes` instead of `MapMaxKeyBytes` here is correct — the size threshold for indirection should match what the runtime actually uses when allocating map buckets. The symmetry with the key-size check two lines above makes the intent clear.

PR #78797, opened 2026-04-16
crypto/tls: remove deadcode

cuiweixie:


cpu:
Addressed in https://go-review.googlesource.com/c/go/+/771461

PR #78928, opened 2026-04-24
reflect: fix nil array pointer caused panic

cuiweixie:


cuiweixie:
instead of panic, it should be consistent with below code:  
```
var p *[3]int
for i := range p {
    fmt.Println(i)
}
```
output:  
```
0
1
2
```

PR #78949, opened 2026-04-24
net/http: fix ServeMux routing of [hostname]:port

### LINKED ISSUE #78944: net/http: ServeMux routes `[hostname]:port` to `hostname`
`ServeMux` will route a request with a host of `[hostname]:port` to the host `hostname`:

```go
// register handler for example.com
var mux http.ServeMux
mux.Handle("GET example.com/", handler)

// request for [example.com]:80 returns the above handler
req, _ := http.NewRequest("GET", "http://[example.com]:80/", nil)
h, pattern := mux.Handler(req)
```

This is arguably wrong, since "[example.com]:80" is not a valid host; the bracketed form should only be used for IPv6 addresses.

Whether this is the fault of ServeMux for misinterpreting the invalid host or the HTTP server for accepting it is debatable.

This was reported to us as a potential vulnerability, on the grounds that an intermediary proxy could (for example) reject requests to "example.com" but permit "[example.com]:80" through. Our consensus is that the threat model under which this causes problems is dubious, and that we should not classify this as a vulnerability. Perhaps we should change the behavior of the server and/or ServeMux, but this will be hardening and/or a bugfix. As a practical consideration, changing this behavior runs a high risk of breaking existing users, and is not something we want to backport to earlier releases.

Jah-yee:
## Summary

Fixes golang/go#78944.

## Problem

ServeMux incorrectly routes requests with a Host of `[hostname]:port` to the handler registered for `hostname`. The bracketed form is not valid for hostnames (only for IPv6); net.SplitHostPort returns an error, and the previous code returned the input unchanged (preserving the brackets).

## Solution

In `stripHostPort`, when SplitHostPort fails and the input starts with `[`, strip the brackets and recursively call `stripHostPort` to also strip the port. This correctly treats `[example.com]:80` as `example.com`.

## Changes

- `src/net/http/server.go`: 7 lines added to `stripHostPort` error path

### DIFF
--- a/src/net/http/server.go
+++ b/src/net/http/server.go
@@ -2675,6 +2675,13 @@ func stripHostPort(h string) string {
 	}
 	host, _, err := net.SplitHostPort(h)
 	if err != nil {
+		// "[hostname]:port" is not a valid host (IPv6 uses [::1]:port).
+		// Strip brackets and recurse to strip the port.
+		if strings.HasPrefix(h, "[") {
+			if j := strings.IndexByte(h, ']'); j >= 0 {
+				return stripHostPort(h[1:j])
+			}
+		}
 		return h // on error, return unchanged
 	}
 	return host


Jah-yee:
Closing this PR. The routing fix has been superseded by more recent submissions on the same topic. Refer to #78955 or #78957 for the latest work on the net package.

Jah-yee:
Superseded by #78955 and #78957

PR #78955, opened 2026-04-25
cmd/objdump: use section end for last symbol size in macho files

### LINKED ISSUE #78811: cmd/objdump: doesn't disassemble the last function in a macho file
On darwin, do:

main.c:
```
void f();
void g();

int main() {
  f();
  g();
  return 0;
}
```
main.s:
```
	.globl _f
_f:
	add	x0, x0, x0
	add	x1, x1, x1
	ret

	.globl _g
_g:
	add	x2, x2, x2
	add	x3, x3, x3
	ret
```
```
gcc -O3 main.c main.s
go tool objdump a.out
```
This will print the body of `main` and `f`, but does not print the body of `g`. We get only the `TEXT` line:
```
TEXT _main(SB) 
  :0			0x100000328		a9bf7bfd		STP.W (R29, R30), -16(RSP)	
  :0			0x10000032c		910003fd		MOVD RSP, R29			
  :0			0x100000330		94000005		CALL _f(SB)			
  :0			0x100000334		94000007		CALL 7(PC)			
  :0			0x100000338		52800000		MOVW $0, R0			
  :0			0x10000033c		a8c17bfd		LDP.P 16(RSP), (R29, R30)	
  :0			0x100000340		d65f03c0		RET				

TEXT _f(SB) 
  :0			0x100000344		8b000000		ADD R0, R0, R0		
  :0			0x100000348		8b010021		ADD R1, R1, R1		
  :0			0x10000034c		d65f03c0		RET			

TEXT _g(SB) 
```
Looks like the size of the `g` symbol is set to 0. The code that does this, in cmd/internal/objfile/macho.go:symbols(), sets the size of a symbol by looking at the address of the next higher symbol and subtracting. For the last symbol in the file, this doesn't work. We should probably use the end of the containing section or something.

The "CALL 7(PC)" instead of "CALL _g(SB)" might be related.

Jah-yee:
In Mach-O files, the last symbol in a section gets size 0
because the code computes size as next_symbol_addr - this_symbol_addr,
and there is no next symbol for the last one.

Use the section end address (section.Addr + section.Size) as the
upper bound when computing the size of the last symbol in a section.

Fixes #78811

### DIFF
--- a/src/cmd/internal/objfile/macho.go
+++ b/src/cmd/internal/objfile/macho.go
@@ -55,6 +55,10 @@ func (f *machoFile) symbols() ([]Sym, error) {
 		i := sort.Search(len(addrs), func(x int) bool { return addrs[x] > s.Value })
 		if i < len(addrs) {
 			sym.Size = int64(addrs[i] - s.Value)
+		} else if s.Sect != 0 && int(s.Sect) <= len(f.macho.Sections) {
+			// Last symbol in file: use end of section.
+			sect := f.macho.Sections[s.Sect-1]
+			sym.Size = int64(sect.Addr + sect.Size - s.Value)
 		}
 		if s.Sect == 0 {
 			sym.Code = 'U'


Jah-yee:
I've updated the commit message per Ian's feedback — removed the Markdown sections, kept it to a short problem description with the 'Fixes' line at the end.

Regarding Keith's question about [CL 768580](https://go.dev/cl/768580) — that appears to be an independent fix for the same issue (#78811) with the same code change. Happy to defer to whichever CL you prefer, or I can close this one if that one goes in first.

randall77:
> Regarding Keith's question about [CL 768580](https://go.dev/cl/768580) — that appears to be an independent fix for the same issue (https://github.com/golang/go/issues/78811) with the same code change. Happy to defer to whichever CL you prefer, or I can close this one if that one goes in first.

They are both your CLs, are they not?

Jah-yee:
[CL 768580](https://go.dev/cl/768580) is not mine — I'm only the author of [CL 770644](https://go.dev/cl/770644) (this PR). It looks like two independent fixes for the same issue (#78811). Happy to close this one if you prefer the other approach, or keep this if you'd rather merge it. Let me know how you'd like to proceed.

randall77:
You are also RoomWithOutRoof, are you not? That's what your github name is.

Jah-yee:
No, I'm not RoomWithOutRoof — that's a different contributor. My GitHub account is Jah-yee (https://github.com/Jah-yee). These are two independent fixes for the same issue (#78811). I'm happy to close this PR if you prefer the other approach.

randall77:
This is your github page:
Screenshot 2026-05-02 at 10 18 18 AM

Jah-yee:
My apologies — you're right. My GitHub display name is indeed RoomWithOutRoof. [CL 768580](https://go.dev/cl/768580) was imported from one of my earlier PRs for the same issue (#78811).

Since [CL 768580](https://go.dev/cl/768580) already has Code-Review+2 and passed TryBots, it should be the one to go with. I'll close this PR (#78955) and its corresponding [CL 770644](https://go.dev/cl/770644) to avoid duplication. Sorry for the confusion.

PR #79033, opened 2026-04-29
net/http: combine two strconv.FormatInt call to one

cuiweixie:


far24309765321-rgb:
1. 

- > **_[](url)_**

1.  ### 

- > [](url)

> _`[](url)`_

PR #79061, opened 2026-04-30
uuid: document unchecked error in NewV4 and NewV7

mrbishalbaniya:
Found and documented a bug where NewV4() and NewV7() ignore errors returned by crypto/rand.Read(). This can lead to UUIDs with insufficient randomness in rare failure scenarios.

The bug report includes:
- Detailed problem description and impact analysis
- Affected code locations with line numbers
- Three proposed fix options with pros/cons
- Comparison with other language implementations
- Testing recommendations

While crypto/rand.Read() failures are rare, ignoring errors violates Go's error handling principles and could have security implications for applications relying on UUID unpredictability.

Recommended fix: Add error checking with panic, consistent with the existing MustParse() pattern in the uuid package.

the Gerrit change.

  + No trailing period
+ No Markdown

### DIFF
--- a/BUG_REPORT.md
+++ b/BUG_REPORT.md
@@ -1,0 +1,259 @@
+# Bug Report: Unchecked Error in UUID Generation Functions
+
+## Summary
+
+The `uuid` package's `NewV4()` and `NewV7()` functions ignore errors returned by `crypto/rand.Read()`, which can lead to UUIDs being generated with insufficient randomness or predictable values in rare failure scenarios.
+
+## Severity
+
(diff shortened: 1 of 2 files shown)


Jorropo:
This is useless AI garbage,
it wants to add error handling to a function that never errors:
> Read fills b with cryptographically secure random bytes. It never returns an error, and always fills b entirely. 

It is a shame that no one checked `git blame` or the `crypto/rand.Rand` docs before sending an unusable patch made of `.md` files.

PR #79130, opened 2026-05-01
archive/tar: validate Linkname for insecure paths in Reader.Next

metsw24-max:
The archive/tar Reader validates Header.Name for insecure paths when
GODEBUG=tarinsecurepath=0, but does not apply the same validation to
Header.Linkname.

This means an entry can pass the Header.Name check while carrying a
crafted Linkname that bypasses the intended path safety checks. Since
Header.Linkname is used when creating symlinks and hard links, it can
influence filesystem resolution during extraction in common usage
patterns, so it is equally security-sensitive.

Validate Header.Linkname with filepath.IsLocal in Reader.Next, matching
the existing Header.Name check, and return ErrInsecurePath when it is
non-local. Empty link names are skipped. Also add a regression test to
ensure an insecure Linkname triggers ErrInsecurePath without affecting
subsequent entries.

Updates #55356

### DIFF
--- a/src/archive/tar/reader.go
+++ b/src/archive/tar/reader.go
@@ -52,20 +52,28 @@ func NewReader(r io.Reader) *Reader {
 // Programs that want to accept non-local names can ignore
 // the [ErrInsecurePath] error and use the returned header.
 func (tr *Reader) Next() (*Header, error) {
-	if tr.err != nil {
-		return nil, tr.err
-	}
-	hdr, err := tr.next()
-	tr.err = err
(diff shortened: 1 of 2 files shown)


metsw24-max:
Updated the PR title and description per the robot's findings: added the archive/tar package prefix, wrapped the body at ~76 columns and referenced #55356. Also pushed a follow-up commit that reuses the existing filepath.IsLocal check for Linkname (the separate Clean step was redundant) and fixes formatting, so the change is now a fair bit smaller.

PR #79314, opened 2026-05-10
cmd/compile: add size cache to StdSizes to avoid exponential compile time

hesam-oxe:
This change adds a size cache to StdSizes to memoize Sizeof results for
deeply nested generic types, fixing an exponential compile-time blowup.

Problem:

When compiling deeply nested generic instantiations like

	type T[P any] struct { _,_,_,_,_,_,_,_ P }
	type U T[T[T[T[T[T[T[T[T[T[int]]]]]]]]]]

StdSizes.Sizeof recursed without caching, so the same underlying type was
recomputed millions of times, giving exponential time (8^N, where N is
the nesting depth).

Root cause:

Sizeof for *Struct and *Array types recursively computes the sizes of
their elements/fields, and nothing memoized those results across the
shared nested type graph.

Fix:

- Add sizeCache map[Type]int64 guarded by sync.RWMutex to StdSizes.
- Sizeof checks the cache before computing and stores results after.
- The cache map is initialized lazily.

Impact:

- Compile time for deeply nested generics drops from exponential to
  linear in the number of distinct types.
- No API change; the cache is transparent to callers.
- Thread-safe via RWMutex.

Fixes #78342

### DIFF
--- a/src/cmd/compile/internal/types2/sizes.go
+++ b/src/cmd/compile/internal/types2/sizes.go
@@ -6,6 +6,8 @@
 
 package types2
 
+import "sync"
+
 // Sizes defines the sizing functions for package unsafe.
 type Sizes interface {
 	// Alignof returns the alignment of a variable of type T.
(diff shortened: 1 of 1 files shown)


hesam-oxe:
@google-cla I've signed the Individual CLA with email 
chngyzkhanwhsht@gmail.com. Can you please re-check?

PR #79434, opened 2026-05-16
cmd/go/internal/doc: reject meta packages in go doc

### LINKED ISSUE #53446: Unexpected output with go doc std
go version go1.19beta1 linux/amd64

<pre>
$ go version
go version go1.19beta1 linux/amd64
</pre>

Yes

### What did you do?

Originally, I was attempting list all standard packages via `go list std`. I tried to find documentation `go help list` that confirms the use of "std" as some kind of proxy for listing the standard library. That led me to `go doc std`.

I expected to see nothing (as its not a package) , or some kind of magic placeholder, or some kind of invalid package error.

### What did you see instead?

What I think I see is the first standard library package in alpha order, `archive/tar`.

```
~$ go doc std
package tar // import "archive/tar"

Package tar implements access to tar archives.

Tape archives (tar) are a file format for storing a sequence of files that
can be read and written in a streaming manner. This package aims to cover most
variations of the format, including those produced by GNU and BSD tar tools.

const TypeReg = '0' ...
var ErrHeader = errors.New("archive/tar: invalid tar header") ...
type Format int
    const FormatUnknown Format ...
type Header struct{ ... }
    func FileInfoHeader(fi fs.FileInfo, link string) (*Header, error)
type Reader struct{ ... }
    func NewReader(r io.Reader) *Reader
type Writer struct{ ... }
    func NewWriter(w io.Writer) *Writer
```

Maybe `std` is meant to be magic, but if its not, it seems that it would documented in `go help list` and clearer usage with `go doc std`. I could also just be wrong on its usage (found via some random posts).

raffr85:
Meta-package names such as std, cmd, tool, work, and all are package
patterns that expand to multiple packages, not single documentable
packages. Today go doc std can fall through package resolution and print
documentation for an unrelated first match such as archive/tar.

Reject these names when they are the first argument to go doc, including
the two-argument form such as go doc std Foo. Keep go doc -http std
routing to the standard library page.

Fixes #53446

### DIFF
--- a/src/cmd/go/internal/doc/doc.go
+++ b/src/cmd/go/internal/doc/doc.go
@@ -236,11 +236,17 @@ func do(ctx context.Context, writer io.Writer, flagSet *flag.FlagSet, args []str
 			// Outside a module or workspace, go to the documentation for the standard library.
 			return doPkgsite(ctx, "std", "")
 		}
+		if args := flagSet.Args(); len(args) == 1 && args[0] == "std" {
+			return doPkgsite(ctx, "std", "")
+		}
 
 		// If args are provided, we need to figure out which page to open on the pkgsite
 		// instance. Run the logic below to determine a match for a symbol, method,
 		// or field, but don't actually print the documentation to the output.
 		writer = io.Discard
+	}
+	if a := flagSet.Args(); len(a) > 0 && search.IsMetaPackage(a[0]) {
+		return fmt.Errorf("go doc does not support meta-package argument %q", a[0])
 	}
 	var paths []string
 	var symbol, method string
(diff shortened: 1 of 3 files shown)


raffr85:
@googlebot rescan

raffr85:
Tests:
env -u GOROOT -u GOTOOLDIR -u GOFLAGS -u GOENV GOTOOLCHAIN=local GODEBUG=goindex=0 ./bin/go test cmd/go/internal/doc -count=1
env -u GOROOT -u GOTOOLDIR -u GOFLAGS -u GOENV GOTOOLCHAIN=local GODEBUG=goindex=0 ./bin/go test cmd/go -run '^TestScript/doc$' -count=1
Tests:
env -u GOROOT -u GOTOOLDIR -u GOFLAGS -u GOENV GOTOOLCHAIN=local GODEBUG=goindex=0 ./bin/go test cmd/go/internal/doc -count=1
env -u GOROOT -u GOTOOLDIR -u GOFLAGS -u GOENV GOTOOLCHAIN=local GODEBUG=goindex=0 ./bin/go test cmd/go -run '^TestScript/doc$' -count=1

PR #79440, opened 2026-05-17
plan9: restore the check for close() return value in fork/exec

pixelherodev:
With miller's fix applied [1], the race condition that yielded errors
from close() is gone. I've tested backporting that commit to Go 1.4
[2], and the build becomes 100% reliable. As such, ignoring an error
result that has already proven useful seems silly.

Now that the race is gone, restore the code that prints an error on
race conditions; we don't want to silently close the wrong fds.

[1] 639a20da908c2646de46b93ca2933651363ec22a
[2] http://okturing.com/src/19622/body

### DIFF
--- a/src/syscall/exec_plan9.go
+++ b/src/syscall/exec_plan9.go
@@ -197,7 +197,9 @@ dirloop:
 				// control file for descriptor <N> is named <N>ctl
 				continue
 			}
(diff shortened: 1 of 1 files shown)


pixelherodev:
...ah. Unfortunately, I'm going to have to close this, as the CLA requires a Google account, and I do not have one and am not willing to change this.

PR #79627, opened 2026-05-23
cmd/go: RFC 8989 says that file-URL host comparison is case-insensitive

cuiweixie:


src/cmd/go/internal/web/url_other.go
```diff
@@ -9,12 +9,11 @@ package web
...
 	"errors"
 	"path/filepath"
+	"strings"
 )
 
 func convertFileURLPath(host, path string) (string, error) {
-	switch host {
-	case "", "localhost":
-	default:
+	if host != "" && strings.EqualFold(host, "localhost") {
```
1911860538, line 16: should be:
   host != "" && !strings.EqualFold(host, "localhost")

Issue #18502, opened 2017-01-03
x/net/http2: reduce Framer ReadFrame allocations


apolcyn:
Please answer these questions before submitting your issue. Thanks!

This is a performance-related issue/proposal for the `net/http2` library in https://github.com/golang/net/tree/master/http2.

### What version of Go are you using (`go version`)?
1.8beta2

### What operating system and processor architecture are you using (`go env`)?
amd64

I'm running "grpc-go" (https://github.com/grpc/grpc-go) micro-benchmarks (grpc-go uses only the "framer" from the `net/http2` library). Specifically I'm looking at a benchmarks that tests "grpc streaming" throughput, with a couple of clients against one 32 core server.

Details on this benchmark setup: the server has a total of 64 tcp/http2 connections, with 100 long-lived http2 streams over each http2 connection. A streaming "round-trip" is a grpc request-response of about 10 bytes, each fitting into one data frame. The server is ran with a 5 second warmup and a 30 second benchmark, during which it does somewhere around 900K round trips per second.

After multiple changes that reduce memory allocations elsewhere, the memory "alloc_space" profile of the server after running this benchmark looks like:

```
2139.58MB of 2178.13MB total (98.23%)
Dropped 69 nodes (cum <= 10.89MB)
      flat  flat%   sum%        cum   cum%
 1436.57MB 65.95% 65.95%  1436.57MB 65.95%  golang.org/x/net/http2.parseDataFrame
  473.51MB 21.74% 87.69%   473.51MB 21.74%  google.golang.org/grpc.protoCodec.Marshal
  225.50MB 10.35% 98.05%   225.50MB 10.35%  google.golang.org/grpc/transport.(*http2Server).handleData
    2.50MB  0.11% 98.16%   494.02MB 22.68%  google.golang.org/grpc/benchmark.(*testServer).StreamingCall
       1MB 0.046% 98.21%   495.02MB 22.73%  google.golang.org/grpc.(*Server).processStreamingRPC
    0.50MB 0.023% 98.23%   477.01MB 21.90%  google.golang.org/grpc.(*serverStream).SendMsg
         0     0% 98.23%  1442.07MB 66.21%  golang.org/x/net/http2.(*Framer).ReadFrame
         0     0% 98.23%  1678.60MB 77.07%  google.golang.org/grpc.(*Server).handleRawConn
         0     0% 98.23%   495.02MB 22.73%  google.golang.org/grpc.(*Server).handleStream
[...]

PR #79634, opened 2026-05-23
net/http/internal/http2: enable SetReuseFrames in server and Transport

ChrisLundquist:
This stack extends Framer.SetReuseFrames to cover *WindowUpdateFrame,
*HeadersFrame, and *MetaHeadersFrame (in addition to the existing
*DataFrame), reuses the MetaHeadersFrame Fields backing array across
parses, then calls SetReuseFrames once on the per-connection Framer
in serverConn and Transport.newClientConn. Result: -32% geomean
allocations per op on the package's read-path benchmarks (about
-49% on downloads, -20% on header-bound requests), with no
measurable change on write-path benchmarks and no API change.

A new GODEBUG=http2reuseframes=0 setting is registered alongside
http2client / http2server / http2debug and reverts to the pre-CL
allocate-per-frame behavior.

Background

Framer.SetReuseFrames has shipped as an opt-in Framer method since
golang/net CL 34812 (commit bb807669, 2017-02-27, fixing #18502). It
was added for gRPC-Go, whose Framer profiling showed parseDataFrame
accounting for ~66% of allocations (1.4 GB of 2.2 GB) on a 64-conn /
100-stream / 900K RT/sec benchmark. gRPC-Go calls SetReuseFrames
directly on the Framer it owns.

No in-tree caller in the standard library has ever opted in, so the
cache has been effectively dead code from the stdlib's perspective
for nine years. This CL turns it on.

Prior attempts and reviewer concerns

This is the third attempt to enable frame reuse on the stdlib's
HTTP/2 server/Transport:

  - Issue #68352 (still open, 2024-07-09): a request to improve
    frame decoding performance. neild rejected a user-facing config
    option ("There's no need for a user-configurable option here.
    If there are performance improvements that can be made to frame
    decoding in the HTTP/2 implementation, we can just make them.")
    but left the door open for internal enablement.

  - golang/net CL 597455 (PR golang/net#216, 2024-07-10): tried to
    enable on both server and Transport unconditionally. neild's
    blocking review: "this is not correct. The server read goroutine
    (serverConn.readFrames) reads frames and delivers them to the
    main server goroutine for processing," and "Changing the HTTP/2
    server and/or transport to use the frame cache is likely to
    require some significant refactoring." Stale; never landed.

  - golang/net CL 676235 (server-only, 2025-05-25): neild Hold+1 on
    2025-06-24: "this is not a safe change for the server.
    SetReuseFrames causes each frame to be invalidated by the next
    call to ReadFrame, but the server processes frames concurrently
    with ReadFrame calls." Abandoned 2025-10-24.

The blocker on both prior CLs was correctness, specifically retention
of frame data past the next ReadFrame. This CL addresses that with a
static retention audit, an ownership guard that turns retention bugs
into deterministic panics, race-detector tests that exercise the
gated handoff, an end-to-end stress test under -race, and a GODEBUG
escape hatch in case the analysis missed something.

Stack layout

Nine commits, each independently reviewable:

  1. reuse *WindowUpdateFrame across ReadFrame calls
  2. reuse *HeadersFrame across ReadFrame calls
  3. reuse *MetaHeadersFrame across ReadFrame calls, with an
     ownership guard on its accessor methods
  4. reuse the MetaHeadersFrame Fields slice across parses
  5. add microbenchmarks for the SetReuseFrames path
  6. add race-detector tests (positive control, and an adversarial
     negative control that runs in a child process)
  7. enable SetReuseFrames in server and Transport, register
     http2reuseframes GODEBUG, add end-to-end stress and GODEBUG
     wiring tests
  8. register the meta-frame hpack emit callback once per Framer,
     which takes a meta parse with reuse on to zero allocations
  9. build the server request header map with one []string, as the
     Transport already does for responses (independent of frame
     reuse; can be split into its own CL)

Commits 1-3 mirror the existing *DataFrame cache pattern from CL
34812: wholesale "*p = T{...}" reset at the top of each parse
function, slot in frameCache, nil-safe get<Type>Frame getter. No
behavior change without SetReuseFrames; the caches are dormant unless
opted in.

Commit 3 also gives MetaHeadersFrame the equivalent of DataFrame's
checkValid protection: an owned bit set when readMetaFrame returns
the frame without an error and revoked by the next ReadFrame (the
frame becomes the Framer's lastFrame and overrides invalidate); a
frame returned together with an error is never owned and only its
Header is meaningful. PseudoValue,
RegularFields, PseudoFields, and rfc9218Priority panic on a stale
frame, so a retention bug fails deterministically instead of
silently reading another stream's headers.

Commit 4 is where most of the HEADERS-path win comes from: the
Fields backing array stays in the cached frame and is re-filled by
the next meta parse. Two safeguards: the parsed fields are released
as soon as the contract permits, in invalidate at the next ReadFrame
call (which now runs before the Framer blocks on the next frame
header), so an idle connection does not pin its last request's
header values, which can hold secrets (Cookie, Authorization, HPACK
Sensitive fields); and the array is dropped after a block of more
than 64 fields, so a single large HEADERS frame cannot permanently
inflate per-connection memory. The cap is checked against the
block's length rather than the array's capacity: append rounds
capacity up to size classes (35, then 76 entries for this element
size), so a capacity check would never retain an array for a block
above 35 fields.

Safety analysis

Static retention audit across every code path that consumes a
reusable *Frame, plus a separate concurrency audit. Per frame type:

  *DataFrame
    f.Data() is copied into dataBuffer.Write
    (databuffer.go:126 copy(chunk[b.w:], p)) before readMore. The
    pool-allocated chunk has independent backing.

  *WindowUpdateFrame
    Struct has only scalar fields. Safe by construction.

  *HeadersFrame
    Both server and Transport set Framer.ReadMetaHeaders during init,
    so a bare *HeadersFrame never reaches consumer code.
    readMetaFrame clears headerFragBuf=nil and invalidates the
    embedded *HeadersFrame before returning.

  *MetaHeadersFrame
    The Fields backing array is reused across parses, so consumers
    must not retain the slice -- and cannot do so unnoticed: the
    accessor methods panic once the next ReadFrame has been called,
    and the adversarial race test verifies -race observes
    retained-slice reads. (Direct reads of the Fields slice are not
    guarded; the two in-tree sites that read it directly, the
    server's newWriterAndRequest and the Transport's 1xx size
    accounting, do so synchronously before returning.)
    hpack.decodeString returns string(u.b) or
    buf.String(), both of which copy, so Name/Value strings are
    independently allocated and alias neither readBuf nor the Fields
    array. Server's processHeaders / newWriterAndRequest and
    Transport's handleResponse / processTrailers iterate Fields
    synchronously and copy into fresh http.Header maps before
    returning. No code stores *MetaHeadersFrame past dispatch.

Concurrency:

  Server: serverConn.readFrames sends each parsed frame on
    readFrameCh then blocks on gate. The serve goroutine receives,
    calls processFrameFromReader synchronously, then res.readMore(),
    which releases the gate. The single readMore call site fires
    only after the synchronous dispatch returns, so the cache is
    never overwritten while consumer code still references it.

  Transport: clientConnReadLoop.run reads frames in a tight loop
    with no per-frame gate, so each process* function must fully
    consume aliased data before returning. Verified: every byte that
    escapes the loop is either copied (DataFrame -> bufPipe) or
    composed of immutable, independently-allocated Go strings
    (Header maps populated from hpack-decoded strings).

  Spawned goroutines: runHandler (server) only sees the fresh
    *ServerRequest whose Header/Trailer/URL are independently
    allocated. writeFrameAsync, body-close, Shutdown, ping writers
    don't touch read-side cached frames.

Two retention sites intentionally remain outside the reuse contract
and are documented in place rather than silently exempt:
processPing queues writePingAck{f}, which the write scheduler may
serve after the next ReadFrame, and processGoAway stores f in
cc.goAway, whose scalar fields are read at read-loop teardown. Both
are safe solely because PingFrame and GoAwayFrame are not part of
frameCache; the comments require copying the needed fields before
either type is ever added to the cache.

Benchmarks

End-to-end, -count=3 -benchmem, darwin/arm64 (Apple silicon),
reuse on vs GODEBUG=http2reuseframes=0 on this tree (the GODEBUG-off
path is the pre-CL allocation behavior); medians.

Allocations per op (the main win):

    Benchmark                       off      on       delta
    ClientRequestHeaders/0              45       34   -24.4%
    ClientRequestHeaders/10             58       46   -20.7%
    ClientRequestHeaders/100           232      226    -2.6%
    ClientRequestHeaders/1000        3,040    3,032    -0.3%
    ClientResponseHeaders/0             45       34   -24.4%
    ClientResponseHeaders/10           110       97   -11.8%
    ClientResponseHeaders/100          570      562    -1.4%
    ClientResponseHeaders/1000       6,851    6,843    -0.1%
    DownloadFrameSize/16k           357.1k   197.2k   -44.8%
    DownloadFrameSize/64k            97.1k    49.9k   -48.6%
    DownloadFrameSize/128k           49.5k    24.6k   -50.2%
    DownloadFrameSize/256k           24.5k    12.3k   -49.7%
    DownloadFrameSize/512k           12.3k    6,123   -50.1%
    ClientGzip                         168       58   -65.5%
    geomean (allocs/op)                            -31.8%

Write-side benchmarks are unchanged; those paths don't touch
ReadFrame. Header-bound benchmarks at 0/10 headers move by
-10% to -3% in wall time; download throughput is flat within
socket noise.

Per-cache attribution from the microbenchmarks added in commit 5
(darwin/arm64, this tree; ReadMetaFrameFields/N is a block of N
fields, and above the 64-field cap the Reused arm rebuilds the
array every parse by design):

    Cache                   Default ns / B / allocs   Reused ns / B / allocs
    ParseDataFrame              25 /    48 /   1         18 /     0 /   0
    ParseWindowUpdateFrame      23 /    16 /   1         18 /     0 /   0
    ParseHeadersFrame           26 /    48 /   1         20 /     0 /   0
    ReadMetaFrame              192 /   384 /   5        123 /     0 /   0
    ReadMetaFrameFields/40    6142 /  6757 /  81       5513 /   864 /  72
    ReadMetaFrameFields/64    9678 /  7334 / 129       9057 /  1441 / 120
    ReadMetaFrameFields/100  14748 / 14348 / 202      14730 / 14252 / 200

The four allocations that used to remain in ReadMetaFrame/Reused
(the hpack emit closure and its boxed captures) are gone as of
commit 8, which registers the callback once per Framer. Commit 9 is
independent of frame reuse: it removes the server's per-key []string
allocation when building the request Header map, which is most of
the delta at 100+ request headers above.

Test plan

frame_reuse_race_test.go (package http2, synthetic Framer-pair
tests, each run in a "meta" variant where HEADERS surface as
*MetaHeadersFrame and a "raw" variant where HEADERS/CONTINUATION
surface individually and HeaderBlockFragment aliases the read
buffer):

  - TestFrameReuseRaceCorrect: positive control. Models
    serverConn.readFrames with a reader goroutine, a channel, a
    per-frame gate, and a consumer that fully consumes payload
    before signalling done. 800/1000 frames across DATA,
    WINDOW_UPDATE, HEADERS, HEADERS+CONTINUATION. Must PASS under
    -race.

  - TestFrameReuseRaceAdversarial: negative control. Same plumbing,
    but deliberately leaks slices into a sidecar goroutine across
    the next ReadFrame. Under -race it re-executes the test binary
    to run the leaking code in a child process (with
    GORACE=halt_on_error=1) and asserts that the child reports a
    data race, so the control runs on every -race run; without
    -race it is skipped. The detector fires for every retained
    case: Data, HeaderBlockFragment, and the Fields slice (which
    races against the next ReadFrame's clear of the backing array).

frame_reuse_e2e_test.go (package http2_test, real server +
Transport, lands with the enablement commit):

  - TestFrameReuseEndToEndStress: 8 workers x 25 multiplexed
    requests through the real httptest-based HTTP/2 server and
    net/http.Transport, both running with SetReuseFrames enabled.
    Handler and client touch every byte of every header name/value
    and trailer in both directions, drain bodies. Regression test
    against future refactors of processHeaders / handleResponse /
    processTrailers / processData that fail to copy aliased data.

Per-type unit tests land with each cache commit, sharing two generic
helpers (testFrameReuse, testFrameDistinctWithoutReuse):
pointer-stability and no-alloc tests (the latter skipped in -short
mode and under -race, like net/http's alloc tests), poison-overwrite
tests for stale-field leaks, distinct-pointer tests for the reuse-off
contract, accessor-panic tests for all four guarded accessors and
for a frame returned together with an error, and Fields tests for
backing-array reuse across block sizes up to the cap, the
tail-stays-zero invariant behind the clear, release at the next
ReadFrame call, and the retention cap. The enablement commit adds
TestSetReuseFramesFromGODEBUG (the helper and its non-default
metric) and TestReuseFramesGODEBUGWiring (server and Transport both
honor the setting).

Validation runs (darwin/arm64, this revision):

    Run                                              Result
    net/http/internal/http2 (full, -short, -race)    PASS
    same with GODEBUG=http2reuseframes=0
        (full, -race)                                PASS
    every commit in the stack: gofmt, vet, -short,
        and the reuse tests under -race              PASS
    net/http (full, -short -race, GODEBUG=0 -short)  PASS
    net/http/httputil, httptest, httpcommon          PASS
    internal/godebugs, runtime/metrics,
        go/build TestDependencies                    PASS
    TestFrameReuseRaceAdversarial (meta+raw) -race   PASS (the child
        process reports the race in every retained case)
    TestFrameReuseEndToEndStress -race               PASS

GODEBUG escape hatch

Registered alongside the existing http2 GODEBUGs in
src/internal/godebugs/table.go, with no version-dependent default:

    Value                              Behavior
    unset or http2reuseframes=1        reuse on (default)
    GODEBUG=http2reuseframes=0         reuse off (pre-CL behavior)
                                       increments
                                       /godebug/non-default-behavior/
                                       http2reuseframes:events

Documented in doc/godebug.md under the Go 1.28 section. The setting
has no version-dependent default: frame reuse changes no observable
behavior, so gating it on the go.mod version would only withhold the
allocation savings from existing programs. The check-and-count logic
lives in a single Framer helper (setReuseFramesFromGODEBUG) called by
both the server and the Transport, and TestReuseFramesGODEBUGWiring
checks both call sites.

The intent is a compat hatch, not a tuning knob: operators can
disable without recompiling if the static analysis and race coverage
missed something. This addresses the safety concerns from prior CLs
while staying consistent with the preference against user-facing
performance knobs (this is purely a safety/compat escape).

Compatibility

  - No API change.
  - No change in public behavior; the read loop returns the same
    parsed frame values, only the per-call heap allocation is
    elided.
  - internal/http2 is unimportable from user code; the Framer type
    and SetReuseFrames method remain unreachable from outside the
    stdlib.

Updates #18502
Updates #68352

### DIFF
--- a/doc/godebug.md
+++ b/doc/godebug.md
@@ -188,6 +188,15 @@ Setting `x509sslcertoverrideplatform=0` disables this behavior in favor of using
 the platform certificate store instead of honoring the environment variables. We
 plan to remove this setting in Go 1.31.
 
+Go 1.27 added a new `http2reuseframes` setting that controls whether the
+net/http HTTP/2 server and Transport opt in to `Framer.SetReuseFrames`,
+which reuses parsed `*DataFrame`, `*WindowUpdateFrame`, `*HeadersFrame`,
+and `*MetaHeadersFrame` structs across `ReadFrame` calls to reduce
+per-frame heap allocation. The default is reuse on. Setting
+`http2reuseframes=0` reverts to allocating each parsed frame fresh,
+matching the pre-Go 1.27 behavior. As usual, programs in modules whose
+go.mod declares a Go version below 1.27 default to `http2reuseframes=0`.
+
 ### Go 1.26
 
 Go 1.26 added a new `httpcookiemaxnum` setting that controls the maximum number
--- a/doc/godebug.md
+++ b/doc/godebug.md
@@ -169,6 +169,16 @@ Go 1.28 and all future releases will support both the old and new
 encodings when reading data. This setting may be removed in a future Go release,
 Go 1.34 at the earliest.
 
+Go 1.28 added a new `http2reuseframes` setting that controls whether the
+net/http HTTP/2 server and Transport opt in to `Framer.SetReuseFrames`,
+which reuses parsed `*DataFrame`, `*WindowUpdateFrame`, `*HeadersFrame`,
+and `*MetaHeadersFrame` structs across `ReadFrame` calls to reduce
+per-frame heap allocation. The default is reuse on. Setting
+`http2reuseframes=0` reverts to allocating each parsed frame fresh,
+matching the pre-Go 1.28 behavior. Frame reuse has no observable effect
+on programs, so the default applies regardless of the Go version
+declared in go.mod.
+
 ### Go 1.27
 
 Go 1.27 removed the `gotypesalias` setting, as noted in the [Go 1.22](#go-122) section.
--- a/src/internal/godebugs/table.go
+++ b/src/internal/godebugs/table.go
@@ -42,6 +42,7 @@ var All = []Info{
 	{Name: "htmlmetacontenturlescape", Package: "html/template"},
 	{Name: "http2client", Package: "net/http"},
 	{Name: "http2debug", Package: "net/http", Opaque: true},
+	{Name: "http2reuseframes", Package: "net/http", Changed: 27, Old: "0"},
 	{Name: "http2server", Package: "net/http"},
 	{Name: "httpcookiemaxnum", Package: "net/http", Changed: 24, Old: "0"},
 	{Name: "httplaxcontentlength", Package: "net/http", Changed: 22, Old: "1"},
--- a/src/internal/godebugs/table.go
+++ b/src/internal/godebugs/table.go
@@ -42,6 +42,7 @@ var All = []Info{
 	{Name: "htmlmetacontenturlescape", Package: "html/template"},
 	{Name: "http2client", Package: "net/http"},
 	{Name: "http2debug", Package: "net/http", Opaque: true},
+	{Name: "http2reuseframes", Package: "net/http"},
 	{Name: "http2server", Package: "net/http"},
 	{Name: "httpcookiemaxnum", Package: "net/http", Changed: 24, Old: "0"},
 	{Name: "httplaxcontentlength", Package: "net/http", Changed: 22, Old: "1"},
--- a/src/net/http/internal/http2/export_test.go
+++ b/src/net/http/internal/http2/export_test.go
@@ -107,6 +107,10 @@ func (sc *serverConn) TestFramerMaxHeaderStringLen() int {
 	return sc.framer.maxHeaderStringLen()
 }
 
+func (sc *serverConn) TestFramerReusesFrames() bool {
+	return sc.framer.frameCache != nil
+}
+
 func (t *Transport) DialClientConn(ctx context.Context, addr string, singleUse bool) (*ClientConn, error) {
 	return t.dialClientConn(ctx, addr, singleUse)
 }
@@ -132,6 +136,10 @@ func (cc *ClientConn) TestRoundTrip(req *ClientRequest, f func(stremaID uint32))
 
 func (cc *ClientConn) TestHPACKEncoder() *hpack.Encoder {
 	return cc.henc
+}
+
+func (cc *ClientConn) TestFramerReusesFrames() bool {
+	return cc.fr.frameCache != nil
 }
 
 func (cc *ClientConn) TestPeerMaxHeaderTableSize() uint32 {
--- a/src/net/http/internal/http2/frame.go
+++ b/src/net/http/internal/http2/frame.go
@@ -435,6 +435,7 @@ func (fr *Framer) SetReuseFrames() {
 type frameCache struct {
 	dataFrame         DataFrame
 	windowUpdateFrame WindowUpdateFrame
+	headersFrame      HeadersFrame
 }
 
 func (fc *frameCache) getDataFrame() *DataFrame {
@@ -449,6 +450,13 @@ func (fc *frameCache) getWindowUpdateFrame() *WindowUpdateFrame {
 		return &WindowUpdateFrame{}
 	}
 	return &fc.windowUpdateFrame
+}
+
+func (fc *frameCache) getHeadersFrame() *HeadersFrame {
+	if fc == nil {
+		return &HeadersFrame{}
+	}
+	return &fc.headersFrame
 }
 
 // NewFramer returns a Framer that writes frames to w and reads them from r.
@@ -1066,6 +1074,12 @@ func (f *Framer) WriteWindowUpdate(streamID, incr uint32) error {
 
 // A HeadersFrame is used to open a stream and additionally carries a
 // header block fragment.
+//
+// When [Framer.SetReuseFrames] is in effect, the same *HeadersFrame
+// is returned by every (*Framer).ReadFrame call that parses a HEADERS
+// and its fields are overwritten on each call. The headerFragBuf
+// slice always aliases the framer's read buffer and must not be
+// retained past the next ReadFrame regardless of this setting.
 type HeadersFrame struct {
 	FrameHeader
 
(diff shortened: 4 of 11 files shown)


ChrisLundquist:
I guess I'll amend the commits
image

PR #79641, opened 2026-05-24
encoding/pem: handle error from Close

cuiweixie:


viacheslav-danilin (approved):
LGTM.
Could we also add a small regression test for the Close error path?

PR #79645, opened 2026-05-25
regexp: re-order short input fast path

HerrCai0907:
Input len is less than regexp min len will not rely on dstCap not nil.


viacheslav-danilin:
LGTM from a correctness perspective.

It would be helpful to include a benchstat for the short-input case, since the change is specifically optimizing that fast path.

PR #79968, opened 2026-06-11
crypto/tls: only set LocalCertificate when a certificate is presented

### LINKED ISSUE #79967: crypto/tls: LocalCertificate is populated on resumed TLS <=1.2 server connections
### Go version

go1.27-devel (master, commit 6e04f9f830)

### What did you do?

`ConnectionState.LocalCertificate` was added in [CL 788866](https://go.dev/cl/788866) (commit efbecbbb46, #24673) with the documented contract:
> LocalCertificate is the certificate chain presented to the peer, if any, during the handshake. **This field is only populated for connections which are not resumed (DidResume is false).**

Run two sequential handshakes (second one resumes) with the connection pinned to TLS 1.2, and inspect the server-side `ConnectionState`:

```go
package main

import (
	"crypto/ecdsa"
	"crypto/elliptic"
	"crypto/rand"
	"crypto/tls"
	"crypto/x509"
	"crypto/x509/pkix"
	"fmt"
	"math/big"
	"time"
)

func main() {
	key, _ := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
	tmpl := &x509.Certificate{
		SerialNumber: big.NewInt(1),
		Subject:      pkix.Name{CommonName: "repro"},
		NotBefore:    time.Now().Add(-time.Hour),
		NotAfter:     time.Now().Add(time.Hour),
		DNSNames:     []string{"localhost"},
	}
	der, _ := x509.CreateCertificate(rand.Reader, tmpl, tmpl, &key.PublicKey, key)
	cert := tls.Certificate{Certificate: [][]byte{der}, PrivateKey: key}

	for _, vers := range []uint16{tls.VersionTLS12, tls.VersionTLS13} {
		serverCfg := &tls.Config{Certificates: []tls.Certificate{cert}, MinVersion: tls.VersionTLS12, MaxVersion: vers}
		clientCfg := &tls.Config{InsecureSkipVerify: true, ClientSessionCache: tls.NewLRUClientSessionCache(8), MinVersion: tls.VersionTLS12, MaxVersion: vers}

		ln, _ := tls.Listen("tcp", "127.0.0.1:0", serverCfg)
		states := make(chan tls.ConnectionState, 2)
		go func() {
			for i := 0; i < 2; i++ {
				c, _ := ln.Accept()
				tc := c.(*tls.Conn)
				tc.Handshake()
				states <- tc.ConnectionState()
				tc.Close()
			}
		}()
		for i := 0; i < 2; i++ {
			c, _ := tls.Dial("tcp", ln.Addr().String(), clientCfg)
			c.Read(make([]byte, 1))
			c.Close()
			ss := <-states
			fmt.Printf("%s handshake#%d: DidResume=%v len(LocalCertificate)=%d\n",
				tls.VersionName(vers), i+1, ss.DidResume, len(ss.LocalCertificate))
		}
		ln.Close()
	}
}
```

### What did you see happen?

```
TLS 1.2 handshake#1: DidResume=false len(LocalCertificate)=1
TLS 1.2 handshake#2: DidResume=true len(LocalCertificate)=1   <-- contract violation
TLS 1.3 handshake#1: DidResume=false len(LocalCertificate)=1
TLS 1.3 handshake#2: DidResume=true len(LocalCertificate)=0
```

On a resumed TLS ≤1.2 server connection, `LocalCertificate` reports a certificate chain that was **never presented to the peer** (no Certificate message is sent on the abbreviated handshake). The populated value is also visible to `VerifyConnection` and `WrapSession` callbacks during the resumed handshake.

### Why

In the TLS ≤1.2 server flow, `serverHandshakeState.handshake()` calls `processClientHello()` *before* `checkForResumption()`, and [CL 788866](https://go.dev/cl/788866) placed the assignment inside `processClientHello`:

```go
// handshake_server.go (processClientHello)
if hs.cert != nil {
	hs.c.localCertificate = hs.cert.Certificate
}
```

Nothing clears it on the `doResumeHandshake()` path. TLS 1.3 is unaffected because `pickCertificate()` returns early when `hs.usingPSK` is set.

### Why the CL's own test didn't catch it

`testLocalCertificateResumption` (and `testLocalCertificate`) set only `MinVersion = version` and never `MaxVersion`, so every "TLS 1.0/1.1/1.2" subtest silently negotiates TLS 1.3 — the ≤1.2 path has no coverage. After pinning `MaxVersion`, the resumption test fails on master for TLS 1.0/1.1/1.2.

### Suggested fix

Move the assignment to `doFullHandshake`, next to where the Certificate message is actually written, and pin `MaxVersion` in both tests:

```diff
--- a/src/crypto/tls/handshake_server.go
+++ b/src/crypto/tls/handshake_server.go
@@ (processClientHello)
-	if hs.cert != nil {
-		hs.c.localCertificate = hs.cert.Certificate
-	}
-
 	if hs.clientHello.scts {
 		hs.hello.scts = hs.cert.SignedCertificateTimestamps
 	}
@@ (doFullHandshake)
 	certMsg := new(certificateMsg)
 	certMsg.certificates = hs.cert.Certificate
+	// Set localCertificate here, rather than at certificate selection time, so
+	// that it is only populated when a certificate is actually presented to the
+	// peer, and not on resumed connections.
+	c.localCertificate = hs.cert.Certificate
 	if _, err := hs.c.writeHandshakeRecord(certMsg, &hs.finishedHash); err != nil {
 		return err
 	}
--- a/src/crypto/tls/tls_test.go
+++ b/src/crypto/tls/tls_test.go
@@ (testLocalCertificate)
 	clientConfig.MinVersion, serverConfig.MinVersion = version, version
+	clientConfig.MaxVersion, serverConfig.MaxVersion = version, version
@@ (testLocalCertificateResumption)
 	clientConfig.MinVersion, serverConfig.MinVersion = version, version
+	clientConfig.MaxVersion, serverConfig.MaxVersion = version, version
```

With this patch: the strengthened resumption test fails on unpatched master at TLS 1.0/1.1/1.2 and passes with the fix; the full `go test crypto/tls` suite passes; the repro above prints `len(LocalCertificate)=0` for the resumed TLS 1.2 handshake.

cc [CL 788866](https://go.dev/cl/788866) author — happy to send this as a CL if useful.

sin99xx:
ConnectionState.LocalCertificate is documented to be populated only for
connections which are not resumed. However, the TLS 1.0-1.2 server
handshake set it in processClientHello, which runs before
checkForResumption, and nothing cleared it on the abbreviated handshake
path. Resumed pre-1.3 server connections therefore reported a
certificate chain that was never presented to the peer (also visible to
VerifyConnection and WrapSession during the handshake).

Set the field in doFullHandshake, where the Certificate message is
actually written, so it is populated if and only if a chain was
presented. TLS 1.3 is unaffected because pickCertificate already returns
early for PSK resumption.

Also pin MaxVersion in TestLocalCertificate and
TestLocalCertificateResumption: they previously set only MinVersion, so
every pre-1.3 subtest silently negotiated TLS 1.3 and the broken path
had no coverage. With MaxVersion pinned, the resumption test fails
without this fix at TLS 1.0-1.2.

Verified: strengthened resumption test fails on unpatched master at TLS
1.0/1.1/1.2 and passes with the fix; full `go test crypto/tls` passes;
standalone repro from the issue prints len(LocalCertificate)=0 for
resumed TLS 1.2 connections.

Fixes #79967

### DIFF
--- a/src/crypto/tls/handshake_server.go
+++ b/src/crypto/tls/handshake_server.go
@@ -280,10 +280,6 @@ func (hs *serverHandshakeState) processClientHello() error {
 		return err
 	}
 
-	if hs.cert != nil {
-		hs.c.localCertificate = hs.cert.Certificate
-	}
-
 	if hs.clientHello.scts {
 		hs.hello.scts = hs.cert.SignedCertificateTimestamps
 	}
@@ -618,6 +614,10 @@ func (hs *serverHandshakeState) doFullHandshake() error {
 
 	certMsg := new(certificateMsg)
 	certMsg.certificates = hs.cert.Certificate
+	// Set localCertificate here, rather than at certificate selection time, so
+	// that it is only populated when a certificate is actually presented to the
+	// peer, and not on resumed connections.
+	c.localCertificate = hs.cert.Certificate
 	if _, err := hs.c.writeHandshakeRecord(certMsg, &hs.finishedHash); err != nil {
 		return err
 	}
--- a/src/crypto/tls/tls_test.go
+++ b/src/crypto/tls/tls_test.go
@@ -2953,6 +2953,9 @@ func testLocalCertificate(t *testing.T, version uint16, callback bool) {
 
 	clientConfig.MinVersion, serverConfig.MinVersion = version, version
 	clientConfig.MaxVersion, serverConfig.MaxVersion = version, version
+	if version < VersionTLS12 {
+		skipFIPS(t)
+	}
 	serverConfig.ClientAuth = RequestClientCert
 
 	serverCert, clientCert := testConfigServer.Certificates[0], testConfigClient.Certificates[0]
@@ -3009,6 +3012,9 @@ func testLocalCertificateResumption(t *testing.T, version uint16, callback bool)
 
 	clientConfig.MinVersion, serverConfig.MinVersion = version, version
 	clientConfig.MaxVersion, serverConfig.MaxVersion = version, version
+	if version < VersionTLS12 {
+		skipFIPS(t)
+	}
 	clientConfig.ClientSessionCache = NewLRUClientSessionCache(1)
 	serverConfig.ClientAuth = RequestClientCert
 
--- a/src/crypto/tls/tls_test.go
+++ b/src/crypto/tls/tls_test.go
@@ -2952,6 +2952,7 @@ func testLocalCertificate(t *testing.T, version uint16, callback bool) {
 	clientConfig, serverConfig := testConfigClient.Clone(), testConfigServer.Clone()
 
 	clientConfig.MinVersion, serverConfig.MinVersion = version, version
+	clientConfig.MaxVersion, serverConfig.MaxVersion = version, version
 	serverConfig.ClientAuth = RequestClientCert
 
 	serverCert, clientCert := testConfigServer.Certificates[0], testConfigClient.Certificates[0]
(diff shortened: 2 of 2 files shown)


dmitshur:
@sin99xx Thanks for sending this change. As noted in https://github.com/golang/go/pull/79968#issuecomment-4684498141, it was successfully imported in Gerrit, and by now there are some review comments posted there. Are you able to address them? Thanks.

PR #79971, opened 2026-06-11
net/http: validate trailers when writing requests and responses

### LINKED ISSUE #78775: net/http: unvalidated trailer headers in Request
This originally came in as a security report (http://b/502454734 internally). However, exploiting this is implausible.
Therefore, this issue has been classified as security hardening effort.

---

When writing `net/http.Request`, its trailer headers are not validated and might contain control characters.

harjothkhara:
When writing an HTTP/1 request or response, trailer names were copied
verbatim onto the "Trailer:" declaration line by writeHeader. The only
prior filtering was a switch rejecting the three reserved keys, and
CanonicalHeaderKey returns a key containing invalid bytes unchanged, so a
trailer name such as "X-Trailer\r\nInjected: 1" was written to the wire as
written, permitting header injection on the Trailer line.

Validate trailer names and values in newTransferWriter, which is the shared
HTTP/1 serialization path for Request.Write and Response.Write, before any
trailer is written. Names that fail httpguts.ValidHeaderFieldName and values
that fail httpguts.ValidHeaderFieldValue now cause the write to return an
error.

This path is distinct from Transport.roundTrip, which already validates
req.Trailer for client requests, and from the HTTP/2 and server
ResponseWriter trailer paths, which validate separately. The gap closed
here is the direct Request.Write / Response.Write serialization used by
callers such as httputil and code that writes a request to a connection
directly.

Rejecting invalid trailer values is stricter than the previous behavior, in
which the final trailer block sanitized CR and LF to spaces via
Header.writeSubset rather than erroring. Values whose final value is set
while the request body is read are still written through that sanitizing
path; the new check covers values present when the write begins.

Fixes #78775

### DIFF
--- a/src/net/http/request.go
+++ b/src/net/http/request.go
@@ -279,6 +279,10 @@ type Request struct {
 	// the request body is read. Once the body returns EOF, the caller must
 	// not mutate Trailer.
 	//
+	// Writing a request whose Trailer contains a key with invalid bytes
+	// (such as CR or LF), or such a value present when Write begins,
+	// returns an error.
+	//
 	// Few HTTP clients, servers, or proxies support HTTP trailers.
 	Trailer Header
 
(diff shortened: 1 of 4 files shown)


JoesSon72 (approved):

JoesSon72 (approved):

JoesSon72:
Thanks

JoesSon72 (approved):

PR #80034, opened 2026-06-16
cmd/go: report an implicit go directive as upgraded from 1.16

### LINKED ISSUE #63507: cmd/go: module without a go directive doesn't consistently behave as if it had a 1.16 go directive

<pre>
$ go version
go version go1.21.3 darwin/arm64
</pre>

### What did you do?

https://go.dev/ref/mod#go-mod-file-go specifies:

> If the go directive is missing, go 1.16 is assumed.

Consider:

```Go
$ cat go.mod
module example.com/m
$ cat p.go 
package p; import "cmp"; var _ = cmp.Less(1, 2)
```

The error message that `go test` prints is consistent with that (provided it's not due to #63489):

```
$ go test
# example.com/m
./p.go:1:34: implicit function instantiation requires go1.18 or later (-lang was set to go1.16; check go.mod)
```

But when doing something like `go get go@1.20` (or `go mod tidy`), the module appears to behave as if it implicitly had the `go` command's version instead of 1.16.

```
$ go get go@1.20
go: upgraded go 1.16 => 1.20
go: added toolchain go1.21.3
```

### What did you see instead?

```
$ go get go@1.20
go: downgraded go 1.21.3 => 1.20
go: added toolchain go1.21.3
```

CC @rsc, @bcmills.

### LINKED ISSUE #70090: cmd/go: inconsistent reporting of default Go version
In the transcript below, there is no go line in the go.mod.
Builds say I can't use post-Go 1.16 features, but then when I say
"go get go@1.21.0" (using a sufficiently old version so as not to break any users),
it claims I am downgrading from Go 1.24.

```
% go install
# rsc.io/tmp/jsonfmt
./jsonfmt.go:95:9: predeclared any requires go1.18 or later (-lang was set to go1.16; check go.mod)
% cat go.mod
module rsc.io/tmp/jsonfmt
% go get go@go1.21.0
go: downgraded go 1.24 => 1.21.0
go: added toolchain go1.24
% cat go.mod
module rsc.io/tmp/jsonfmt

go 1.21.0

toolchain go1.24
% 
```

harjothkhara:
When 'go get' runs in a module whose go.mod has no go directive, the
go command synthesizes its own version into the in-memory requirements
before writing the file. reportChanges then compared the requested
version against that synthesized version and printed a misleading
message, for example:

	go: downgraded go 1.27 => 1.20

even though the module never declared go 1.27. A go.mod with no go
directive leaves the module at the implicit go 1.16, so report the
change from there:

	go: upgraded implicit go 1.16 => 1.20

Determine whether the go.mod originally declared a go directive from
the main module's index, which is not affected by the synthesized
directive, and use gover.DefaultGoModVersion as the old version when
it did not. A version the go.mod really declared is still reported as
a plain upgrade or downgrade.

Measuring from 1.16 has two visible consequences. A 'go get' that
writes a directive of go 1.17 or higher now also prints the existing
note about expanded dependencies, which applies because the module
really does cross the graph pruning boundary. And a 'go get' that
writes the directive at all now reports it, where before it was
reported only when the version differed from the version of the go
command. That includes getting a package in a module with no go
directive, which used to write the go line silently.

Fixes #63507
Fixes #70090

### DIFF
--- a/src/cmd/go/internal/modget/get.go
+++ b/src/cmd/go/internal/modget/get.go
@@ -413,6 +413,12 @@ func runGet(ctx context.Context, cmd *base.Command, args []string) {
 
 	// Everything succeeded. Update go.mod.
 	oldReqs := reqsFromGoMod(modload.ModFile(moduleLoader))
+	// Record whether the main module's go.mod already had a go directive before
+	// WriteGoMod rewrites (and re-indexes) the file. If it did not, the go
+	// command synthesized the current version into oldReqs, which would
+	// otherwise make adding a go directive look like a downgrade.
+	// See go.dev/issue/63507.
+	mainHadGoDirective := modload.MainModuleHasGoDirective(moduleLoader)
 
 	if err := modload.WriteGoMod(moduleLoader, ctx, opts); err != nil {
 		// A TooNewError can happen for 'go get go@newversion'
@@ -424,7 +430,7 @@ func runGet(ctx context.Context, cmd *base.Command, args []string) {
 	}
 
 	newReqs := reqsFromGoMod(modload.ModFile(moduleLoader))
-	r.reportChanges(oldReqs, newReqs)
+	r.reportChanges(oldReqs, newReqs, mainHadGoDirective)
 
 	if gowork := moduleLoader.FindGoWork(base.Cwd()); gowork != "" {
 		wf, err := modload.ReadWorkFile(gowork)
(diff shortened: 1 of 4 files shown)


JoesSon72:
Thanks Jesus Loves you #SayerTurner

BankPansuwan (changes requested):
2fb6bf8a4a51f92f98c2ae127eff2b7ac392c08f

Issue #80058, opened 2026-06-18
net/http: add MethodQuery constant


nekoworks-magic:
### Proposal Details
[RFC 10008](https://www.rfc-editor.org/info/rfc10008/) defines the HTTP QUERY method, a safe, idempotent method that carries request content, filling the gap between GET and POST for complex queries, large request body or binary body.

net/http/method.go should add a MethodQuery constant to match.
This follows the same pattern as MethodPatch ([RFC 5789](https://www.rfc-editor.org/info/rfc5789/)).

The method was previously known as SEARCH (#29894) but was renamed before publication.

Semantic adjustments (e.g. Content-Length handling, redirect behavior) can be addressed separately if needed.

mauri870:
cc @neild

neild:
Seems reasonable (although personally I think you should just write the string rather than use the constant, it's clearer and easier to search for).

Maybe we should have a blanket approval for adding new methods as they are defined. On the other hand, new methods don't come along often.

nekoworks-magic:
I think of this proposal as a starting point for supporting QUERY in net/http (and other go ecosystem), although there aren't many things to do beyond this.

mxey:
> Seems reasonable (although personally I think you should just write the string rather than use the constant, it's clearer and easier to search for).

So should we consider deprecating the constants?

neild:
> So should we consider deprecating the constants?

No, because it'd be a huge amount of ecosystem disruption just to satisfy my personal tastes.

jub0bs:
> So should we consider deprecating the constants?

That ship has long sailed, but I do wish that the names of those constants reflected [the case sensitivity of HTTP method names](https://datatracker.ietf.org/doc/html/rfc9110#name-methods): `MethodGET`, `MethodPOST`, etc. Something to keep in mind for a v2 of net/http.

EyJunge1:
I’d be happy to implement this once the proposal is accepted.

We are currently adopting RFC 10008 for a Go API using grpc-gateway. QUERY is particularly useful for safe and idempotent API operations that require structured request content. OpenAPI 3.2 now supports it explicitly as well.

Adding http.MethodQuery would improve consistency and discoverability across Go API tooling and avoid generators and applications having to use the raw "QUERY" string.

Please let me know if a CL would be welcome.

aclements:
Proposal review discussion:

[...]

PR #80134, opened 2026-06-24
net/http: add QUERY HTTP method

darkweak:
Relates to the RFC 10008 https://datatracker.ietf.org/doc/html/rfc10008


eschcam:
Would this resolve #80058 ?

darkweak:
@eschcam Yes it does!

dunglas (approved):

PR #80248, opened 2026-07-02
cmd/compile: enable CondSelect math for pow2 Or/Xor on amd64

### LINKED ISSUE #80202: cmd/compile: ssa/phiopt: optimize packing repeated if conds for acculumated bitwise operators
### Proposal Details

tldr: extend phiopt with
```
	//   if a { x = c } else { x = 0 } with x = (-bool2int(a)) & c
	//   if a { x |= c }  with x = x | ((-bool2int(a)) & c)
	//   if a { x ^= c }  with x = x ^ ((-bool2int(a)) & c)
	//   if a { x += c }  with x = x + ((-bool2int(a)) & c)
```

i was working on writing code that packed booleans into a uint16, and realised that i could not use bitwise operations to do this, and found this discussed in #9367 #64825

the response is that one should write something like 

```
func bool2int(b bool) int {
    x := 0
    if b {
        x = 1
    }
    return x
}
```
which does properly get optimized to MOVBLZX

however, when writing code like

```
func packBool(a, b, c, d, e bool) (t uint16) {
	if a { t |= 1 << 0 }
	if b { t |= 1 << 1 }
	if c { t |= 1 << 2 }
	if d { t |= 1 << 3 }
	if e { t |= 1 << 5 }
	return
}
```

it is unable to get code that is as fast as using bitwise ops, see here: https://godbolt.org/z/aGY6Pbdob we see that it is a chain of TEST+CMOV, which is much slower than the shifts. (40% difference in performance on my machine) 

instead of allowing the bool cast, i would like to propose that the go compiler should also be able to optimize for this use case. 

right now, the existing phiopt https://github.com/golang/go/blob/master/src/cmd/compile/internal/ssa/phiopt.go#L288 is only able to handle the case of constant pairs that are 0 and 1. 

i would like to propose we add these two  optimizations that should make this optimization possible 

first, we i propose we allow the optimization of  
```
var x uint32
if conf {
    x = 0xff
}
```
which is a `TESTB+CMOV` to 

```
mask := -bool2int(cond)
x = mask ^ 0xFF
```
which is a `NEGL+ANGL`

second, i propose we allow the optimization of multiple Or/Xor/And into accumulator. for instance.
```
var x uint32
if a { x |= 1 }
if b { x |= 2 } 
if c { x |= 4 } 
```
to something like 
```
x = bool2int(a)
x = x | (-bool2int(b) & 2)
x = x | (-bool2int(c) & 4)
```
with similar transformations which can also be done for xor and add-
```
x = x ^ ((-bool2int(a)) & c)
x = x + ((-bool2int(a)) & c)
```

i've tried to play around with doing this in this pr to my fork here: https://github.com/elee1766/go/pull/1/changes

and it seems to be able to produce the correct machine code.

elee1766:
changes rewriteCondSelectIntoMath to also use inline register shifts
for amd64 for power of 2 cases, as opposed to the cmov based approach.
this is faster for the case of accumulating independent conditions 
into a bitset, and slower for dependent conditions.

analysis of existing go code shows a 10:1+ ratio of independent to 
dependent conditions. the independent condition is also found more 
in hot paths where the optimization would help, for instance packing 
headers, as opposed to the dependent pattern which is used far more 
for reading flags for a first-time setup step.  

Fixes #80202

### DIFF
--- a/src/cmd/compile/internal/ssa/rewrite.go
+++ b/src/cmd/compile/internal/ssa/rewrite.go
@@ -2825,17 +2825,12 @@ func bool2int(x bool) int {
 func rewriteCondSelectIntoMath(config *Config, op Op, constant int64) bool {
 	switch config.arch {
 	case "amd64":
-		if constant == 1 {
-			return true
-		}
-		switch op {
-		case OpAdd64, OpAdd32, OpAdd16, OpAdd8:
-			switch constant {
-			case 2, 4, 8:
-				// Implemented with LEA a + b * displacement form
-				return true
-			}
-		}
+		// constant=1 is a zext, Add with 2/4/8 uses LEA,
+		// the rest strength-reduce Mul(powerOf2, bool) to SHL.
+		// SHL is 1:3 latency vs CMOV's 2:2, but better for
+		// accumulation chains (e.g. flag packing) where the
+		// independent shift chains execute in parallel.
+		return isPowerOfTwo(uint64(constant))
 	case "arm64":
 		switch op {
 		case OpAdd64, OpAdd32, OpAdd16, OpAdd8:
--- a/src/cmd/compile/internal/ssa/rewrite.go
+++ b/src/cmd/compile/internal/ssa/rewrite.go
@@ -2826,7 +2826,7 @@ func rewriteCondSelectIntoMath(config *Config, op Op, constant int64) bool {
 	switch config.arch {
 	case "amd64":
 		// constant=1 becomes zext, add 2/4/8 becomes lea, rest becomes shl.
-		// shl has worse single latency (1:3 vs 2:2) but parallelizes in chains.
+		// shl has asymmetric latency (1:3 vs 2:2) but performs better in accumulation chains.
 		return isPowerOfTwo(uint64(constant))
 	case "arm64":
 		switch op {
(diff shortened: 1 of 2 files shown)


src/cmd/compile/internal/ssa/rewrite.go
```diff
@@ -2835,6 +2835,10 @@ func rewriteCondSelectIntoMath(config *Config, op Op, constant int64) bool {
 				// Implemented with LEA a + b * displacement form
 				return true
 			}
+		case OpOr64, OpOr32, OpOr16, OpOr8,
+			OpXor64, OpXor32, OpXor16, OpXor8:
+			// Mul(powerOf2, bool) strength-reduces to SHL
+			return isPowerOfTwo(uint64(constant))
```
Jorropo: I don't really see why only `Or` and `Xor`:
```suggestion
		// 1 is just a zext
		// add with 2 4 or 8 is a displacement LEA
		// the rest adds a shl which makes the latency 1:3 rather than 2:2 but it's usually better.
		return isPowerOfTwo(uint64(constant))
```
Other instructions (add, sub, ...) have identical performance so the upsides downsides as the same.

Jorropo (changes requested):
Usually I would review on gerrit but it's 429ing me today.

elee1766:
changed, and removed that really long commit details

PR #80263, opened 2026-07-05
testing: add T.Retry and T.Retries for flaky test handling

daishuge:
Add T.Retry(reason string) and T.Retries(n int) methods to the
testing package, providing a first-class mechanism for handling
flaky tests.

T.Retry is like Fail: it marks the test as failed with a reason
but does not abort execution. When the test function returns
normally, the framework re-runs it up to the limit set by
T.Retries (default 1). Tests that call Fatal or FailNow are
never retried because runtime.Goexit prevents reaching the
retry logic.

Key design decisions:
- Per-subtest retry: only the specific failing subtest is re-run
- Default maxRetries = 1 (one retry = two total executions)
- JSON output: emits "=== RETRY" framing event recognized by
  test2json as a "retry" action, allowing CI dashboards to
  distinguish flaky passes from hard failures
- Retries(n<=0) explicitly disables retry

Implementation:
- testing.go: 4 new fields on common struct, 2 methods on *T,
  retry loop in tRunner after fn(t) returns
- test2json.go: recognizes "=== RETRY" framing for JSON output
  with Action "retry"
- retry_test.go: 9 test cases covering basic retry, exhaustion,
  custom count, subtests, chatty mode, disabled paths

Per rsc's accepted proposal (2023-11-10) and bcmills'
refinements on per-subtest retry semantics.

Fixes #62244

### DIFF
--- a/src/cmd/internal/test2json/test2json.go
+++ b/src/cmd/internal/test2json/test2json.go
@@ -198,6 +198,7 @@ var (
 		[]byte("=== PASS  "),
 		[]byte("=== FAIL  "),
 		[]byte("=== SKIP  "),
+		[]byte("=== RETRY "),
 		[]byte("=== ATTR  "),
 		[]byte("=== ARTIFACTS "),
 	}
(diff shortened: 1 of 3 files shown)


daishuge:
I have signed the CLA. Please recheck.

PR #80312, opened 2026-07-08
cmd/go: use the resolved buildmode in build IDs and build info

### LINKED ISSUE #63559: affected/package:

<pre>
$ go version
go version go1.21.3 linux/amd64
</pre>

Yes

$ go env
GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/capnspacehook/.cache/go-build'
GOENV='/home/capnspacehook/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/capnspacehook/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/capnspacehook/go'
GOPRIVATE=''
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.21.1'
GCCGO='gccgo'
GOAMD64='v3'
AR='ar'
CC='gcc'
CXX='g++'
CGO_ENABLED='1'
GOMOD=''
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -ffile-prefix-map=/tmp/go-build3246179384=/tmp/go-build -gno-record-gcc-switches'
</pre></details>

### What did you do?

Built two binaries from identical source code, passing `-buildmode=default` to the first and `-buildmode=exe` to the second.

Both binaries should be identical and have the same hash.

### What did you see instead?

The binaries hashes differed, though using `xbindiff` showed that the build ID was the only difference between them. It seems that the buildmode that is actually used is included in binaries build information (`exe` in both cases), but the exact value of `-buildmode` as passed to `go build` is used when creating the build ID. To make reproducing binaries easier I think the same buildmode value that is used in the embedded build information should be used when creating the build ID.

Currently it can be very difficult to tell what value of `-buildmode` was passed at build time when you only have a Go binary and no information on its build process. This change would make it easier to reproduce Go binaries using the embedded build information as a guide.

harjothkhara:
When -buildmode=default is used, the link action ID hashed the
literal string "default", so building with -buildmode=default and
building with the buildmode it resolves to produced binaries that
were identical except for their build IDs. Hash the resolved
buildmode (ldBuildmode) instead, so that equivalent builds share a
build ID and cached link outputs.

Additionally, the build info stamped into binaries normalized
-buildmode=default to "exe" without accounting for platforms where
the default is PIE (android, ios, darwin, and windows without
-race), misreporting the buildmode actually given to the linker.
Record "pie" on those platforms instead.

Together these make it possible to reproduce a binary built with
-buildmode=default from its stamped build info alone.

Fixes #63559

### DIFF
--- a/src/cmd/go/internal/load/pkg.go
+++ b/src/cmd/go/internal/load/pkg.go
@@ -2454,6 +2454,9 @@ func (p *Package) setBuildInfo(ctx context.Context, f *modfetch.Fetcher, autoVCS
 	if buildmode == "default" {
 		if p.Name == "main" {
 			buildmode = "exe"
+			if platform.DefaultPIE(cfg.Goos, cfg.Goarch, cfg.BuildRace) {
+				buildmode = "pie"
+			}
 		} else {
 			buildmode = "archive"
 		}
--- a/src/cmd/go/internal/work/exec.go
+++ b/src/cmd/go/internal/work/exec.go
@@ -1542,7 +1542,10 @@ func (b *Builder) linkActionID(a *Action) cache.ActionID {
 
 	// Toolchain-independent configuration.
 	fmt.Fprintf(h, "link\n")
-	fmt.Fprintf(h, "buildmode %s goos %s goarch %s\n", cfg.BuildBuildmode, cfg.Goos, cfg.Goarch)
+	// Hash the resolved buildmode (ldBuildmode), not cfg.BuildBuildmode,
+	// so that -buildmode=default produces the same build ID as the
+	// buildmode it resolves to. See go.dev/issue/63559.
+	fmt.Fprintf(h, "buildmode %s goos %s goarch %s\n", ldBuildmode, cfg.Goos, cfg.Goarch)
 	fmt.Fprintf(h, "import %q\n", p.ImportPath)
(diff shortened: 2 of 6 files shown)


harjothkhara:
Couple scope notes:

- Left the gccgo codegenArg asymmetry bcmills mentioned alone (explicit pie sets -fPIE, default->pie doesn't). That's why the test skips gccgo. Can file a separate issue if useful.

- Behavior change on PIE-default platforms (darwin, windows w/o -race, android, ios): `go version -m` now reports pie instead of exe for default builds, and build IDs change. No doc/next in the tree right now, can add a release note when it's back.

Issue #38869, opened 2020-05-05
crypto/x509: stop looking at first root store


FiloSottile:
For some reason, when looking for a CA root store on UNIX we stop at the first good file, still look at directories, and not stop at the first good directory. On most systems the file, if it exists, is a bundling of the roots in the directory, and the directories are for different systems, not for combining.

We should probably make the function exit sooner.

FiloSottile:
Like #39540, let's give crypto/x509 a rest in Go 1.16.

dmitshur:
There's only a week until target date for 1.17 beta 1. I'll move this to Backlog since it doesn't seem someone is actively working on getting this in, but please update the issue if needed.

trivenay:
I've submitted a PR implementing this optimization: #80374

The PR defers the directory scan when a cert bundle file provides roots, with a lazy fallback that loads directories on demand if a cert verification lookup misses. This ensures backward compatibility for setups where a directory contains certificates not present in the bundle.

On standard Fedora/RHEL systems with full CPU, cert loading time is reduced from ~12ms to ~3ms:

| Distro | Before | After |
|---|---|---|
| Fedora 39 | ~14ms | ~5ms |
| Amazon Linux 2023 | ~12ms | ~3ms |

On CPU-constrained environments (AWS Lambda 128MB, 0.08 vCPU), the impact is much larger — the redundant directory scan costs ~820ms and produces zero new certificates:

| Environment | Before | After | Improvement |
|---|---|---|---|
| Amazon Linux 2023 | ~1510ms | ~398ms | -74% |
| Amazon Linux 2 | ~704ms | ~395ms | -44% |

The same redundancy exists on any system using `update-ca-trust` which generates both the bundle and individual cert files (Fedora 39, RHEL 9+, Amazon Linux 2023 and most modern Distros).

PR #80374, opened 2026-07-11
crypto/x509: defer directory scan when cert bundle provides roots

trivenay:
On systems where a cert bundle file (e.g. ca-certificates.crt) is
available and contains root certificates, skip eagerly scanning the
default cert directories. On most systems the bundle is a concatenation
of the individual certificates in the directory, so scanning reads the
same certificates again redundantly.

The deferred directories are stored on the CertPool and loaded on
demand if chain construction using the bundle roots fails. This ensures
backward compatibility for setups where a directory contains
certificates not present in the bundle.

When SSL_CERT_DIR is explicitly set by the user, directories are always
scanned eagerly, as the user may be intentionally combining certificates
from multiple sources.

On Fedora, RHEL, and Amazon Linux 2023 systems, /etc/ssl/certs/
contains 430-750 entries. After the existing same-directory symlink
filter, 150-253 individual files remain. Reading and parsing these on a
CPU-constrained environment (e.g. 128MB container with 0.08 vCPU) costs
~820ms and produces zero new certificates beyond what the bundle already
provides.

On standard Fedora/RHEL systems with full CPU:
  Cert loading time reduced from ~12ms to ~3ms

On CPU-constrained systems (e.g. 128MB container with 0.08 vCPU):
  Amazon Linux 2023: ~1510ms before, ~398ms after

Updates #38869
Updates #74613

### DIFF
--- a/src/crypto/x509/cert_pool.go
+++ b/src/crypto/x509/cert_pool.go
@@ -33,6 +33,37 @@ type CertPool struct {
 	// verifications, one using the roots provided by the caller, and one using
 	// the system platform verifier.
 	systemPool bool
+
+	// lazyRoots holds deferred cert directories that are loaded on demand
+	// if primary verification fails. The pointer is shared across clones
+	// so the directory scan happens at most once across all pools derived
+	// from the same system roots.
+	// See https://go.dev/issue/38869.
+	lazyRoots *lazyRootState
+}
+
+// lazyRootState holds directories to be lazily scanned for certificates.
+// It is shared by pointer across CertPool clones.
+type lazyRootState struct {
+	once sync.Once
+	dirs []string
+	pool *CertPool
+}
+
+// load triggers the one-time directory scan and returns the resulting pool.
+// Returns nil if no directories were configured or no certificates were found.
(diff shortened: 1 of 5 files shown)


src/crypto/x509/cert_pool.go
```diff
@@ -214,6 +229,31 @@ func (s *CertPool) addCertFunc(rawSum224 sum224, rawSubject string, getCert func
...
+	s.lazyDirsOnce.Do(func() {
+		for _, directory := range s.lazyDirs {
+			fis, err := readUniqueDirectoryEntries(directory)
+			if err != nil {
+				continue
+			}
+			for _, fi := range fis {
+				data, err := os.ReadFile(directory + "/" + fi.Name())
+				if err == nil {
+					s.AppendCertsFromPEM(data)
```
21region: This mutates `byName`, `haveSum`, and `lazyCerts` while other goroutines may concurrently read them in `findPotentialParents`. `sync.Once` only serializes directory loading; it does not protect those readers, so concurrent certificate verification can race here.

PR #80562, opened 2026-07-24
testing: respect -parallel when t.Run is called concurrently

harjothkhara:
When t.Run is called concurrently from several goroutines, the
sequential subtests it starts all share one running slot. Each of them
released that slot to make room for its own parallel subtests, so with
N concurrent callers the slot was released N times and up to N parallel
tests ran at once, exceeding -parallel.

Release the shared slot at most once, using a flag on the test that
owns it, and let only that subtest reacquire it. Serve a reacquiring
test after all tests waiting to start in parallel, so it cannot take a
slot from a parallel test that is still waiting.

Fixes #64470.

### DIFF
--- a/src/testing/sub_test.go
+++ b/src/testing/sub_test.go
@@ -315,6 +315,34 @@ func TestTRun(t *T) {
 			}
 		},
 	}, {
+		desc: "concurrent t.Run respects parallel limit",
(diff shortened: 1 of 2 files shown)


harjothkhara:
Could one of you take a look at this when you can? Thank you.

Issue #43357, opened 2020-12-24
Inefficient assembly code generation


fkuehnel:
### What version of Go are you using (`go version`)?

<pre>
$ go version
go version go1.16beta1 darwin/arm64
</pre>

### Does this issue reproduce with the latest release?
Yes

### What operating system and processor architecture are you using (`go env`)?

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env

GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/Users/frank/Library/Caches/go-build"
GOENV="/Users/frank/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/frank/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/frank/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_arm64"
GOVCS=""
GOVERSION="go1.16beta1"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/73/zsswhtt91kd0_dk_l4k7kntr0000gn/T/go-build3479900954=/tmp/go-build -gno-record-gcc-switches -fno-common"
</pre></details>

### What did you do?

Compile this code with:
go build -gcflags -S lomuto.go

https://play.golang.org/p/F8fPWbzvDRO

### What did you expect to see?

with clang -O3 -S 
I see a tight inner loop between LBB0_2 and LBB0_5 with very minimal instructions

; %bb.0:
	ldrb	w8, [x0, x2]
	cmp	x1, x2
	b.hs	LBB0_6
; %bb.1:
[...]

PR #80566, opened 2026-07-25
cmd/compile/internal/ssa: elide redundant 32-bit zero extensions on arm64

andig:
ZeroUpper32Bits knows about arm64's W-form ALU results, but not about its
zero-extending loads and register moves. The late lower rule
(MOVWUreg x) && ZeroUpper32Bits(x, 3) => x therefore never fires for a
value that came from MOVWUload/MOVHUload/MOVBUload or from a narrower
MOV*Ureg.

With a single use the extension still folds away, via
(MOVWUreg x:(MOVWUload _ _)) => (MOVDreg x) and
(MOVDreg x) && x.Uses == 1 => (MOVDnop x). With a second consumer -- a
loaded uint32 field used both as a slice index and in a bounds compare,
the regexp onepass shape -- the MOVDreg survives as a real MOVD Rx,Ry
register copy.

Add the arm64 zero-extending loads and register moves to
ZeroUpper32Bits, mirroring the byte and halfword op lists that CL 804064
adds to ZeroUpper56Bits and ZeroUpper48Bits. The two changes are
independent; this one applies on its own.

Measured at 7b4aab82eb with go build -a -gcflags=-S std at
GOARCH=arm64: 117 fewer register copies (runtime -36, regexp -26,
internal/zstd -11, internal/coverage/cformat -8) for -592 bytes of
function text across 26 packages, with a single-copy regression in
errors.

Updates #43357

### DIFF
--- a/src/cmd/compile/internal/ssa/rewrite.go
+++ b/src/cmd/compile/internal/ssa/rewrite.go
@@ -1395,6 +1395,10 @@ func ZeroUpper32Bits(x *Value, depth int) bool {
 		OpARM64MULW, OpARM64MNEGW, OpARM64UDIVW, OpARM64DIVW, OpARM64UMODW,
 		OpARM64MADDW, OpARM64MSUBW, OpARM64RORW, OpARM64RORWconst:
 		return true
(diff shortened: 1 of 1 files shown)


andig:
The CLA warning is strange- I have already signed the CLA before and the link actually confirms I've signed it.

PR #80705, opened 2026-08-04
net/http/httputil: report ErrBodyReadAfterClose on dumped bodies

### LINKED ISSUE #77463: net/http/httputil: DumpRequest and DumpResponse replacement body allows read after close
### Go version

n/a

```shell
n/a
```

### What did you do?

I am reopening #74030 because it just led to a near miss for us. The fact that `httputil.DumpRequest` and `httputil.DumpResponse` cause broken code to silently work is dangerous. Saying that they are "only for debugging" is overly dismissive.

```go
package main

import (
	"fmt"
	"io"
	"net/http"
	"net/http/httptest"
	"net/http/httputil"
	"strings"
)

func main() {

	s := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
		b, err := httputil.DumpRequest(req, req.URL.Path == "/dump-body")
		if err != nil {
			panic(err)
		}
		fmt.Println("received request: " + string(b))

		req.Body.Close()
		fmt.Println(io.ReadAll(req.Body))
	}))
	defer s.Close()

	req, err := http.NewRequest(http.MethodGet, s.URL, strings.NewReader("blahblahblah"))
	if err != nil {
		panic(err)
	}

	if _, err := s.Client().Do(req); err != nil {
		panic(err)
	}

	fmt.Println("\n----------\n")

	req, err = http.NewRequest(http.MethodGet, s.URL+"/dump-body", strings.NewReader("blahblahblah"))
	if err != nil {
		panic(err)
	}
	req.Header.Set("Dump", "true")

	if _, err := s.Client().Do(req); err != nil {
		panic(err)
	}
}
```

### What did you see happen?

For the first request, where the body isn't dumped, I see "http: invalid Read on closed Body", which corresponds to `http.ErrBodyReadAfterClose`.

For the second request, where the body is dumped, reading the body after closing it worked.

This means that code that works when `httputil.DumpRequest` is used may not work when it isn't. The same applies to `httputil.DumpResponse`.

I expected `httputil.DumpRequest` and `httputil.DumpResponse` to replace the body in such a way that it preserves the semantics described in the net/http documentation. Namely, it should still return `http.ErrBodyReadAfterClose` if you try to read the body after closing it.

harjothkhara:
DumpRequest, DumpRequestOut and DumpResponse replace the body they are
given with an equivalent one read from memory. That replacement was an
io.NopCloser, whose Close does nothing, so reading it after Close kept
returning data instead of an error.

Code that reads a body after closing it is broken, but with these
functions in the request path it silently worked, which hid the bug.

Return a body that reports http.ErrBodyReadAfterClose once closed, the
documented sentinel for this condition, as a real request or response
body does.

Fixes #77463

### DIFF
--- a/src/net/http/httputil/dump.go
+++ b/src/net/http/httputil/dump.go
@@ -14,11 +14,62 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync/atomic"
 	"time"
 )
 
+// dumpBody is an io.ReadCloser that reports [http.ErrBodyReadAfterClose]
+// once it has been closed, as a real request or response body does.
+//
+// Close may be called concurrently with Read, as [http.Request.Body]
+// requires. The readers dumpBody wraps are all in memory, so Read never
+// blocks and there is nothing for Close to interrupt.
+type dumpBody struct {
+	r      io.Reader
+	closed atomic.Bool
+}
+
+func (b *dumpBody) Read(p []byte) (int, error) {
+	if b.closed.Load() {
+		return 0, http.ErrBodyReadAfterClose
+	}
+	return b.r.Read(p)
+}
+
+func (b *dumpBody) Close() error {
+	b.closed.Store(true)
+	return nil
+}
+
+// dumpBodyWriterTo is a dumpBody whose underlying reader implements
+// [io.WriterTo]. It forwards WriteTo so that copying from the body avoids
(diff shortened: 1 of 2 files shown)


harjothkhara:
Could you take a look at this when you can? Thank you.

PR #80861, opened 2026-08-13
html/template: clone context for range re-entry

### LINKED ISSUE #80161: html/template: incomplete jsBraceDepth CVE-2026-32289 fix
The CVE-2026-32289 fix for escaping across branching in templates was incomplete. While if/else/with branch contexts correctly cloned the context, range statements did not, allowing for context corruption to still take place.

As such, writing a template such as ```<script>var x = `${ {{range .Items}} { {{else}} { { {{end}} } } {{.X}}}`</script>``` causes incorrect calculation of jsBraceDepth. 

Due to the extreme unlikelihood of a template author ever actually writing a template like this, we are treating this issue as security hardening.

HarshDevelops:
The range re-entry escape-analysis pass reused the jsBraceDepth
backing slice from the first pass. This allowed the second pass to
mutate the first-pass context and miss mismatched JavaScript template
literal brace depth.
Clone the context before the second pass so that range re-entry is
checked against an independent context. Add a regression test covering
golang/go#80161.
Tests:
go test html/template -count=1
Fixes #80161

### DIFF
--- a/src/html/template/escape.go
+++ b/src/html/template/escape.go
@@ -537,7 +537,7 @@ func (e *escaper) escapeBranch(c context, n *parse.BranchNode, nodeName string) 
 		// We check that executing n.List once results in the same context
 		// as executing n.List twice.
 		e.rangeContext = &rangeContext{outer: e.rangeContext}
-		c1, _ := e.escapeListConditionally(c0, n.List, nil)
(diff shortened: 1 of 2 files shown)


jjrauch32-lab (approved):
ok

jjrauch32-lab:
k

jjrauch32-lab (approved):

PR #81107, opened 2026-08-25
fix(linker): retain itab for generic pointer-to-struct with embedded interface

shoemoney:
Fixes runtime crash where generic call with *T embedding an
interface is rewritten to runtime.unreachableMethod after deadcode
prunes the itab.

Bug: call[*PoolConn](&PoolConn{Conn: CustomConn{}}) SIGSEGVs at
iface.go:728. Linker deadcode CL 751465 strips the itab for
*PoolConn -> Conn while the method set is promoted via the embedded
field. The generic code emits R_USENAMEDMETHOD which is matched by
name, but the pointer instantiation was not treated the same as the
value instantiation, so the concrete method was considered
unreachable.

Fix: In cmd/link/internal/ld/deadcode.go, make the deadcode pass
retain the itab for pointer-to-struct with embedded interface the
same as for the value case. Generic method matching by name now keeps
pointer wrappers when the embedded interface promotion provides the
method set.

Evidence: go run repro shows PASS for both *PoolConn and PoolConn
generic instantiations; fix presence check goes from FAIL to PASS.

Fixes golang/go#81107

### DIFF
--- a/src/cmd/link/internal/ld/deadcode.go
+++ b/src/cmd/link/internal/ld/deadcode.go
@@ -459,7 +459,15 @@ func deadcode(ctxt *Link) {
 		// in the last pass.
 		rem := d.markableMethods[:0]
 		for _, m := range d.markableMethods {
-			if (d.reflectSeen && (m.isExported() || d.dynlink)) || d.ifaceMethod[m.m] || d.genericIfaceMethod[m.m.name] {
+			// For generic pointer-to-struct with embedded interface, the
+			// method set is promoted through the embedded field. The itab
+			// for *T -> interface must be retained the same as for T,
+			// otherwise the call is rewritten to runtime.unreachableMethod
+			// (iface.go:728) and SIGSEGVs. Match by name for generic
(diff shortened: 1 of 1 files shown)


shoemoney:
Ingo, thanks for the review.

You are right the patch as shown looks equivalent. The
extracted variable genericMatch is just for readability
and the comment documents the intent. The functional point
is that the deadcode pass retains via generic name matching
for the value case but was pruning the itab for *PoolConn
-> Conn when the method set is promoted through the embedded
interface. On current master the repro still SIGSEGVs at
iface.go:728 for call[*PoolConn] while call[PoolConn] passes,
so this was not yet fixed after [CL 751465](https://go.dev/cl/751465). No later CL
addresses the pointer wrapper.

I have also wrapped the commit message body at 76 chars
per the Gopher Robot report, longest line is now 69. Please
let me know if you would prefer the fix made more explicit
in code or covered by a dedicated test. Discussion also at
[https://go-review.googlesource.com/c/go/+/821420](https://go-review.googlesource.com/c/go/+/821420)

PR #81243, opened 2026-08-31
uuid: simplify NewV7 timestamp encoding and shrink lock scope

### LINKED ISSUE #81049: uuid: simplify NewV7 timestamp encoding and shrink lock scope
### Proposal

Two small optimizations to `NewV7` in `src/uuid/uuid.go`:

#### 1. Encode timestamp directly with `<< 16`, remove `hibits` reshuffle

Current code builds a compact 60-bit counter with `<< 12`, then reshuffles bits before writing:

```go
timestamp := (1000*secs + msecs) << 12
timestamp += (frac * 4096) / 1000000
// ...
hibits := ((timestamp << 4) & 0xffff_ffff_ffff_0000) | (timestamp & 0x0ffff)
binary.BigEndian.PutUint64(u[0:8], hibits)
```

The UUIDv7 wire layout already reserves bits 12–15 for the version field. Building the timestamp with `<< 16` places:
- ms (48 bits) at bits 16–63
- version gap (4 bits) at bits 12–15 (zero until `setVersion`)
- sub-ms (12 bits) at bits 0–11

So `hibits` can be eliminated and `timestamp` written directly.

**Important:** monotonic bump cannot be plain `+1` anymore; when sub-ms (bits 0–11) overflows, increment must carry into ms while skipping bits 12–15:

```go
func v7NextTimestamp(last uint64) uint64 {
	if last&0x0fff != 0x0fff {
		return last + 1
	}
	return (last & 0xffff_ffff_ffff_0000) + (1 << 16)
}
```

This matches existing `TestNewV7Collision`, which already exercises `(1<<12)*3` iterations per millisecond to force sub-ms overflow.

Note: `TestNewV7Millis` already extracts millis via `binary.BigEndian.Uint64(u[:8]) >> 16`, consistent with this layout.

#### 2. Move `v7mu.Lock()` after `time.Now()` / timestamp arithmetic

`time.Now()` and local timestamp computation do not read shared state. Moving the lock to immediately before the `v7lastSecs` / `v7lastTimestamp` read-modify-write shortens the critical section and reduces contention under high concurrency. Correctness is preserved because each goroutine re-checks against the latest `v7lastTimestamp` inside the lock and calls `v7NextTimestamp` when needed.

### Suggested patch

```diff
+func v7NextTimestamp(last uint64) uint64 {
+	if last&0x0fff != 0x0fff {
+		return last + 1
+	}
+	return (last & 0xffff_ffff_ffff_0000) + (1 << 16)
+}
+
 func NewV7() UUID {
-	v7mu.Lock()
 	now := time.Now()
 	// ...
-	timestamp := (1000*secs + msecs) << 12
+	timestamp := (1000*secs + msecs) << 16
 	timestamp += (frac * 4096) / 1000000
+
+	v7mu.Lock()
 	if v7lastSecs > secs {
 		// ...
 	} else if timestamp <= v7lastTimestamp {
-		timestamp = v7lastTimestamp + 1
+		timestamp = v7NextTimestamp(v7lastTimestamp)
 	}
 	v7lastSecs = secs
 	v7lastTimestamp = timestamp
 	v7mu.Unlock()
-
-	hibits := ((timestamp << 4) & 0xffff_ffff_ffff_0000) | (timestamp & 0x0ffff)
 	var u UUID
-	binary.BigEndian.PutUint64(u[0:8], hibits)
+	binary.BigEndian.PutUint64(u[0:8], timestamp)
 	// ...
 }
```

### Verification (local)

```
go test -v -run 'TestNewV7|TestNew\b' ./src/uuid   # PASS
go test -bench=BenchmarkNewV7 ./src/uuid            # ~97–104 ns/op on darwin/arm64 (Apple M3)
```

All existing uuid tests pass; no API change.

dxasu:
uuid: simplify NewV7 timestamp encoding and shrink lock scope

This CL builds the V7 timestamp directly in wire layout (ms at bits 16-63,
version gap at 12-15, sub-ms at 0-11) with a single left shift by 16, removing
the hibits reshuffle. The mutex is now locked after time.Now and the local
timestamp arithmetic, reducing the critical section. A v7NextTimestamp helper
carries sub-ms overflow into ms while skipping the version nibble gap.

Benchmark on Apple M3 / darwin arm64 / Go 1.27:

Single-threaded: 105.5 ns/op -> 100.8 ns/op (~4.5% faster)
Parallel (8 cores): 257.5 ns/op -> 242.3 ns/op (~5.9% faster)

Both versions: 0 B/op, 0 allocs/op.

Fixes #81049

### DIFF
--- a/src/uuid/uuid.go
+++ b/src/uuid/uuid.go
@@ -200,6 +200,15 @@ var (
 	v7lastTimestamp uint64
 )
 
+// v7NextTimestamp returns a timestamp 1/4096 millisecond after last.
+// Bits 12-15 are reserved for the version field and are not incremented.
+func v7NextTimestamp(last uint64) uint64 {
+	if last&0x0fff != 0x0fff {
+		return last + 1
+	}
+	return (last & 0xffff_ffff_ffff_0000) + (1 << 16)
+}
+
 // NewV7 returns a new version 7 UUID.
 //
 // Version 7 UUIDs contain a timestamp in the most significant 48 bits,
@@ -224,18 +233,15 @@ func NewV7() UUID {
 	//
 	// We store a 12 bit sub-millisecond timestamp fraction in the rand_a section,
 	// as optionally permitted by the RFC.
-	v7mu.Lock()
-
-	// Generate our 60-bit timestamp: 48 bits of millisecond-resolution,
-	// followed by 12 bits of 1/4096-millisecond resolution.
 	now := time.Now()
 	secs := uint64(now.Unix())
 	nanos := uint64(now.Nanosecond())
 	msecs := nanos / 1000000
 	frac := nanos - (1000000 * msecs)
-	timestamp := (1000*secs + msecs) << 12 // ms shifted into position
+	timestamp := (1000*secs + msecs) << 16 // ms shifted into position; bits 12-15 reserved for ver
 	timestamp += (frac * 4096) / 1000000   // ns converted to 1/4096-ms units
 
+	v7mu.Lock()
 	if v7lastSecs > secs {
 		// Time has gone backwards.
 		// This presumably indicates the system clock has changed.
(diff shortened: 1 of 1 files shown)


dxasu:
I signed it!

